PR #1, opened 2018-06-06
Initial commit of the exporter daemon and its Go exporter

rakyll:
This commit contains an initial prototype for the exporter daemon
that can discovered via and endpoint file on the host.

The exporter looks for the endpoint file to see the availability
of the daemon and exports if it is running.

Exporter daemon will allow OpenCensus users to export
without having to link a vendor-specific exporter in their final
binaries. We are expecting the prototype exporter is going to
be implemented in every language and registered by default.

### DIFF
--- a/cmd/opencensusd/main.go
+++ b/cmd/opencensusd/main.go
@@ -1,0 +1,93 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Program opencensusd collects OpenCensus stats and traces
+// to export to a configured backend.
+package main
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net"
+	"os"
+	"os/signal"
+	"path/filepath"
+
+	pb "github.com/census-instrumentation/opencensus-proto/gen-go/exporterproto"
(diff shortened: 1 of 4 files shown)


cmd/opencensusd/main.go
```diff
@@ -0,0 +1,93 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Program opencensusd collects OpenCensus stats and traces
+// to export to a configured backend.
+package main
```
ramonza: Can we rename this repo to "census-instrumentation/opencensusd"?

exporter/exporter.go
```diff
@@ -0,0 +1,177 @@
...
+
+	initOnce sync.Once
+
+	clientMu       sync.Mutex
+	clientEndpoint string
+	spanClient     exporterproto.Export_ExportSpanClient
+}
+
+func (e *Exporter) init() {
+	go func() {
```
ramonza: Do we need to start a background Goroutine? Why not just check at export time?

cmd/opencensusd/main.go
rakyll: opencensusd will only be one of the binaries/components from the service. We are planning to add more in the future. That's why the repo is named opencensus-service.

exporter/exporter.go
rakyll: I didn't want to check the file and the changes at every time there is export. There are numerous number of exports happening at any time and we don't want to be too aggressively looking up for the file. Btw, we are lazily starting the goroutine at the export time.

ramonza: You could use the a rate limiter to prevent too much checking.

rakyll: In terms of behavior, it is not going to change anything but complicate the implementation.

What's the problem of running a goroutine? It doesn't have any cost. It is not like allocating an actual OS thread or anything.

ramonza: I know but it will appear in stack dumps etc. In general, I would be surprised if a library starts a background process automatically without any opt-in by the user.

PR #2, opened 2018-06-06
Add travis build configuration

rakyll:


rakyll:
@ramonza @g-easy, can you take a look?

.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: go
+
+go:
+  - 1.10.x
+
+before_script:
+  - GO_FILES=$(find . -iname '*.go' | grep -v /vendor/)  # All the .go files, excluding vendor/ if any
+
+script:
+  - if [ -n "$(gofmt -s -l $GO_FILES)" ]; then echo "gofmt the following files:"; gofmt -s -l $GO_FILES; exit 1; fi
```
ramonza: should we use goimports instead?

rakyll: goimports are only for editing time, they only autoimport the missing packages.

PR #6, opened 2018-06-22
Make endpoint file a JSON

### LINKED ISSUE #5: Convert endpoint file to a JSON file
This will give us more flexibility to add more information about opencensusd if required in the future.

rakyll:
It allows us to extend what else we are going to store
as metadata in the file.

Fixes #5.

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -15,15 +15,54 @@
 package internal
 
 import (
+	"encoding/json"
+	"io/ioutil"
 	"os"
 	"os/user"
 	"path/filepath"
 	"runtime"
 )
 
-// DefaultEndpointFile is the location where the
+// Service contains metadata about the exporter service.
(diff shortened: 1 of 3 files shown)


internal/internal.go
```diff
@@ -15,15 +15,54 @@
...
-// DefaultEndpointFile is the location where the
+// Service contains metadata about the exporter service.
+type Service struct {
+	Endpoint string `json:"endpoint"`
+}
+
+// WriteToEndpointFile writes service metadata to
+// canonical endpoint file.
+func (s *Service) WriteToEndpointFile() (path string, err error) {
+	data, err := json.Marshal(&s)
```
ramonza: Do you need to pass a pointer to a pointer here?

rakyll: Hoho, no. Fixed.

Issue #8, opened 2018-06-29
Export to configured backends


rakyll:
Define an experimental configuration format or flags and export the data from opencensusd to backends.

Note that there is a concern with this model with Prometheus because of the way Prometheus servers are sharded. Prometheus is discovering the nodes itself and this is how servers are sharded. We should be careful about not breaking this model with the exporter daemon. That's why I am suggesting us not to export a Prometheus /metrics endpoint for now.

I am suggesting to export to the following backends for now:

* Zipkin
* Datadog
* Stackdriver

odeke-em:
@rakyll you already implemented tracing exporting to those backends and as per the new direction, we haven't yet added metrics/stats so this issue is effectively solved. Thanks for working on it!

PR #9, opened 2018-06-29
Add experimental exporter support for traces

rakyll:
Example config.yaml:

stackdriver:
  project: "projectid"
  enableTraces: true

zipkin:

Exporter will export the traces both to Stackdriver
and Zipkin.

Updates #8.


cmd/opencensusd/exporters.go
```diff
@@ -0,0 +1,145 @@
...
+func flush() {
+	for _, fn := range flushers {
+		fn()
+	}
+}
+
+type config struct {
+	Datadog struct {
+		// Namespace specifies the namespaces to which metric keys are appended.
+		Namespace string `yaml:"namespace_addr,omitempty"`
```
ramonza: I think we should use camelCase for yaml attributes, since this is more consistent with Kubernetes config.

cmd/opencensusd/exporters.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// Configure Zipkin.
+	if endpoint := c.Zipkin.Endpoint; endpoint != "" {
+		// TODO(jbd): Propagate service name, hostport and more metadata from each node.
+		localEndpoint, err := openzipkin.NewEndpoint("server", "")
+		if err != nil {
+			log.Fatalf("Cannot configure Zipkin exporter: %v", err)
+		}
+		reporter := http.NewReporter(endpoint)
```
ramonza: I think this needs to be closed at the end of the program. It doesn't support flush but maybe we should change flushers to closers?

rakyll:
@ramonza I am thinking about a registration pattern. Exporter registers its config parsing method and makers for stats/trace exporters. We can even utilize the shared libraries to dynamically load themselves in runtime.

cmd/opencensusd/exporters.go
rakyll: Done.

cmd/opencensusd/exporters.go
rakyll: Done.

rakyll:
@songy23, for your sanity, don't look at the vendor files :)

rghetia:
I am not familiar with Go yet so my comment may not make sense. 
I see large number of files being checked in which are vendor files. I guess these are similar vendor files in nodejs. Is there a way to not checkin all these vendor files to the repo?

PR #16, opened 2018-08-24
Use grpc-gateway to add HTTP/JSON endpoint option

draffensperger:
This depends on https://github.com/census-instrumentation/opencensus-proto/pull/77 but wanted to put this out early to show the concept of how it would be used.

If the `listen_http` argument is passed to `opencensusd` it will accept HTTP/JSON requests and proxy them to its own gRPC endpoint.

This would enable clients who do not have gRPC support (older systems, languages with no gRPC compilation) and also browser clients (some authentication middle layer / API gateway would be a best practice too, but just allowing HTTP is a start, and users could use API gateways from a cloud provider).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -37,6 +37,7 @@
 [[projects]]
   name = "github.com/golang/protobuf"
   packages = [
+    "jsonpb",
     "proto",
     "protoc-gen-go/descriptor",
     "ptypes",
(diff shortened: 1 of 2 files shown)


odeke-em:
Hello @draffensperger, thank you for this PR! Unfortunately the direction of this repository and implementation changed and for now we aren't yet adding a gateway but perhaps maybe later on once we have a working and deployed version. Please feel free to reopen later, I am closing this for housekeeping purposes. Thank you.

PR #17, opened 2018-08-29
Fix import path from gen-go files.

songy23:
~~Do not merge until https://github.com/census-instrumentation/opencensus-proto/pull/86 is merged.~~


songy23:
Thanks!

PR #28, opened 2018-09-20
all: rewrite from groundup

odeke-em:
Due to the complexity of setting up repositories, it has
been advised that we reuse this repository. However, since
the agent spec has radically changed and is being
implemented from the ground up, we are rewriting this
code hence the deletions.


odeke-em:
Thanks for the review @songy23! I think this change is going to be stalled as @rakyll needs this repo as is for some demos, as per offline email communication -- unless am mistaken.

songy23:
Make sense to me. Let's hold off this PR for now, meanwhile we can still do the implementation in other PRs.

bogdandrutu:
For the moment to not break the old prototype consider to move it into a directory "/old" or something like that.

odeke-em:
No longer necessary. Replaced by the proper code placements #34 #35 #36.

PR #31, opened 2018-09-21
spanreceiver: define SpanReceiver interface

### LINKED ISSUE #30: trace: define the interface that the various interceptors will forward their spans to
We need to define a mechanism for which all the various interceptors will forward their spans to.
It needs to receive a node to uniquely identify the spans being forwarded.

Perhaps for brevity
```go
type SpanReceiver interface {
       ReceiveSpans(node *commonpb.Node, spans ...*tracepb.Span) (*Acknowledgement, error)
}
```

and when creating an interceptor, we'll pass in a `SpanReceiver` e.g.
```go
func zipkininterceptor.New(sr SpanReceiver)
func ocinterceptor.New(sr SpanReceiver)
```

odeke-em:
An interface `SpanReceiver` that receives *tracepb.Span instances
and sends back an acknowledgement and an error.

The return signature is `(*Acknowledgement, error)` because it
would useful to return to callers information about how many spans
were successful, errors that were encountered, allowing for batches
of spans to be written like we would with `io.Write`. This will allow
callers to retry, report success and failure rates too.

Fixes #30

### DIFF
--- a/spanreceiver/spanreceiver.go
+++ b/spanreceiver/spanreceiver.go
@@ -1,0 +1,30 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanreceiver
+
+import (
+	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
+	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
+)
+
+// SpanReceiver is an interface that receives spans from a Node identifier.
+type SpanReceiver interface {
+	ReceiveSpans(node *commonpb.Node, spans ...*tracepb.Span) (*Acknowledgement, error)
+}
+
+type Acknowledgement struct {
+	SavedSpans   uint64
+	DroppedSpans uint64
+}


spanreceiver/spanreceiver.go
```diff
@@ -0,0 +1,29 @@
...
+
+package spanreceiver
+
+import (
+	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
+)
+
+// SpanReceiver is an interface that receives spans.
+type SpanReceiver interface {
+	ReceiveSpans(...*tracepb.Span) (*Acknowledgement, error)
```
bogdandrutu: We also need to receive the Node. Even though we send it once we will cache that value and always make it available.

odeke-em: Yeah, you read my mind, I was literally pushing that up as your comment came in :)

odeke-em:
Thanks for the review @bogdandrutu!

spanreceiver/spanreceiver.go
```diff
@@ -0,0 +1,30 @@
...
+package spanreceiver
+
+import (
+	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
+	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
+)
+
+// SpanReceiver is an interface that receives spans from a Node identifier.
+type SpanReceiver interface {
+	ReceiveSpans(node *commonpb.Node, spans ...*tracepb.Span) (*Acknowledgement, error)
```
bogdandrutu: Not a go expert, but what is the advantage to use variadic here vs an array?

odeke-em: Great question! It makes for a nice API usage when calling items individually
```go
ReceiveSpans(node, span1, span2)
```
vs requiring users to create an array
```go
spans := []*tracepb.Span{span1, span2}
ReceiveSpans(node, spans)
// or
ReceiveSpans(node, []*tracepb.Span{span1, span2})
```
but also works in the general case of when we have an array
```go
ReceiveSpans(node, spans...)
```

odeke-em: Great question! It makes for a nice API usage when calling items individually
```go
ReceiveSpans(node, span1, span2)
```
vs requiring users to create a list first
```go
spans := []*tracepb.Span{span1, span2}
ReceiveSpans(node, spans)
// or
ReceiveSpans(node, []*tracepb.Span{span1, span2})
```
but also works in the general case of when we have a list
```go
ReceiveSpans(node, spans...)
```

PR #34, opened 2018-09-21
trace/interceptor: implement OpenCensus interceptor

### LINKED ISSUE #32: trace/interceptors: define and implement the opencensus interceptor
The OpenCensus interceptor receives *tracepb.Span batches from OpenCensus library applications like `ocagent-exporter` would. I've already filed a PR https://github.com/census-instrumentation/opencensus-service/pull/25 but that wasn't defined as an interceptor.

odeke-em:
Implemented the OpenCensus interceptor that
opencensus-agent-exporters in the various languages
will send traces to. It is initialized with at least
a non-nil spanreceiver.SpanReceiver.

I also added an end-to-end test to ensure that traces
are properly proxied to the spanreceiver.SpanReceiver.

Fixes #32

### DIFF
--- a/interceptor/opencensus/opencensus.go
+++ b/interceptor/opencensus/opencensus.go
@@ -1,0 +1,105 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ocinterceptor
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	"google.golang.org/api/support/bundler"
+
+	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
+	agenttracepb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1"
+	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
+	"github.com/census-instrumentation/opencensus-service/spanreceiver"
+)
+
+type OCInterceptor struct {
+	spanSink         spanreceiver.SpanReceiver
+	spanBufferPeriod time.Duration
+}
+
+func New(sr spanreceiver.SpanReceiver, opts ...OCOption) (*OCInterceptor, error) {
+	if sr == nil {
+		return nil, errors.New("needs a non-nil spanReceiver")
+	}
+	oci := &OCInterceptor{spanSink: sr}
+	for _, opt := range opts {
+		opt.WithOCInterceptor(oci)
+	}
+	return oci, nil
+}
+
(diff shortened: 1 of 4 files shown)


interceptor/opencensus/opencensus.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return oci, nil
+}
+
+var _ agenttracepb.TraceServiceServer = (*OCInterceptor)(nil)
+
+var errUnimplemented = errors.New("unimplemented")
+
+func (oci *OCInterceptor) Config(tcs agenttracepb.TraceService_ConfigServer) error {
+	// TODO: Implement when we define the config receiver/sender.
```
bogdandrutu: Good understanding of the design here. Thanks.

interceptor/opencensus/opencensus.go
```diff
@@ -0,0 +1,106 @@
...
+	return errUnimplemented
+}
+
+// Export is the gRPC method that receives streamed traces from
+// OpenCensus-traceproto compatible libraries/applications.
+func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) error {
```
bogdandrutu: I think there is a small problem here, Export can be called by multiple tasks in the same time, but in your current implementation you have only on bundler, here are the problems that you need to take care:
* Bundle has to be protected with a mutex if it is not thread-safe (gorutine safe) because it is shared between multiple calls.
* You need to add in the bundle {node, span[]} because if you receive from multiple nodes you need to keep the association between node and spans.

Alternatively (which probably is better) is to move bundle to be a local variable in the Export and avoid having it as a class member which cause all the problems I described.

odeke-em: @bogdandrutu I removed the bundler field that was causing the confusion, my apologies. PTAL

interceptor/opencensus/opencensus.go
```diff
@@ -0,0 +1,105 @@
...
+			flattenedListOfSpans = append(flattenedListOfSpans, listOfSpans...)
+		}
+		oci.spanSink.ReceiveSpans(node, flattenedListOfSpans...)
+	})
+	spanBufferPeriod := oci.spanBufferPeriod
+	if spanBufferPeriod <= 0 {
+		spanBufferPeriod = 2 * time.Second // Arbitrary value
+	}
+	traceBundler.DelayThreshold = spanBufferPeriod
+	traceBundler.BundleCountThreshold = 50 // Arbitrary value
```
bogdandrutu: Should this be an option?

odeke-em: Great idea. I only made an option for the span buffer period as I thought perhaps that's the most configuration, but for sure I'll break out an option even for countthreshold

interceptor/opencensus/opencensus.go
```diff
@@ -0,0 +1,112 @@
...
+	// TODO: Implement when we define the config receiver/sender.
+	return errUnimplemented
+}
+
+// Export is the gRPC method that receives streamed traces from
+// OpenCensus-traceproto compatible libraries/applications.
+func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) error {
+	// Firstly we MUST receive the node identifier to initiate
+	// the service and start accepting exported spans.
+	const maxTraceInitRetries = 15 // Arbitrary value
```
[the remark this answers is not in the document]
odeke-em: Roger that, I've filed https://github.com/census-instrumentation/opencensus-service/issues/43 for multiplexing support. In regards to making that an option, I think for now we can leave it as is and since that logic is changing soon we'll revisit.

odeke-em:

Issue #26, opened 2018-09-20
trace/exporters: define how exporters receive traces


odeke-em:
We need to be able to convert traces received by the agent into exportable traces.

For the purpose of having each library being able to upload traces to the agent, currently traces received are of the trace-proto, we need to define the various trace transformers and perhaps provide protocol converters based off the data model per backend and the various protocols e.g. thrift, gRPC, HTTP. OpenCensus Go already has a bunch of exporters but transforming back to opencensus-go trace has a few limitations like attributes on Annotations being untyped https://godoc.org/go.opencensus.io/trace#Annotation but otherwise we can transform them back to OpenCensus-Go traces and perhaps it might make sense for us to reuse those already implemented exporters.

In the future, a nice to have(not important right now) would be a scheme for translating trace-proto fields to those of an exporter's backend and then we can auto-generate the clients depending on the transport type, as more backends are added.

odeke-em:
The simplest option here would be to translate:
   `*tracepb.Span`
to
   `*trace.SpanData`
and then use all the already implemented OpenCensus-Go exporters.

Interestingly though, opencensus-agent-exporters translate from library-spanData to `*tracepb.Span` e.g. the ocagent-go-exporter translates from `*trace.SpanData` to `*tracepb.Span`.

### Synopsis
This task will be to make a translator for `*tracepb.Span` to `*trace.SpanData`

odeke-em:
I've mailed out two PRs that provide each half of conversion

Conversion|PR
---|---
OCSpanDataToProtoSpan|https://github.com/census-instrumentation/opencensus-service/pull/35
ProtoSpanToOCSpanData|https://github.com/census-instrumentation/opencensus-service/pull/36

songy23:
> The simplest option here would be to translate:
*tracepb.Span
to
*trace.SpanData
and then use all the already implemented OpenCensus-Go exporters.

Yes I think that's fine for the first version. In the future we may consider updating the original Go exporters so that they can accept `SpanProto` directly. That should be a user-invisible change so we can update it anytime.

odeke-em:
>Yes I think that's fine for the first version. In the future we may consider updating the original Go exporters so that they can accept SpanProto directly. That should be a user-invisible change so we can update it anytime.

This would require us though to rewrite them which is kind of what I had talked about at the beginning of this issue but we'll instead go with the current scalable option of reusing existent exporters.

[...]

PR #35, opened 2018-09-21
translator/trace: OCSpanDataToProtoSpan

odeke-em:
Provides a function to translate OpenCensus SpanData
into *tracepb.Span, with tests. It is the first half of
the trace converters task.

Updates #26

### DIFF
--- a/translator/trace/spandata_to_protospan.go
+++ b/translator/trace/spandata_to_protospan.go
@@ -1,0 +1,169 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


bogdandrutu:
Not sure we need this here (cannot see for the moment when we will use this). Also do we not need this functionality in the ocagent-exporter?

songy23:
> Not sure we need this here (cannot see for the moment when we will use this). Also do we not need this functionality in the ocagent-exporter?

Agree, AFAIS this should be a functionality for Agent exporters.

odeke-em:
@bogdandrutu @songy23 currently we don't need it but I implemented it as it is an item on our task list. Yes this functionality is used in the Go-exporter but I was wondering if perhaps we should have it in one place and import from a set of common libraries.

odeke-em:
This PR was an overzealous interpretation of the tasks. We already have this functionality in the agent-exporters. I'll close it for now but the functionality is trivial to retrieve if we ever need it.

Thanks y'all for the reviews!

Issue #33, opened 2018-09-21
vendor: either update vendored packages or remove them


odeke-em:
Am currently writing code to implement the OpenCensus Interceptor and "go.opencensus.io/trace" is vendored which is interrupting my tests.

We should either update these or remove them entirely and then use go modules perhaps.

bogdandrutu:
I would suggest to remove and use go modules. Thanks.

rakyll:
Please don't stop vendoring. Go modules support vendoring and is still the current way of having reproducible builds in Go.

odeke-em:
>Please don't stop vendoring. Go modules support vendoring and is still the current way of having reproducible builds in Go.

In deed, I mentioned this at the beginning of this issue. I sent the first PR to remove them lest we'd have massive PRs with additions and removals. The follow-up PR is going to be for `go mod` vendorship

PR #40, opened 2018-09-21
vendor: remove vendoring by dep

odeke-em:
Firstly removing vendor/ and dep
which will be followed up by a change
switching to using modules. For now
this change enables Travis CI to build.
Could have actually just updated dependencies
using dep but that'd be an extra change and
we are already revamping this repository.

Updates #33


odeke-em:
The rest are code deletions except for
screen shot 2018-09-21 at 12 39 02 pm
which used a no longer supported import path.

odeke-em:
Thanks @bogdandrutu @songy23 for the review!

PR #45, opened 2018-09-28
exporter: plumbing to glue SpanReceiver with OC exporters

odeke-em:
* Implemented the exporter.TraceExporter interface whose
method
```go
ExportSpanData(*commonpb.Node, ...*trace.SpanData) error
```

is feeds into a transformer that creates a spanreceiver.SpanReceiver
of which that signature
```go
ReceiveSpans(*commonpb.Node, ...*trace.SpanData) (*spanreceiver.Acknowledgement, error)
```
Also provided a dummy transformer which takes trace.Exporter-s and
creates a spanreceiver.SpanReceiver that that is passed to each
interceptor.

* Also provided an end-to-end application which if copied and pasted
is the prototype for our cmd/ocagent code that exports traces to the
various traditional OpenCensus trace.Exporter-s that have grown around
the OpenCensus ecosystem. The config parsing phase will determine
which trace.Exporter-s will be used. In the demo, application,
I am able to send traces to both Stackdriver and Jaeger simultaneously

#### Exhibits
To run this demo, we'll export to:

Exporter|Help setting it up
---|---
Stackdriver Tracing|https://opencensus.io/codelabs/stackdriver/
Jaeger|https://opencensus.io/codelabs/jaeger/ 

On running `cd exporter` and then
```shell
GOOGLE_APPLICATION_CREDENTIALS=~/Downloads/census-demos-a835580cf6cc.json go test -v
```
screen shot 2018-09-28 at 1 57 26 am
screen shot 2018-09-28 at 1 57 43 am
screen shot 2018-09-28 at 1 57 33 am


odeke-em:
This PR is independent and a parallel task to https://github.com/census-instrumentation/opencensus-service/pull/44 so don't worry about the order of review and merging.

bogdandrutu:
Please fix the build.

odeke-em:
>Please fix the build.

The build failure is coming from the old files which would require deletions, but that'll make this PR huge. I'll send the deletions PR as a separate PR.

odeke-em:
@bogdandrutu @songy23 I've fixed the Travis build PTAL.

exporter/trace_exporter.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type TraceExporterSpanReceiver interface {
+	TraceExporter
+	spanreceiver.SpanReceiver
+}
+
+// OCExportersToTraceExporterSpanReceiver is a convenience function that transforms
+// traditional OpenCensus trace.Exporter-s into an opencensus-service TraceExporter.
+// The resulting TraceExporter ignores the passed in node. To make use of the node,
```
songy23: Can you give an example on how exporters (like Stackdriver Trace) can use `Node`?

exporter/trace_exporter.go
```diff
@@ -0,0 +1,77 @@
...
+func (tse *toOCExportersTransformer) ReceiveSpans(node *commonpb.Node, spans ...*tracepb.Span) (*spanreceiver.Acknowledgement, error) {
+	// Firstly transform them into spanData.
+	spanDataList := make([]*trace.SpanData, 0, len(spans))
+	for _, span := range spans {
+		spanData, err := tracetranslator.ProtoSpanToOCSpanData(span)
+		if err == nil {
+			spanDataList = append(spanDataList, spanData)
+		}
+	}
+
```
songy23: Please leave a note here that we may enhance the `SpanData` before they got sent to exporters.

exporter/trace_exporter.go
odeke-em: @songy23 great question. I've filed #46 but basically Stackdriver Monitoring should be able to use the NodeInfo when we create a new exporter to match up the "opencensus_task" identifier with the node when we create an exporter per language-PID combination and then pass in a DefaultMonitoringLabel so for example
```go
func (e *stackdriverKeeper) exporterForNodeLocked(node *commonpb.Node) (*stackdriver.Exporter, error) {
	langPIDKey := extractLangPIDKey(node)
	if exp, ok := e.perLangPID[langPIDKey]; ok && exp != nil {
		return exp, nil
	}

	nodeLabels := new(stackdriver.Labels)
	nodeLabels.Set("opencensus_task", langPIDKey, "Opencensus task identifier")
	// Otherwise now create the exporter
	sde, err := stackdriver.NewExporter(stackdriver.Options{
		DefaultMonitoringLabels: nodeLabels,
	})
	if err != nil {
		return nil, err
	}
	// Now memoize it
	e.perLangPID[langPIDKey] = sde
	return sde, nil
}
```

exporter/trace_exporter.go
odeke-em: For sure, I'll do so right now.

exporter/trace_exporter.go
odeke-em: The AWS X-Ray exporter can also use an option to identify the service name https://docs.aws.amazon.com/xray/latest/api/API_Service.html which we shall be able to retrieve from the Node. However, we also need the Node to be able to generate the various zPages.

odeke-em:

PR #48, opened 2018-10-02
cmd/ocagent: retrofitted with new protocol

odeke-em:
With this change, cmd/ocagent now adheres
to the new protocol and currently runs
the OpenCensus interceptor while exporting
received spans to the various and previously
already created YAML configured exporters such as:
* Stackdriver Tracing
* DataDog
* Zipkin

which are statically imported.

Now any program that has an ocagent-exporter
can send traces to this service. The service's
OpenCensus Interceptor Port can be configured by
```shell
ocagent -oci-port=<value>
```
which by default is `55678`

After this change we can add both more trace exporters
and more trace interceptors.

Essentially this can be our v0.0.1 of the new trace
agent/service.

### DIFF
--- a/cmd/ocagent/exporterparser/datadog.go
+++ b/cmd/ocagent/exporterparser/datadog.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package exporter
+package exporterparser
 
 import (
 	"log"
@@ -24,23 +24,22 @@ import (
 )
 
 type dataDogConfig struct {
-	Datadog struct {
+	Datadog *struct {
 		// Namespace specifies the namespaces to which metric keys are appended.
 		Namespace string `yaml:"namespace,omitempty"`
 
 		// TraceAddr specifies the host[:port] address of the Datadog Trace Agent.
 		// It defaults to localhost:8126.
-		TraceAddr string `yaml:"traceAddr,omitempty"`
+		TraceAddr string `yaml:"trace_addr,omitempty"`
 
 		// MetricsAddr specifies the host[:port] address for DogStatsD. It defaults
 		// to localhost:8125.
-		MetricsAddr string `yaml:"metricsAddr,omitempty"`
+		MetricsAddr string `yaml:"metrics_addr,omitempty"`
 
 		// Tags specifies a set of global tags to attach to each metric.
 		Tags []string `yaml:"tags,omitempty"`
 
-		EnableMetrics bool `yaml:"enableMetrics,omitempty"`
-		EnableTraces  bool `yaml:"enableTraces,omitempty"`
+		EnableTracing bool `yaml:"enable_tracing,omitempty"`
 	} `yaml:"datadog,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


cmd/ocagent/exparser/datadog.go
```diff
@@ -51,21 +51,29 @@ func (d *datadogExporter) MakeExporters(config []byte) (se view.Exporter, te tra
...
+	}
+
+	// TODO(jbd): Create new exporter for each service name.
+	de := datadog.NewExporter(datadog.Options{
+		Namespace: dc.Namespace,
+		TraceAddr: dc.TraceAddr,
+		StatsAddr: dc.MetricsAddr,
+		Tags:      dc.Tags,
+	})
+	if dc.EnableMetrics {
```
songy23: Do we need this now?

cmd/ocagent/exparser/exparser.go
```diff
@@ -12,19 +12,26 @@
...
-package exporter
+// This package provides support for parsing and creating the
+// respective exporters given a YAML configuration payload.
+// For now it currently only provides statically imported OpenCensus
+// exporters like:
+//  * Stackdriver Tracing and Monitoring
+//  * DataDog
+//  * Zipkin
+
+package exparser
```
songy23: nit: `exparser` sounds confusing, may name this package to `exporterparser` or just `exporter`?

cmd/ocagent/exparser/stackdriver.go
```diff
@@ -38,24 +39,32 @@ func (s *stackdriverExporter) MakeExporters(config []byte) (se view.Exporter, te
...
+		log.Fatal("Stackdriver config requires a project ID")
+	}
+	exporter, err := stackdriver.NewExporter(stackdriver.Options{
+		ProjectID:    sc.ProjectID,
+		MetricPrefix: sc.MetricPrefix,
+	})
+	if err != nil {
+		log.Fatalf("Cannot configure Stackdriver exporter: %v", err)
+	}
+	if sc.EnableMetrics {
```
songy23: ditto

cmd/ocagent/exparser/datadog.go
odeke-em: @songy23 that was just translating the old code :)

cmd/ocagent/exparser/exparser.go
odeke-em: Yeah I also wondered about that name but:
* exporterparser is too long for a Go package name
* exporter clashes with the already available package names

I'll just bite the bullet and go with `exporterparser` for clarity and that can be changed later on since anyways it is an internal package to `cmd/ocagent/internal/`

songy23: SGTM

cmd/ocagent/exparser/stackdriver.go
odeke-em: Just a translation of the old code :)

cmd/ocagent/exparser/exparser.go
odeke-em: I've updated it, PTAL.

odeke-em:
@songy23 I've removed the vestiges containing `Metrics` in the exporters as requested. PTAL. Also I believe @rakyll is busy and away at a conference so I'll merge this PR (as it has been approved by you) and any revisions are welcome.

Issue #46, opened 2018-10-01
opencensusexporter/stackdriver: Node should translate to unique opencensusTaskKey


odeke-em:
After #45 is merged, we'll have the ability to export traces proxied from multiple node but at the single exit point of the service. This issue is a placeholder to track and discuss how we'll be able to make a unique Stackdriver exporter per PID and take advantage of each Node's information.

odeke-em:
I just suggested some code to enable this and it should be sufficient to get this working for multiplexing to nodes
```go
func (e *stackdriverKeeper) exporterForNodeLocked(node *commonpb.Node) (*stackdriver.Exporter, error) {
	langPIDKey := extractLangPIDKey(node)
	if exp, ok := e.perLangPID[langPIDKey]; ok && exp != nil {
		return exp, nil
	}

	nodeLabels := new(stackdriver.Labels)
	nodeLabels.Set("opencensus_task", langPIDKey, "Opencensus task identifier")
	// Otherwise now create the exporter
	sde, err := stackdriver.NewExporter(stackdriver.Options{
		DefaultMonitoringLabels: nodeLabels,
	})
	if err != nil {
		return nil, err
	}
	// Now memoize it
	e.perLangPID[langPIDKey] = sde
	return sde, nil
}
```

songy23:
Note that `Node` also has information on `MonitoredResource`/resource attributes. We need a way to translate that part of information to `MonitoredResource`/resource attributes too.

odeke-em:
>Note that Node also has information on MonitoredResource/resource attributes.

@songy23 do you mean in Node.attributes as per https://github.com/census-instrumentation/opencensus-proto/blob/df894b5863737edd0b77dade8bac59b64c4a6f98/src/opencensus/proto/agent/common/v1/common.proto#L44-L45?
or should this be in serviceInfo?

songy23:
Yes, for now the resource metadata will be in the `map<string, string> attributes`. Note there's an ongoing discussion about resource specs: https://github.com/census-instrumentation/opencensus-specs/pull/162. Later we may want to add a dedicated message that includes the resource information.

flands:
@songy23 is this still relevant?

songy23:
This is already fixed - closing.

PR #50, opened 2018-10-02
interceptor/opencensus: support multiplexing in Export

### LINKED ISSUE #43: trace/interceptor/opencensus: Export should support node multiplexing
Given the minor updated design from https://github.com/census-instrumentation/opencensus-proto/pull/119, where we could be talking to a "pass-through" like multiplexer that multiple applications behind it, it is imperative that we recognize that if at anytime a Node changes(different from what our initiation had received), we should upload it with a specialized node.

This issue will be solvable after the initial OpenCensus trace interceptor is merged from https://github.com/census-instrumentation/opencensus-service/pull/34

odeke-em:
Add multiplexing/"pass through" mode where spans from
sources other than the initiating node can be sent
and they should be properly proxied to their final
destination. This change also adds in tests to ensure
this behavior and to avoid regressions in the future.

Fixes #43

### DIFF
--- a/interceptor/opencensus/opencensus.go
+++ b/interceptor/opencensus/opencensus.go
@@ -53,6 +53,11 @@ func (oci *OCInterceptor) Config(tcs agenttracepb.TraceService_ConfigServer) err
 	return errUnimplemented
 }
 
+type spansAndNode struct {
+	spans []*tracepb.Span
+	node  *commonpb.Node
+}
+
 // Export is the gRPC method that receives streamed traces from
 // OpenCensus-traceproto compatible libraries/applications.
 func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) error {
@@ -60,7 +65,7 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
 	// the service and start accepting exported spans.
 	const maxTraceInitRetries = 15 // Arbitrary value
 
-	var node *commonpb.Node
+	var initiatingNode *commonpb.Node
 	for i := 0; i < maxTraceInitRetries; i++ {
 		recv, err := tes.Recv()
 		if err != nil {
@@ -68,25 +73,18 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
 		}
 
 		if nd := recv.Node; nd != nil {
-			node = nd
+			initiatingNode = nd
 			break
 		}
 	}
 
-	if node == nil {
-		return fmt.Errorf("failed to receive a non-nil Node even after %d retries", maxTraceInitRetries)
+	if initiatingNode == nil {
+		return fmt.Errorf("failed to receive a non-nil initiating Node even after %d retries", maxTraceInitRetries)
 	}
 
 	// Now that we've got the node, we can start to receive streamed up spans.
 	// The bundler will receive batches of spans i.e. []*tracepb.Span
-	traceBundler := bundler.NewBundler(([]*tracepb.Span)(nil), func(payload interface{}) {
-		listOfSpanLists := payload.([][]*tracepb.Span)
-		flattenedListOfSpans := make([]*tracepb.Span, 0, len(listOfSpanLists)) // In the best case, each spanList has 1 span
-		for _, listOfSpans := range listOfSpanLists {
-			flattenedListOfSpans = append(flattenedListOfSpans, listOfSpans...)
-		}
-		oci.spanSink.ReceiveSpans(node, flattenedListOfSpans...)
-	})
+	traceBundler := bundler.NewBundler((*spansAndNode)(nil), oci.batchSpanExporting)
 	spanBufferPeriod := oci.spanBufferPeriod
 	if spanBufferPeriod <= 0 {
 		spanBufferPeriod = 2 * time.Second // Arbitrary value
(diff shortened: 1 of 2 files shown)


interceptor/opencensus/opencensus_test.go
```diff
@@ -154,7 +158,122 @@ func TestOCInterceptor_endToEnd(t *testing.T) {
...
+	}
+
+	gotBlob, _ := json.Marshal(resultsMapping)
+	wantBlob, _ := json.Marshal(wantContents)
+	if !bytes.Equal(gotBlob, wantBlob) {
+		t.Errorf("Unequal serialization results\nGot:\n\t%s\nWant:\n\t%s\n", gotBlob, wantBlob)
+	}
+}
+
+func nodeToKey(n *commonpb.Node) string {
```
songy23: This is fine for now. Like I mentioned in https://github.com/census-instrumentation/opencensus-service/issues/46#issuecomment-426077190, please add a note that later `Node` could be turned into some other messages that represent monitored resource.

interceptor/opencensus/opencensus.go
```diff
@@ -53,40 +53,38 @@ func (oci *OCInterceptor) Config(tcs agenttracepb.TraceService_ConfigServer) err
 	return errUnimplemented
 }
 
+type nodeSpansPair struct {
```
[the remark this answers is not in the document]
odeke-em: It is a pair though that is node->spans :) but sure I'll make that change to `spansWithNode`.

interceptor/opencensus/opencensus_test.go
odeke-em: This is test code though, just to ensure that the proper nodes were sent over the wire

interceptor/opencensus/opencensus.go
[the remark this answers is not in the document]
odeke-em: I've updated that struct name, PTAL @songy23

interceptor/opencensus/opencensus.go
```diff
@@ -106,7 +104,26 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
...
+		}
+
+		bundlerPayload := &spansWithNode{node: node, spans: recv.Spans}
+		// Otherwise add them to the bundler
+		traceBundler.Add(bundlerPayload, len(bundlerPayload.spans))
+	}
+}
+
+func (oci *OCInterceptor) batchSpanExporting(payload interface{}) {
+	nodeSpanPairs := payload.([]*spansWithNode)
```
songy23: nit: s/nodeSpanPairs/spansWithNode

odeke-em: Aha, sure #searchAndReplace

odeke-em:
Thank you for the review @songy23, I'll wait for @bogdandrutu to provide feedback too then merge it.

interceptor/opencensus/opencensus.go
```diff
@@ -106,7 +104,26 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
 			return err
 		}
 
-		// Otherwise add these spans to the bundler
-		traceBundler.Add(recv.Spans, len(recv.Spans))
+		// As per https://github.com/census-instrumentation/opencensus-service/issues/43,
+		// we support multiplexing. Thus, if a Node hasn't been sent from downstream, assume
+		// that those spans are from the initiatingNode
```
bogdandrutu: The comment says that if the Node is not sent assume spans are from the last received Node:
https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/agent/trace/v1/trace_service.proto#L68

interceptor/opencensus/opencensus.go
```diff
@@ -53,40 +53,38 @@ func (oci *OCInterceptor) Config(tcs agenttracepb.TraceService_ConfigServer) err
...
 		if nd := recv.Node; nd != nil {
-			node = nd
+			initiatingNode = nd
 			break
 		}
 	}
 
-	if node == nil {
-		return fmt.Errorf("failed to receive a non-nil Node even after %d retries", maxTraceInitRetries)
+	if initiatingNode == nil {
```
bogdandrutu: Why is this a special case? Also if the first message contains Node and Spans we don't collect the spans.

odeke-em:
@bogdandrutu thanks for the comments and for catching those 2 bugs for which I've filed https://github.com/census-instrumentation/opencensus-service/issues/51 and https://github.com/census-instrumentation/opencensus-service/issues/53.

However, could we please merge this issue and then I'll send the followup PRs alongside tests?

bogdandrutu:
@odeke-em you fixed only the first comment with these issues, still the second is not addressed.

interceptor/opencensus/opencensus.go
odeke-em: Great catch @bogdandrutu, I've updated the code with the logic to use the last received and non-nil node, if the current message's Node is nil. PTAL.

interceptor/opencensus/opencensus.go
```diff
@@ -100,13 +98,35 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
...
 		recv, err := tes.Recv()
 		if err != nil {
 			return err
 		}
 
-		// Otherwise add these spans to the bundler
-		traceBundler.Add(recv.Spans, len(recv.Spans))
+		// If a Node hasn't been sent from downstream, use the last received node.
+		node := recv.Node
+		if node == nil {
```
odeke-em: @bogdandrutu @songy23 how does this look?

songy23: LGTM

bogdandrutu: I would write it a bit different:
```go
if recv.Node != nil {
  lastNonNilNode = recv.Node
}
// Otherwise add them to the bundler.
bundlerPayload := &spansAndNode{node: lastNonNilNode, spans: recv.Spans}
traceBundler.Add(bundlerPayload, len(bundlerPayload.spans))
```

odeke-em: Aha, nice reduction! I actually was going to write a variant of that but it was
too concise to not shock my future self and shout out the protocol.
I'll update that shortly though, thanks!

odeke-em: Updated, thank you and PTAL!

odeke-em:
Thank you @songy23 @bogdandrutu for the reviews!

Issue #54, opened 2018-10-02
Cleanup README.


bogdandrutu:
We have old text that is no longer relevant for the current version.

PR #56, opened 2018-10-02
README: re-organize the sections.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-service/issues/54.

As I read through README, I found most of the old contents (especially the initiatives) are still relevant, so I just slightly modified the order of sections.

/cc @rakyll

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,11 @@ to decide which service backend they want to export to, before they
 distribute their binary instrumented by OpenCensus.
 
 OpenCensus Service is trying to eliminate these requirements. With
-OpenCensus Service, users do not need to redeploy or restart their binary
+OpenCensus Service, users do not need to redeploy or restart their applications
 as long as it has the OpenCensus Agent exporter. All they need to do is
-just configuring and deploying OpenCensus Service, then it will automatically
-collect traces and metrics and export to any backend of users' choice.
+just configure and deploy OpenCensus Service separately. OpenCensus Service
+will then automatically collect traces and metrics and export to any
+backend of users' choice.
 
 Currently OpenCensus Service consists of two components,
 [OpenCensus Agent](#opencensus-agent) and [OpenCensus Collector](#opencensus-collector).
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# OpenCensus service
+# OpenCensus Service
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
-# OpenCensus service
+# OpenCensus Service
 
-OpenCensus service is an experimental component that collects and
+OpenCensus Service is an experimental component that collects and
 exports from the OpenCensus instrumented processes available from the
```
bogdandrutu: Mention the support for other formats, as well as the fact that we support multiple signals (metrics).

README.md
```diff
@@ -1,6 +1,6 @@
...
 same host machine.
```
bogdandrutu: This is not true anymore.

README.md
```diff
@@ -13,7 +13,9 @@ with OpenCensus, they have to modify their code to register an
 exporter and redeploy. Asking our users recompile and redeploy is
```
bogdandrutu: Also another case is if somebody wants to distribute a binary instrumented with OC which export should they use, with the agent they can distribute the binary with the agent exporter and the user will run the agent with the exporter of choice.

README.md
```diff
@@ -11,9 +13,18 @@ during an incident. Even though our users can benefit from having
...
 with OpenCensus, they have to modify their code to register an
 exporter and redeploy. Asking our users recompile and redeploy is
-not an ideal at an incident time.
+not an ideal at an incident time. In addition, currently users need
+to decide which service backend they want to export to, before they
+distribute their binary instrumented by OpenCensus.
 
-OpenCensus service is trying to eliminate this requirement.
+OpenCensus Service is trying to eliminate these requirements. With
+OpenCensus Service, users do not need to redeploy or restart their binary
```
odeke-em: restart their applications

README.md
```diff
@@ -11,9 +13,18 @@ during an incident. Even though our users can benefit from having
...
+as long as it has the OpenCensus Agent exporter. All they need to do is
+just configuring and deploying OpenCensus Service, then it will automatically
```
odeke-em: just configure and deploy the OpenCensus service separately. OpenCensus service will then
automatically collect traces...

PR #57, opened 2018-10-03
interceptor/opencensus: fix protocol misreadings

### LINKED ISSUE #51: interceptor/opencensus: the first message MAY include spans so collect them
My understanding of the protocol was that the very first message consists of the Node from the client, and after it has been accepted, then the client can send spans. However that was a mutual exclusivity misinterpretation, it MAY include spans. Corrected offline by @bogdandrutu.

### LINKED ISSUE #53: interceptor/opencensus: return error immediately if the first message does not have a node
Currently I am doing a retry waiting for the first node to be sent, 15 times. However, we should fail fast
https://github.com/census-instrumentation/opencensus-service/blob/e939042488a2fe395203c32612a90f208e2476a7/interceptor/opencensus/opencensus.go#L59-L78

Corrected offline by @bogdandrutu

odeke-em:
* Return immediately if the first message doesn't
have a non-nil Node. Previously we leniently
spun 15 times as a best case retry but that's
unnecessary
* If the first message is valid and has spans, those
spans MUST be consumed and NOT discarded.

Fixes #51
Fixes #53

### DIFF
--- a/interceptor/opencensus/opencensus.go
+++ b/interceptor/opencensus/opencensus.go
@@ -16,7 +16,6 @@ package ocinterceptor
 
 import (
 	"errors"
-	"fmt"
 	"time"
 
 	"google.golang.org/api/support/bundler"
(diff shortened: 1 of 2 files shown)


odeke-em:

PR #60, opened 2018-10-04
cmd/ocagent: added version command

### LINKED ISSUE #59: cmd/ocagent: provide version/help command
As we build and ship this agent, it is very useful that the binary provide users with some version information like:

* Git hash on which that cmd/ocagent was built
* Go version that was used to compile cmd/ocagent
* Operating system
* Architecture

These are useful for bug reports but will become even more useful if/when we ship out plugins for exporters(ability to dynamically link against exporter shared objects/DLLs), of which the plugins need to be compiled against the proper version of Go.

odeke-em:
Added a `version` command to the CLI which
will print out the information leading to when
and how the binary was built. The fields printed
out are:

* Version: The version of ocagent itself
* Goversion: The version of Go that compiled this binary
* OS: The OS it was built for (runtime.GOOS)
* Architecture: The architecture of the system
* GitHash: The hash of the ocagent commit

For example:
```shell
Version      0.0.1
GitHash      7ec3ea8
Goversion    devel +62e5215a2a Thu Oct 4 04:08:08 2018 +0000
OS           darwin
Architecture amd64
```

this will aid in:
* easy bug diagnoses and fixes
* plugin management whereby the exporter plugins MUST
match the version of Go that ocagent was built with inorder
for them to be application binary compatible
* reproducible builds

Also added a Makefile with a couple of directives
that allow for cross-compilation too.

Switched the flag parser from the inflexible "flag"
package in favour of "github.com/spf13/cobra"

Fixes #59


songy23:
Please fix the build.

odeke-em:
The Travis build failure doesn't make any sense: nothing is out of the blue in this PR but it fails to find dependencies 
screen shot 2018-10-04 at 12 53 54 pm

songy23:
I think the problem is `go.sum` and `go.mod`. Could you try splitting the PR into two: one with the command updates (README, Makefile, gitignore) and one with the rest of changes?

odeke-em:
Great, idea @songy23 I've removed go.mod and go.sum in https://travis-ci.org/census-instrumentation/opencensus-service/builds/437362301 and it is building alright. Thank you!

odeke-em:
Actually I've found out that the presence of a Makefile trips out Travis. I'll mark it as `DO NOT MERGE` for now and work on other things then circle back to it before release.

odeke-em:
Weird Travis failure, works alright locally. Let me restart the Travis build with a push.

odeke-em:

Issue #63, opened 2018-10-04
Instrument interceptors with OpenCensus


bogdandrutu:
## Trace
For trace, we start a Span every time we start a new batch (will end when we call the ReceiveSpans). Because the RPC is a bidirectional and can live long we need to decouple this span from the RPC span, but because they are related we should add the RPC span as parent link (here is a good example where links as useful).

## Stats
Measure:
* `oc.io/interceptor/received_spans` - long value

Tags:
* `opencensus_interceptor`

Measurement:
For `oc.io/interceptor/received_spans` we record a measurement every time we receive a new message from the client even if the number is 0.

Views:
We should install a distribution view with buckets {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 14, 16, 18, 20, 25, 30, 35, 40, 45, 50, 60, 70, 80, 90, 100, 150, 200, 250, 300, 450, 500, 600, 700, 800, 900, 1000, 1200, 1400, 1600, 1800, 2000}

bogdandrutu:
Please also install the ocgrpc plugin for the interceptor gRPC server.

odeke-em:
Great suggestions @bogdandrutu! Also I think in the future(after we've finished rolling out this instrumentation), would it make sense for us to extract information from the node? For example:
* process_identifier --> "opencensus_task" for stats and perhaps service name for metrics?

odeke-em:
>

Same thoughts, I actually have a TODO in the code as per
https://github.com/census-instrumentation/opencensus-service/blob/d91241ee30b42e86f7d439307acd3527ab544a93/interceptor/opencensus/trace_interceptor.go#L89-L91

but I think the even better way would be to create a helper method say in `internal` that creates a gRPC server for us since every interceptor that needs it will have to remember to install the ocgrpc plugin which could become error-prone.

bogdandrutu:
@odeke-em just having a todo to also add tags from the Node is fine for the moment.

odeke-em:
>@odeke-em just having a todo to also add tags from the Node is fine for the moment.

@bogdandrutu perfect. I've done that!

>Because the RPC is a bidirectional and can live long we need to decouple this span from the RPC span, but because they are related we should add the RPC span as parent link (here is a good example where links as useful).

[...]

PR #65, opened 2018-10-05
interceptor/opencensus: record number of received spans

odeke-em:
Instrument interceptor/opencensus with stats to record
the number of received spans. This is accomplished by
a helper to process spans that sends over spans even
if the number of spans is 0. Also record with tag_key
"opencensus_interceptor" whose value is the name
of the respective interceptor.

The test to ensure that 0 length spans are also added
is currently disabled because:
* Issue https://github.com/census-instrumentation/opencensus-go/issues/862
is not yet implemented which requests that the OpenCensus-Go
stats worker provide a method Flush to flush all data. Without
it, metrics will contain stale data from previous tests.

Add tests to lock-in this behavior.

Updates #63

### DIFF
--- a/interceptor/opencensus/opencensus.go
+++ b/interceptor/opencensus/opencensus.go
@@ -24,6 +24,7 @@ import (
 	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
 	agenttracepb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1"
 	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
+	"github.com/census-instrumentation/opencensus-service/internal"
 	"github.com/census-instrumentation/opencensus-service/spanreceiver"
 )
 
@@ -89,8 +90,20 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
 		return errTraceExportProtocolViolation
 	}
 
-	var lastNonNilNode *commonpb.Node
+	spansMetricsFn := internal.NewReceivedSpansRecorderStreaming(tes.Context(), "opencensus")
 
+	processReceivedSpans := func(ni *commonpb.Node, spans []*tracepb.Span) {
+		// Firstly, we'll add them to the bundler.
+		if len(recv.Spans) > 0 {
+			bundlerPayload := &spansAndNode{node: ni, spans: recv.Spans}
+			traceBundler.Add(bundlerPayload, len(bundlerPayload.spans))
+		}
+
+		// We MUST unconditionally record metrics from this reception.
+		spansMetricsFn(ni, recv.Spans)
+	}
+
+	var lastNonNilNode *commonpb.Node
 	// Now that we've got the first message with a Node, we can start to receive streamed up spans.
 	for {
 		// If a Node has been sent from downstream, save and use it.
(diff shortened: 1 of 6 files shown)


interceptor/opencensus/opencensus.go
```diff
@@ -88,20 +90,28 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
...
+	spansMetricsFn := internal.NewReceivedSpansRecorder(context.Background(), "opencensus")
+
+	processReceivedSpans := func(ni *commonpb.Node, spans []*tracepb.Span) {
+		// Firstly, we'll add them to the bundler.
+		if len(recv.Spans) > 0 {
+			bundlerPayload := &spansAndNode{node: ni, spans: recv.Spans}
+			traceBundler.Add(bundlerPayload, len(bundlerPayload.spans))
+		}
 
+		// We MUST Unconditionally record metrics from this reception.
```
bogdandrutu: s/Unconditionally/unconditionally

interceptor/opencensus/opencensus.go
```diff
@@ -88,20 +90,28 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
 		return errTraceExportProtocolViolation
 	}
 
-	var lastNonNilNode *commonpb.Node
+	spansMetricsFn := internal.NewReceivedSpansRecorder(context.Background(), "opencensus")
```
bogdandrutu: Use tags from the current context (they include tags emitted by the oc-library-agent-exporter)?

interceptor/opencensus/opencensus.go
odeke-em: Thanks, I'll update that.

interceptor/opencensus/opencensus.go
odeke-em: Great reminder, that'd entail changing that to `tes.Context()`, thanks.

bogdandrutu:
Probably forgot to push the commit. I cannot see any update.

odeke-em:
>Probably forgot to push the commit. I cannot see any update.

@bogdandrutu in deed, PTAL!

interceptor/opencensus/opencensus.go
```diff
@@ -89,20 +90,32 @@ func (oci *OCInterceptor) Export(tes agenttracepb.TraceService_ExportServer) err
...
 
-		// Otherwise add them to the bundler.
-		if len(recv.Spans) > 0 {
-			bundlerPayload := &spansAndNode{node: lastNonNilNode, spans: recv.Spans}
-			traceBundler.Add(bundlerPayload, len(bundlerPayload.spans))
-		}
+		// Note: In normal operations, the context of a streaming RPC usually doesn't
+		// change but we haven't examined all the corner cases for gRPC streaming
+		// hence we'll always pass in the context, in case the underlying implementation
+		// changes a detail. TODO (@odeke-em) investigate if streamer.Context() can change.
```
bogdandrutu: Cannot change and probably it shouldn't.

odeke-em: Thanks for the confirmation! I spent sometime yesterday digging through grpc-go to investigate if there any situations where this context can change but didn't seem to find any.

odeke-em: @bogdandrutu updated, PTAL.

odeke-em:

Issue #66, opened 2018-10-05
Please pass a context between components


bogdandrutu:
As a tracing project we should follow the best practices for tracing/monitoring:
* SpanReceiver passes the Context from the interceptor to the core component:
https://github.com/census-instrumentation/opencensus-service/blob/master/spanreceiver/spanreceiver.go#L24
* Exporter should accept a Context from the core component.

odeke-em:
Thanks for filing this issue @bogdandrutu!

For sure we can add these and a context as the first argument works to our own advantage.

However, perhaps we need a balance of applying those tips:

* SpanReceiver didn't take in a context.Context because the purpose of a SpanReceiver is to act as a blind sink but with the agent-core I can now see why we need a context -- makes sense to update this
* TraceExporter doesn't take in a context because like the OpenCensus-Go exporter, it's purpose is just to transform produced spans into SpanData that the various exporters are plugged into -- doesn't make sense to add a context.Context
* Interceptors take in a context as the first argument because they are receiving data from the wild, could use cancellation, could use propagation and handoff work to an intermediary that then writes spans to a spanreceiver

odeke-em:
I've updated SpanReceiver to take a context.Context with https://github.com/census-instrumentation/opencensus-service/pull/68, however for TraceExporter I am not convinced that it should a context.Context

bogdandrutu:
The reason to take a Context in the exporter is to be able to add tags in the core agent (or propagate tags from the interceptor) so that the metrics that we will record in the exporter about exported spans can be broken down by these tags.

odeke-em:
>The reason to take a Context in the exporter is to be able to add tags in the core agent (or propagate tags from the interceptor) so that the metrics that we will record in the exporter about exported spans can be broken down by these tags.

Wise decision, thank you @bogdandrutu!

PR #68, opened 2018-10-06
spanreceiver: update ReceiveSpans to take in a context.Context

odeke-em:
Updates #66


odeke-em:

Issue #47, opened 2018-10-01
proposal: cmd/ocagent/exporters: dynamic loading of exporters given configurations


odeke-em:
The current method of attack for enabling exporters is by configuring some sort of file that's fed into cmd/ocagent and coupled with statically declared imports e.g. in the raw imports before compiling it, match up against imported exporters. This could work for the mean time but we should have the ability to support even closed source exported and binary only exporters.

## Proposal
The main cmd/ocagent should be able to load in various exporters from shared object code that can then be fed the configurations and parse out their behavior then return to us an already enabled exporter.

### Interfaces that each binary only exporter should implement
Each binary only exporter should have a constructor:
   `func NewExporterWithConfig(config string) (interface{}, error)`

where the return argument `interface{}` is the result of how the exporter is configured or
how it has chosen to interpret the configuration. Each exporter can then decide whether to return a 
`trace.Exporter`, `view.Exporter` in whatever combination e.g.
either of these combinations:

   [`trace.Exporter`, `view.Exporter`, `trace.Exporter + view.Exporter`]

### Pros
This provides amazing flexibility but also allows closed source vendors of various capacities to use the agent if they don't want to make their proprietary exporter source code public or if perhaps they'd like to offer clients tiers of subscription but even more interestingly it can allow us to add and remove exporters dynamically just with a user interface e.g. during critical traffic floods

### Cons
Only works on Linux and macOS due to Go's plugin support. Security concerns from running plugins but that's a necessary risk to scale development and make the agent usable across various environments and deployments.

## Proof of concept
This code below, please paste these files as presented
* Makefile
```shell
so: vendor1_so vendor2_so

vendor1_so:
	go build -buildmode=plugin vendor1.go

vendor2_so:
	go build -buildmode=plugin vendor2.go
```

* vendor1.go
```go
package main

import (
	"log"

	"go.opencensus.io/trace"
)

var _ trace.Exporter = (*A)(nil)

func NewExporterFromConfig(configString string) (interface{}, error) {
	return new(A), nil
}

type A string

func (a *A) ExportSpan(sd *trace.SpanData) {
	log.Printf("B: %+v\n", sd)
}
```

* vendor2.go
```go
package main

import (
	"log"

	"go.opencensus.io/trace"
)

var _ trace.Exporter = (*B)(nil)

func NewExporterFromConfig(configString string) (interface{}, error) {
	return new(B), nil
}

type B int

[...]

PR #69, opened 2018-10-06
exporter: add Plugins interface as well as guide

odeke-em:
Added the TraceExporterPlugin interface of the type
```go
type TraceExporterPlugin interface {
        TraceExportersFromConfig(blob []byte, configType ConfigType) ([]TraceExporter, error)
        Stop(context.Context) error
}
```

as well as a guide and sample of an end-to-end written plugin.

This is the interface and introduction phase of fixing the
target issue. Actual plugin into `cmd/ocagent` will come
in the final change.

Updates #47


odeke-em:

Issue #61, opened 2018-10-04
distribution: create a Docker image of cmd/ocagent


odeke-em:
This issue is an action item for us to create a Docker image that we can pass around for library implementers and partners to be able to use the service/ocagent without having to build and deploy
the app themselves.

Thinking of how we'll distribute this image is also going to force us to think and prioritize how we'll support vendors plugging in their exporters so plugin mechanisms, and also create versions and milestones along with making official releases.

 Whenever ready perhaps we should publish this image on either of:
* DockerHub https://hub.docker.com/
* Google Cloud container registry https://cloud.google.com/container-registry/
and others...

odeke-em:
@songy23 @bogdandrutu did any of y'all wanna create the DockerHub or GCR account for OpenCensus?

odeke-em:
@songy23 I've created the starter Dockerfile in PR #71 and from that, what would be left is publishing then we can close this issue and iterate, having already published the Docker image.

odeke-em:
@rakyll you might be interested in this dockerization and publishing issue, as per interest in https://github.com/census-instrumentation/opencensus-service/issues/73. The file `build_binaries.sh` has the binaries and docker building directives, please feel free to adjust it.

songy23:
Blocked by https://github.com/census-instrumentation/opencensus-service/issues/82 and https://github.com/census-instrumentation/opencensus-service/issues/81, as we discussed today.

rakyll:
I was working on this :)

odeke-em:
I'll bump this to the next milestone.

songy23:
Friendly ping @rakyll - any update on this?

odeke-em:
Kindly paging @bogdandrutu @songy23 for a status on this issue.

rakyll:
"OpenCensus" is taken on Docker hub and I am not the owner, JFYI.

odeke-em:
Looks like someone already published some images from OpenCensus:
* occollector https://hub.docker.com/r/opencensus/occollector
* ocagent https://hub.docker.com/r/opencensus/ocagent
* koperator https://hub.docker.com/r/opencensus/koperator

If that was not someone from this team, that could be a security problem where someone is publishing images under opencensus' name.

pjanotti:
@bogdandrutu created those but there is nothing published yet.

flands:
This will be addressed in the OpenTelemetry Service

PR #71, opened 2018-10-07
build: added Dockerfile + binaries generator bash script

odeke-em:
Updates #61


odeke-em:

PR #77, opened 2018-10-09
cmd/ocagent: OpenCensus interceptor bind address configurable via config.yaml

### LINKED ISSUE #23: cmd/ocagent: allow configuration file to specify OpenCensus interceptor port
The configuration file needs to be updated to support custom host:port to start the agent server .

```
listen: "localhost:777"

exporters:
    stackdriver:
        project: "project-id"
        enableTraces: true

    zipkin:
```

odeke-em:
Use the config.yaml file to configure the OpenCensus
interceptor listener's address, instead of using commandline flags.
This is because all the other configurations are performed
using the config.yaml file so an extra option is just confusing.

The default port to run on is 55678 if unspecified, but otherwise
YAML parseable by key `opencensus_interceptor`, and then using the
field name `address`. For example:

```yaml
opencensus_interceptor:
  address: ":55777"
```

and when run with
```shell
ocagent
```

which will then produce:
```shell
2018/10/08 16:36:47 Running OpenCensus interceptor as a gRPC service at "localhost:55777"
```

Fixes #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,18 +160,21 @@ backend (e.g Jaeger Thrift Span), and then push them to corresponding backend or
 
 First, install ocagent if you haven't.
 
-```
+```shell
 $ go get github.com/census-instrumentation/opencensus-service/cmd/ocagent
 ```
 
+### Configuration file
+
 Create a config.yaml file in the current directory and modify
-it with the exporter configuration. For example, following
-configuration exports both to Stackdriver and Zipkin.
+it with the exporter and interceptor configurations.
 
 
-config.yaml:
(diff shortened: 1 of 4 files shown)


odeke-em:
>I think we also want the host name to be configurable.

I don't follow. That's going to bind on a local port and spin up the service, as opposed to being an address that's already available that the service was running on. I believe the only options available here are "localhost" and "127.0.0.1" of which we already have those internally.

odeke-em:
@songy23 PTAL!

odeke-em:

PR #78, opened 2018-10-09
exporter.TraceExporter: pass a context into ExportSpanData

odeke-em:
Updates #66


odeke-em:

PR #80, opened 2018-10-10
Update the Agent architecture graph.

songy23:
Before:
agent-architecture-old

After:
agent-architecture

Thanks @shahprit for updating it.


odeke-em:
PS, there is a typo, it should be `Jaeger` instead of `Jeager`.

songy23:
@odeke-em Good catch! I just fixed the typo.

PR #93, opened 2018-10-16
all: move exporterparser out of cmd/ocagent

### LINKED ISSUE #92: cmd/ocagent/exportparser: update exporters to be exporter.TraceExporter
Currently the code in cmd/ocagent/exportparser produces exporters that assumed we'd use the old-style OpenCensus exporters directly. In this new style, we pass in a Node, a context and the various backends can do whatever they'd like with the context and Node.

I also just discovered that we need to get this in before instrumenting exporters.

odeke-em:
Moved exporterparser code out of:
    cmd/ocagent/exporterparser
and into
    exporter/exporterparser

because:
a) Makes it easier to test and verify behavior
b) Makes it easier to instrument the various exporters
c) We shouldn't be adding sub-packages in cmd/ocagent
which produces the binary itself

Also changed all the currently available exporters
to return []exporter.TraceExporter since multiple configurations
could be given for an exporter e.g. with Datadog or Stackdriver

Fixes #92


odeke-em:
Thank  you @songy23!

Issue #64, opened 2018-10-04
Instrument exporter with OpenCensus


bogdandrutu:
## Stats
Measure:
* `oc.io/interceptor/exported_spans` - long value

Tags:
* `opencensus_exporter` - the name of the exporter.

Measurement:
For `oc.io/interceptor/exported_spans` we record a measurement every time we export Spans.

Views:
We should install a distribution view with buckets {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 14, 16, 18, 20, 25, 30, 35, 40, 45, 50, 60, 70, 80, 90, 100, 150, 200, 250, 300, 450, 500, 600, 700, 800, 900, 1000, 1200, 1400, 1600, 1800, 2000}

odeke-em:
Blocked by https://github.com/census-instrumentation/opencensus-service/pull/93 but it becomes really trivial after that PR.

odeke-em:
Alright, the metrics for exported spans added in #109 per exporter

odeke-em:
Metrics added, am not sure if we should add traces. I'll close this issue but please feel free to reopen @bogdandrutu.

PR #94, opened 2018-10-16
exporter: add metrics recording helper for Exporters

odeke-em:
Updates #64


odeke-em:
Thank you @helly25 @songy23 for the review!

PR #96, opened 2018-10-16
README: add table of contents and link items

odeke-em:
Useful for navigating around the project but
also to have unique links.


odeke-em:

PR #97, opened 2018-10-16
README: add section for building binaries and Docker image

odeke-em:
Instructions of how to build binaries properly as
well as how to build a Docker image.

By "properly" I mean, a binary that contains the GitHash
of what code was used to build the ocagent binary but
also how to cross-compile and create other binaries using
the provided helper script.


odeke-em:

PR #106, opened 2018-10-17
interceptor/opencensus: skip TestEnsureRecordedMetrics

odeke-em:
`go test -race` produces a slowdown in the program by
between 2-20X yet this test depends on precise timings
in the OpenCensus-Go stats worker. Disable it as it is
very flaky and causing unrelated failures in PRs.


odeke-em:
Thank you @helly25!

PR #109, opened 2018-10-18
exporter/exporterparse: metrics for exported spans

odeke-em:
After each exporter has exported spans, record the
number of spans exported with tag:
    "opencensus_exporter": <name>

Updates #64


odeke-em:
Thank you @helly25!

PR #110, opened 2018-10-18
cmd/ocagent: added zPages for overall ocagent process

### LINKED ISSUE #83: Enable zpages for Agent
This is not about building zpages for the intercepted spans, it is about enabling zpages for the ocagent process.

Should be enabled on the same port as the one we use in oc interceptor.

odeke-em:
Added zPages as the diagnostic helper for the overall
ocagent process itself #inception

By default the zPages run at ":55679/debug". They can be
disabled in the config.yaml file by using the attribute
  `disabled`

```yaml
zpages:
   disabled: true
   port: 8888
```

Also added as an attribute the number of exported spans
for every internal interceptor span.

Fixes #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@
         - [Interceptors](#agent-impl-interceptors)
         - [Agent Core](#agent-impl-agent-core)
         - [Exporters](#agent-impl-exporters)
+    - [Diagnostics](#agent-diagnostics)
+        - [zPages](#agent-zpages)
     - [Building binaries](#agent-building-binaries)
     - [Usage](#agent-usage)
     - [Configuration file](#agent-configuration-file)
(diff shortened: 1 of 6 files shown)


odeke-em:
#### zPages exhibits
screen shot 2018-10-17 at 10 22 05 pm

screen shot 2018-10-17 at 10 24 16 pm

screen shot 2018-10-17 at 10 24 48 pm

odeke-em:
Thanks for the reviews @helly25 and @songy23!

odeke-em:
>Note that there're a bunch of changes to the color and style of zPages. I've filed census-instrumentation/opencensus-go#949 to OC-Go to update it.

Cool, when those are updated in OpenCensus-Go, we'll just need to get the latest. Thanks @songy23!

odeke-em:
@bogdandrutu as per my comment in https://github.com/census-instrumentation/opencensus-service/issues/83#issuecomment-431125662
>Gotcha. In deed, gRPC supports multiplexing on the same port. The only problem is that if running locally without a TLS enabled connection the detection for multiplexing that they recommend is quite buggy and doesn't work and they also spell that out in their docs.
https://godoc.org/google.golang.org/grpc#Server.ServeHTTP
>
```
if r.ProtoMajor == 2 && strings.HasPrefix(
	r.Header.Get("Content-Type"), "application/grpc") {
	grpcServer.ServeHTTP(w, r)
} else {
	yourMux.ServeHTTP(w, r)
}
It doesn't seem to set the "Content-Type" header to "application/grpc"
using only r.ProtoMajor == 2 doesn't work well in case someone decides t
and that's going to be for majority of the cases for people deploying the agent locally or within local networks.
```

The most common deployment is going to have issues sharing the zPages port with the OCInterceptor, so a separate port 55679. However, I'll open up a separate issue for us to discuss if we can actually share that port. Thanks.

PR #111, opened 2018-10-18
interceptor: add zipkin /v2 interceptor

### LINKED ISSUE #24: trace/interceptors: define and implement Zipkin interceptor for "/v2" API
### Definition
An interceptor is a swappable segment that accepts content from a secondary source and then translates that accepted content into the agent's consumable data model.

### Usage
The use of an interceptor would be for code that already exports to Zipkin on a cluster: if we wanted to use OpenCensus, we could accept all the trace traffic from the Zipkin source and translate it into the OpenCensus data model. This is the reverse of an exporter.

To accomplish this, we'll have to study the Zipkin data model at https://zipkin.io/zipkin-api/#/default/post_spans

and then choose either of:
a) hijack the default Zipkin server or accept raw Zipkin traffic
b) Query running Zipkin servers just like the Zipkin UI at localhost:9411/ does

### Evaluating the options:
For a) if we hijacked a Zipkin server, this would be frowned upon but also severly disrupt operations of large companies that are content with their Zipkin deployments but also by the project maintainers. Also if we are to act as a Zipkin server, that would mean that we have to implement a fully compliant Zipkin server

For b) querying running Zipkin servers is a less invasive way and allows gradual roll out of the service say across an indefinite time horizon e.g. 20 months where a company can still use their Zipkin deployment but the agent can ingest traffic after querying it -- this option is less disruptive and more amicable to the Zipkin project, engineering teams and generally to avoid drastic changes. The con here is that engineers have to know that the OpenCensus agent is pulling traces from Zipkin.

### Course of action
Implementing b) as the default is the best course of action as per the evaluation above -- less disruption, less invasive, less code maintenance. However, we should be able to accept traffic as if it were coming directly from a Zipkin instrumented backend.

Our solution will be a hybrid

odeke-em:
Add Zipkin as a trace interceptor
Hooked the interceptor into cmd/ocagent
with a reporter by default "localhost:9412/api/v2/spans"

Fixes #24

### DIFF
--- a/cmd/ocagent/config.go
+++ b/cmd/ocagent/config.go
@@ -150,20 +150,20 @@ func (c *config) checkLogicalConflicts(blob []byte) error {
 	zExporterAddr := zc.EndpointURL()
 	zExporterURL, err := url.Parse(zExporterAddr)
 	if err != nil {
-		return fmt.Errorf("parsing exporter address %q got error: %v", zExporterAddr, err)
+		return fmt.Errorf("parsing ZipkinExporter address %q got error: %v", zExporterAddr, err)
 	}
 
 	zInterceptorHostPort := c.zipkinInterceptorAddress()
 
 	zExporterHostPort := zExporterURL.Host
 	if zInterceptorHostPort == zExporterHostPort {
-		return fmt.Errorf("ZipkinInterceptor address (%q) is the same as the interceptor address (%q)",
+		return fmt.Errorf("ZipkinExporter address (%q) is the same as the interceptor address (%q)",
 			zExporterHostPort, zInterceptorHostPort)
 	}
 	zExpHost, zExpPort, _ := net.SplitHostPort(zExporterHostPort)
 	zInterceptorHost, zInterceptorPort, _ := net.SplitHostPort(zExporterHostPort)
 	if eqHosts(zExpHost, zInterceptorHost) && zExpPort == zInterceptorPort {
-		return fmt.Errorf("ZipkinInterceptor address (%q) aka (%s on port %s)\nis the same as the interceptor address (%q) aka (%s on port %s)",
+		return fmt.Errorf("ZipkinExporter address (%q) aka (%s on port %s)\nis the same as the interceptor address (%q) aka (%s on port %s)",
 			zExporterHostPort, zExpHost, zExpPort, zInterceptorHostPort, zInterceptorHost, zInterceptorPort)
 	}
 
@@ -171,7 +171,7 @@ func (c *config) checkLogicalConflicts(blob []byte) error {
 	zExpIPAddr, _ := net.ResolveIPAddr("ip", zExpHost)
 	zInterceptorIPAddr, _ := net.ResolveIPAddr("ip", zInterceptorHost)
 	if zExpIPAddr != nil && zInterceptorIPAddr != nil && reflect.DeepEqual(zExpIPAddr, zInterceptorIPAddr) {
-		return fmt.Errorf("ZipkinInterceptor address (%q) aka (%+v)\nis the same as the\ninterceptor address (%q) aka (%+v)",
+		return fmt.Errorf("ZipkinExporter address (%q) aka (%+v)\nis the same as the\ninterceptor address (%q) aka (%+v)",
 			zExporterHostPort, zExpIPAddr, zInterceptorHostPort, zInterceptorIPAddr)
 	}
 	return nil
--- a/cmd/ocagent/config.go
+++ b/cmd/ocagent/config.go
@@ -42,8 +42,9 @@ import (
 //      port: 55679
 
 const (
-	defaultOCInterceptorAddress = "localhost:55678"
-	defaultZPagesPort           = 55679
+	defaultOCInterceptorAddress     = "localhost:55678"
+	defaultZPagesPort               = 55679
+	defaultZipkinInterceptorAddress = "localhost:9411"
 )
 
 type config struct {
@@ -52,7 +53,8 @@ type config struct {
 }
 
 type interceptors struct {
-	OpenCensusInterceptor *interceptorConfig `yaml:"opencensus"`
+	OpenCensus *interceptorConfig `yaml:"opencensus"`
+	Zipkin     *interceptorConfig `yaml:"zipkin"`
 }
 
 type interceptorConfig struct {
@@ -70,10 +72,10 @@ func (c *config) ocInterceptorAddress() string {
 		return defaultOCInterceptorAddress
 	}
 	inCfg := c.Interceptors
-	if inCfg.OpenCensusInterceptor == nil || inCfg.OpenCensusInterceptor.Address == "" {
+	if inCfg.OpenCensus == nil || inCfg.OpenCensus.Address == "" {
 		return defaultOCInterceptorAddress
 	}
-	return inCfg.OpenCensusInterceptor.Address
+	return inCfg.OpenCensus.Address
 }
 
 func (c *config) zPagesDisabled() bool {
@@ -94,6 +96,24 @@ func (c *config) zPagesPort() (int, bool) {
 	return port, true
 }
 
+func (c *config) zipkinInterceptorEnabled() bool {
+	if c == nil {
+		return false
+	}
+	return c.Interceptors != nil && c.Interceptors.Zipkin != nil
+}
+
+func (c *config) zipkinInterceptorAddress() string {
+	if c == nil || c.Interceptors == nil {
+		return defaultZipkinInterceptorAddress
+	}
+	inCfg := c.Interceptors
+	if inCfg.Zipkin == nil {
+		return defaultZipkinInterceptorAddress
+	}
+	return inCfg.Zipkin.Address
+}
+
 func parseOCAgentConfig(yamlBlob []byte) (*config, error) {
 	var cfg config
 	if err := yaml.Unmarshal(yamlBlob, &cfg); err != nil {
@@ -101,8 +121,6 @@ func parseOCAgentConfig(yamlBlob []byte) (*config, error) {
 	}
 	return &cfg, nil
 }
-
-type exporterParser func(yamlConfig []byte) (te []exporter.TraceExporter, err error)
 
 // exportersFromYAMLConfig parses the config yaml payload and returns the respective exporters
 func exportersFromYAMLConfig(config []byte) (traceExporters []exporter.TraceExporter, doneFns []func() error) {
@@ -128,6 +146,13 @@ func exportersFromYAMLConfig(config []byte) (traceExporters []exporter.TraceExpo
 				nonNilExporters += 1
 			}
 		}
+
+		// TODO: (@odeke-em) Add a catch here prevent logical errors
+		// e.g. if the Zipkin interceptor port is the exact
+		// same as that of the exporter, DO NOT run, crash hard
+		// lest we'll just recursively send traffic back to selves
+		// which is a potential self-DOS.
+
 		if nonNilExporters > 0 {
 			pluralization := "exporter"
 			if nonNilExporters > 1 {
--- a/cmd/ocagent/config.go
+++ b/cmd/ocagent/config.go
@@ -15,7 +15,12 @@
 package main
 
 import (
+	"fmt"
 	"log"
+	"net"
+	"net/url"
+	"reflect"
+	"strings"
 
 	yaml "gopkg.in/yaml.v2"
 
@@ -42,9 +47,8 @@ import (
 //      port: 55679
 
 const (
-	defaultOCInterceptorAddress     = "localhost:55678"
-	defaultZPagesPort               = 55679
-	defaultZipkinInterceptorAddress = "localhost:9411"
+	defaultOCInterceptorAddress = "localhost:55678"
+	defaultZPagesPort           = 55679
 )
 
 type config struct {
@@ -105,11 +109,11 @@ func (c *config) zipkinInterceptorEnabled() bool {
 
 func (c *config) zipkinInterceptorAddress() string {
 	if c == nil || c.Interceptors == nil {
-		return defaultZipkinInterceptorAddress
+		return exporterparser.DefaultZipkinEndpointHostPort
 	}
 	inCfg := c.Interceptors
-	if inCfg.Zipkin == nil {
-		return defaultZipkinInterceptorAddress
+	if inCfg.Zipkin == nil || inCfg.Zipkin.Address == "" {
+		return exporterparser.DefaultZipkinEndpointHostPort
 	}
 	return inCfg.Zipkin.Address
 }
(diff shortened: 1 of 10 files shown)


cmd/ocagent/config.go
```diff
@@ -23,10 +23,14 @@ import (
 	"github.com/census-instrumentation/opencensus-service/exporter/exporterparser"
 )
 
-const defaultOCInterceptorAddress = "localhost:55678"
+const (
+	defaultOCInterceptorAddress     = "localhost:55678"
+	defaultZipkinInterceptorAddress = "localhost:9412"
```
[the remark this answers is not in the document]
odeke-em: Great question: so if we make it 9411, then after the agent exports to the actual Zipkin at 9411, we'll just make a recursive export back to the interceptor and never ending :)

adriancole:
why would you ever have this and zipkin on the same host? if you had zipkin
wouldn't you just send spans there?

odeke-em:
>why would you ever have this and zipkin on the same host? if you had zipkin
wouldn't you just send spans there?

@adriancole great question: currently am running all this stuff locally on the same machine and using Zipkin as my final backend hence the superficial problem that you've just called out. So sure, we can just intercept on 9411.

adriancole:
in your local test you can reassign zipkin with QUERY_PORT=9412 for testing
https://github.com/openzipkin/zipkin/tree/master/zipkin-server#environment-variables

odeke-em:
>in your local test you can reassign zipkin with QUERY_PORT=9412 for testing
https://github.com/openzipkin/zipkin/tree/master/zipkin-server#environment-variables

Thanks @adriancole!
Btw, I am now using `9411` so that it is a drop-in and I've verified that the final data successfully
has unique servicenames and attributes.

odeke-em:
Alright I think we are gucci to review, I've been able to run the Zipkin interceptor to the agent from a JDBC application from https://opencensus.io/guides/integrations/sql/java_sql/
screen shot 2018-10-23 at 9 38 45 pm

cmd/ocagent/config.go
```diff
@@ -102,7 +126,72 @@ func parseOCAgentConfig(yamlBlob []byte) (*config, error) {
...
+	zExporterURL, err := url.Parse(zExporterAddr)
+	if err != nil {
+		return fmt.Errorf("parsing exporter address %q got error: %v", zExporterAddr, err)
+	}
+
+	zInterceptorHostPort := c.zipkinInterceptorAddress()
+
+	zExporterHostPort := zExporterURL.Host
+	if zInterceptorHostPort == zExporterHostPort {
+		return fmt.Errorf("ZipkinInterceptor address (%q) is the same as the interceptor address (%q)",
```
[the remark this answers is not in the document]
odeke-em: Nice, thanks, I'll update that.

interceptor/zipkin/trace_interceptor.go
```diff
@@ -0,0 +1,395 @@
...
+func New(sr spanreceiver.SpanReceiver) (*ZipkinInterceptor, error) {
+	return &ZipkinInterceptor{spanSink: sr}, nil
+}
+
+func (zi *ZipkinInterceptor) StartTraceInterception(ctx context.Context, spanSink spanreceiver.SpanReceiver) error {
+	zi.spanSink = spanSink
+	return nil
+}
+
+func (zi *ZipkinInterceptor) parseAndConvertToTraceSpans(jsonBlob []byte) (reqs []*agenttracepb.ExportTraceServiceRequest, err error) {
```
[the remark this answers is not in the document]
odeke-em: Yes I believe so as per https://zipkin.io/zipkin-api/#/default/post_spans. They upload with various "Content-Encodings" though e.g. "gzip" with the Zipkin-Java client and none for the Zipkin-Go client. /cc @adriancole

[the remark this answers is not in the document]
odeke-em: For a correction, I meant for HTTP uploaded spans, yes and as per the v2 API. There is also Thrift compatible API /cc @adriancole @basvanbeek to kindly ask if we should be supporting the Thrift sent spans and perhaps some reference material for this.

cmd/ocagent/main.go
```diff
@@ -141,3 +159,28 @@ func runOCInterceptor(addr string, sr spanreceiver.SpanReceiver) (doneFn func()
...
+	zi, err := zipkininterceptor.New(sr)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create the Zipkin interceptor: %v", err)
+	}
+
+	ln, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, fmt.Errorf("Cannot bind Zipkin interceptor to address %q: %v", addr, err)
+	}
+	zipkinRoute := "/api/v2/spans"
```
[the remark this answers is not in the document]
odeke-em: Roger that, thanks, I'll update that.

cmd/ocagent/main.go
```diff
@@ -141,3 +159,28 @@ func runOCInterceptor(addr string, sr spanreceiver.SpanReceiver) (doneFn func()
...
+	mux := http.NewServeMux()
+	mux.Handle(zipkinRoute, zi)
+	go func() {
+		fullAddr := addr + zipkinRoute
+		log.Printf("Running the Zipkin interceptor at %q", fullAddr)
+		if err := http.Serve(ln, mux); err != nil {
```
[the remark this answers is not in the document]
odeke-em: We handle signal reception in https://github.com/census-instrumentation/opencensus-service/blob/80f62ef4e8063f0b2d83fbc6d8472ff8571daa11/cmd/ocagent/main.go#L95-L109

However, I think we could dig deeper into the clean exits in a separate issue as that's a general aesthetic than focused on Zipkin :)

exporter/exporterparser/zipkin.go
```diff
@@ -66,22 +95,99 @@ func ZipkinExportersFromYAML(config []byte) (tes []exporter.TraceExporter, doneF
 	if zc.LocalEndpointURI != "" {
 		localEndpointURI = zc.LocalEndpointURI
 	}
-	// TODO(jbd): Propagate hostport and more metadata from each node.
+	endpoint := zc.EndpointURL()
+	zle, err := newZipkinExporter(endpoint, serviceName, localEndpointURI)
```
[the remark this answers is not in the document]
odeke-em: Yes, it previously used copy+paste from the Zipkin-Go exporter code and examples. This new change can use an empty localEndpointURI. Thanks for the confirmation of the serviceName having to be non-empty, I encountered it while writing tests.

interceptor/zipkin/trace_interceptor.go
```diff
@@ -0,0 +1,395 @@
...
+
+	pr := processBodyIfNecessary(r)
+	slurp, err := ioutil.ReadAll(pr)
+	if c, ok := pr.(io.Closer); ok {
+		_ = c.Close()
+	}
+	_ = r.Body.Close()
+	ereqs, err := zi.parseAndConvertToTraceSpans(slurp)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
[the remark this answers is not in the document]
odeke-em: Great reminder, thanks! I plan on doing that as well for other traces.

interceptor/zipkin/trace_interceptor.go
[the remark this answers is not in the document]
odeke-em: Awesome, thank you for the confirmation @basvanbeek between the differences in /v2 and /v1. In deed, the minimum requirement was to support only "/v2". However for maximum compatibility, perhaps in a follow-up PR I'll submit a change supporting Thrift

cmd/ocagent/config.go
```diff
@@ -102,7 +126,72 @@ func parseOCAgentConfig(yamlBlob []byte) (*config, error) {
 	return &cfg, nil
 }
 
-type exporterParser func(yamlConfig []byte) (te []exporter.TraceExporter, err error)
+// The goal of this function is to catch logical errors such as
+// if the Zipkin interceptor port conflicts with that of the exporter
+// lest we'll have a self DOS because spans will be exported "out" from
+// the exporter, yet be received from the interceptor, then sent back out
+// and back in a never ending loop.
+func (c *config) checkLogicalConflicts(blob []byte) error {
```
pjanotti: Possibly a good candidate for backlog: it seems that the intent is for this to be a general check of consistency on the config but right now it just addresses the zipkin interceptor (unique case so far I guess). We are also doing the unmarshal in various places (while trivial it is nice to avoid it if we can), and as it grows it perhaps deserves its own package so it can be easily shared.

If others agree we can add this to the backlog. /cc @flands

exporter/exporterparser/zipkin.go
```diff
@@ -27,20 +28,51 @@ import (
...
 }
 
+// zipkinExporter is a multiplexing exporter that spawns a new OpenCensus-Go Zipkin
+// exporter per unique node encountered. This is because serviceNames per node define
+// unique services, alongside their IPs. Also it is useful to intercept traffic from
+// Zipkin servers and then transform them back to the final form when creating an
+// OpenCensus spandata.
 type zipkinExporter struct {
-	exporter *zipkin.Exporter
+	mu sync.RWMutex
```
pjanotti: It will be nice to have a comment of it is going to be protected by the mutex.

exporter/exporterparser/zipkin.go
```diff
@@ -66,22 +95,99 @@ func ZipkinExportersFromYAML(config []byte) (tes []exporter.TraceExporter, doneF
...
+	zle.shutdownFns = append(zle.shutdownFns, reporter.Close)
+
+	return zle, nil
+}
+
+// exporterForNode firstly tries to find a memoize OpenCensus-Go Zipkin exporter
+// appropriate for this node. If it doesn't find any, it returns the default exporter
+// that was created at start-time.
+func (ze *zipkinExporter) exporterForNode(node *commonpb.Node) *zipkin.Exporter {
+	ze.mu.Lock()
```
pjanotti: The initial checks don't seem to need be under a writer lock. From what I see it is intended to protect ze.serviceNameToExporter, any reason to not attemp a read lock first and only after that failing trying a writer lock (in the minimal scope possible)?

interceptor/zipkin/trace_interceptor.go
```diff
@@ -0,0 +1,399 @@
...
+	// Trace this method
+	ctx, span := trace.StartSpan(context.Background(), "ZipkinInterceptor.Export")
+	defer span.End()
+
+	// If the starting RPC has a parent span, then add it as a parent link.
+	parentCtx := r.Context()
+	internal.SetParentLink(parentCtx, span)
+
+	pr := processBodyIfNecessary(r)
+	slurp, err := ioutil.ReadAll(pr)
```
pjanotti: It seems advisable to defend against bad data here.

interceptor/zipkin/trace_interceptor.go
```diff
@@ -0,0 +1,399 @@
...
+func (zi *ZipkinInterceptor) StopTraceInterception(ctx context.Context) error {
+	return nil
+}
+
+// processBodyIfNecessary checks the "Content-Encoding" HTTP header and if
+// a compression such as "gzip", "deflate", "zlib", is found, the body will
+// be uncompressed accordingly or return the body untouched if otherwise.
+// Clients such as Zipkin-Java do this behavior e.g.
+//    send "Content-Encoding":"gzip" of the JSON content.
+func processBodyIfNecessary(req *http.Request) io.Reader {
```
pjanotti: nit: naming, this is about selecting the io.Reader according to encoding not actually processing the body

cmd/ocagent/config.go
songy23: Make sense to me, please file an issue to keep track of this.

interceptor/zipkin/trace_interceptor.go
```diff
@@ -0,0 +1,399 @@
...
+	// for grouping within a map, we'll use the .String() value
+	byNodeGrouping := make(map[string][]*tracepb.Span)
+	nodes := make([]*commonpb.Node, 0, len(zipkinSpans))
+	// Now translate them into tracepb.Span
+	for _, zspan := range zipkinSpans {
+		span, node, err := zipkinSpanToTraceSpan(zspan)
+		// TODO:(@odeke-em) record errors
+		if err == nil && span != nil {
+			key := node.String()
+			if _, firstTime := byNodeGrouping[key]; firstTime {
```
pjanotti: This is appending nodes with same key after the first time that the key is processed. So if the key appears only once it won't be present on nodes, which gives the impression that it should have been: 
```go
if _, visited := byNodeGrouping[key]; !visited {
```

adriancole:
@odeke-em your screenshot was stackdriver... :) was that a mistake or are you proxying stackdriver with zipkin via our gcp build?

cmd/ocagent/config.go
pjanotti: Done, here it is issue https://github.com/census-instrumentation/opencensus-service/issues/135

odeke-em:
>@odeke-em your screenshot was stackdriver... :) was that a mistake or are you proxying stackdriver with zipkin via our gcp build?

@adriancole I intercepted spans from a Java application that uses Zipkin and exported to Stackdriver since it was an available exporter not on 9411 :)

interceptor/zipkin/trace_interceptor.go
odeke-em: @pjanotti thanks, it actually will gunzip or zlib deflate the body so it actually processes the body if necessary.
screen shot 2018-10-25 at 11 08 08 am

interceptor/zipkin/trace_interceptor.go
odeke-em: What do you mean? We read all the data into memory and then JSON unmarshal it. Are you suggesting say a max limit? If so, net/http already handles that and has tests about content-length mismatches.

exporter/exporterparser/zipkin.go
odeke-em: >The initial checks don't seem to need be under a writer lock.

This call is only used dynamically during every call to trace.ExportSpan while newZipkinExporter is created once at startTime and also the shared entity hasn't yet been created. That's why we need a lock here and not there. For the other checks we also have locks.

>From what I see it is intended to protect ze.serviceNameToExporter,

It also protects `.stop`
screen shot 2018-10-25 at 11 13 50 am
and is `serviceNameToExporter` is used by `ExportSpanData` so all its methods are protected by the same lock.

>any reason to not attemp a read lock first and only after that failing trying a writer lock (in the minimal scope possible)?

Great question: a readLock then a write lock are going to complicate the logic at the expense of readability and perhaps add more code because an error will need read unlocking on a separate error message. I think if we deem this to be in a very hot path, then we can apply that.

exporter/exporterparser/zipkin.go
odeke-em: Great idea, thanks.

cmd/ocagent/config.go
odeke-em: Thanks for filing it @pjanotti I was thinking the same thing a couple of days back but was stuck at a conference.

interceptor/zipkin/trace_interceptor.go
odeke-em: Great catch, thanks @pjanotti!

exporter/exporterparser/zipkin.go
odeke-em: Done, thanks @pjanotti!

interceptor/zipkin/trace_interceptor.go
pjanotti: Ah, I was incorrectly assuming that they were completely lazy...

interceptor/zipkin/trace_interceptor.go
pjanotti: My assumption about the "decompressing" readers gave the impression that decompression could fail here.

interceptor/zipkin/trace_interceptor.go
odeke-em: Great suggestion though to be conservative with the naming :)

odeke-em:
@basvanbeek you mentioned you'd like to play around with this for a bit, please go ahead. In the meantime we'll work on merging this and finally cutting a v0.0.1 release. If we need to make updates, we can do that as well.

Thank you everyone for the reviews.

interceptor/zipkin/trace_interceptor.go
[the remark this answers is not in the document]
odeke-em: Btw, this was updated yesterday in code but Github is showing in this conversation a stale snippet.

interceptor/zipkin/trace_interceptor.go
```diff
@@ -0,0 +1,395 @@
...
+		Kind:         zipkinSpanKindToProtoSpanKind(zs.Kind),
+		Status:       extractProtoStatus(zs),
+		Attributes:   zipkinTagsToTraceAttributes(zs.Tags),
+		TimeEvents:   zipkinAnnotationsToProtoTimeEvents(zs.Annotations),
+	}
+
+	return pbs, node, nil
+}
+
+func nodeFromZipkinEndpoints(zs *zipkinmodel.SpanModel) *commonpb.Node {
```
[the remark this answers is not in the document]
odeke-em: @basvanbeek the other side doesn't have the concept of a node and I actually do that in this PR as per https://github.com/census-instrumentation/opencensus-service/blob/67df1689f6f364be292f9855c84e03d4bde47e30/exporter/exporterparser/zipkin.go#L141-L177

The only piece remaining is an issue that I've filed on OpenCensus-Go as per https://github.com/census-instrumentation/opencensus-go/issues/959 which will then ensure that we can transparently convert a Zipkin v2 span to our intermediary span and back out and have the exact same data :)

exporter/exporterparser/zipkin.go
pjanotti: Thanks @odeke-em - it seems reasonable: contention is expected to be low and we can improve if we detect the need when stress testing.

One small note: since `mu` just uses Lock()/Unlock() it will be clear if it is a `sync.Mutex` instead of a `sync.RWMutex`.

odeke-em: @pjanotti, gotcha thanks I'll update it to `sync.Mutex` instead of `sync.RWMutex`.

odeke-em: And done, thanks @pjanotti

odeke-em:
Thanks everyone for the reviews!

cmd/ocagent/main.go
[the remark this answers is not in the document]
basvanbeek: yes it was meant as generic remark showing this particular item as an example of unclean shutdown which we should fix project wide.

interceptor/zipkin/trace_interceptor.go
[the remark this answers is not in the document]
basvanbeek: If adding anything, I'd rather see us focus on adding Zipkin protobuf support.

https://github.com/openzipkin/zipkin-api/blob/master/zipkin.proto

PR #118, opened 2018-10-18
config.yaml: categorize and reorganize YAML file

### LINKED ISSUE #90: Reconsider the config file layout
Config file requires some hierarchal structure:

```
interceptors:
    opencensus: ...

exporters:
    zipkin: ...
    stackdriver: ....
```

odeke-em:
Categorize namespaces of each item into:

- exporters
- interceptors
- zpages

which aesthetically looks nicer than the original but also
gives users better context and control over what the agent
does.

For example here is the new configuration file:
```yaml
interceptors:
    opencensus:
        address: "127.0.0.1:55678"

exporters:
    stackdriver:
        project: "project-id"
        enable_tracing: true

    zipkin:

    datadog:
        enable_tracing: false

zpages:
    port: 55679
```

Fixes #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,12 +261,13 @@ it with the exporter and interceptor configurations.
 For example, to allow trace exporting to Stackdriver and Zipkin:
 
 ```yaml
-stackdriver:
-  project: "your-project-id"
-  enable_traces: true
+exporters:
+    stackdriver:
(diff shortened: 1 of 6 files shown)


odeke-em:
Thank you @songy23!

PR #119, opened 2018-10-18
interceptor/opencensus: remove redundant OCInterceptor prefix

### LINKED ISSUE #88: Rename ocinterceptor.OCInterceptor to ocinterceptor.Interceptor
To minimize the stuttering of the names.

odeke-em:
Given that the package name is already ocinterceptor, using
OCInterceptor as a prefix for the type name is now unnecessary.
That naming was because the package was one of the first ones
to be added. Now that the packaging is mature this has come up
as a better suggestion to remove that unnecessary prefix.

Fixes #88


odeke-em:
Thank you @songy23!

PR #120, opened 2018-10-18
interceptor/opencensus: rename OCOption to Option

### LINKED ISSUE #87: Rename ocinterceptor.OCOption to ocinterceptor.Option
The option is an interceptor option, not an OpenCensus option.

odeke-em:
OCOption is redundant within the package ocinterceptor,
hence fix that with a rename to Option.

Fixes #87


odeke-em:
Merci @songy23!

PR #122, opened 2018-10-19
translator/trace: add an Annotations translator

odeke-em:
From examining and comparing the Agent proto
vs the OpenCensus Go SpanData, discovered the
need to translate instead of only expecting
time events. Critical while examining differences
in the Zipkin interceptor.


odeke-em:
Thank you @helly25 for the review and approval!

PR #124, opened 2018-10-21
Add Jaeger Exporter to Agent

pjanotti:
Add Jaeger exporter to ocagent following the same code pattern already
in use for the other exporters.

### DIFF
--- a/cmd/ocagent/config.go
+++ b/cmd/ocagent/config.go
@@ -17,10 +17,9 @@ package main
 import (
 	"log"
 
-	yaml "gopkg.in/yaml.v2"
-
 	"github.com/census-instrumentation/opencensus-service/exporter"
 	"github.com/census-instrumentation/opencensus-service/exporter/exporterparser"
+	yaml "gopkg.in/yaml.v2"
 )
 
 // We expect the configuration.yaml file to look like this:
@@ -113,6 +112,7 @@ func exportersFromYAMLConfig(config []byte) (traceExporters []exporter.TraceExpo
 		{name: "datadog", fn: exporterparser.DatadogTraceExportersFromYAML},
 		{name: "stackdriver", fn: exporterparser.StackdriverTraceExportersFromYAML},
 		{name: "zipkin", fn: exporterparser.ZipkinExportersFromYAML},
+		{name: "jaeger", fn: exporterparser.JaegerExportersFromYAML},
 	}
 
 	for _, cfg := range parseFns {
--- a/exporter/exporterparser/jaeger.go
+++ b/exporter/exporterparser/jaeger.go
@@ -1,0 +1,87 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


exporter/exporterparser/jaeger.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/census-instrumentation/opencensus-service/exporter"
+	"go.opencensus.io/exporter/jaeger"
+	"go.opencensus.io/trace"
+)
+
+// Slight modified version of go/src/go.opencensus.io/exporter/jaeger/jaeger.go
+type jaegerConfig struct {
+	Endpoint          string `yaml:"endpoint,omitempty"`
+	CollectorEndpoint string `yaml:"collector_endpoint,omitempty"`
+	AgentEndpoint     string `yaml:"agent_endpoint,omitempty"`
```
yurishkuro: Why 3 endpoints? Wouldn't it just talk to Jaeger collector?

pjanotti:
Thanks for looking @yurishkuro. Good question regarding testing:

The exporter itself has some tests at https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/jaeger/jaeger_test.go

The exporters code in ocagent itself doesn't have any tests at this moment. These tests should be mainly focusing on dealing with config and handling multiple exporters. I can take a look at adding some testing in this area too.

For this PR I ran a manual single box tests, but, I want to run some load/perf tests on ocagent itself specifically using the exporter but I didn't get to run those yet.

exporter/exporterparser/jaeger.go
pjanotti: I wasn't sure about those but I kept them since they are available on the exporter configuration itself. The exporter code just that handles those are below, with agreement of the parts I certainly can drive the proper changes to any related repo.

https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/jaeger/jaeger.go#L84-L101

cmd/ocagent/config.go
```diff
@@ -17,10 +17,9 @@ package main
 import (
 	"log"
 
-	yaml "gopkg.in/yaml.v2"
-
 	"github.com/census-instrumentation/opencensus-service/exporter"
 	"github.com/census-instrumentation/opencensus-service/exporter/exporterparser"
+	yaml "gopkg.in/yaml.v2"
```
odeke-em: Please revert this line, keep this import separated from the "github.com/census-instrument*" imports.

exporter/exporterparser/jaeger.go
```diff
@@ -0,0 +1,87 @@
...
+		CollectorEndpoint: jc.CollectorEndpoint,
+		AgentEndpoint:     jc.AgentEndpoint,
+		Username:          jc.Username,
+		Password:          jc.Password,
+		Process: jaeger.Process{
+			ServiceName: jc.ServiceName,
+		},
+	})
+	if err != nil {
+		log.Fatalf("Cannot configure Jaeger exporter: %v", err)
```
odeke-em: Please return this error instead of `log.Fatalf` so `return nil, nil, err`

exporter/exporterparser/jaeger.go
odeke-em: @pjanotti as @yurishkuro suggested, please just use the `CollectorEndpoint` only since we'll be sending traces to it, those other endpoints are part of the options but unnecessary for this use case.

pjanotti:
Thanks for the welcome and review @odeke-em!

I'm looking on my side about the CLA issue.

pjanotti:
I signed it.

odeke-em:
Github's availability and consistency problems for the past 24 hours caused that onslaught of comments which weren't posted here but were sent via emails.

Issue #20, opened 2018-09-13
metrics/interceptor: define how the agent will be able to pull metrics from tertiary sources


odeke-em:
Following suit with https://github.com/census-instrumentation/opencensus-service/issues/19, we need to define how the agent will be able to pull metrics from sources like:
* Prometheus
and transform them into OpenCensus Metrics which can then be consumed/exported to the various backends. 

Basically this issue is to define the ability for the agent to be pointed to users that have all their metrics going to "System A" but are slowly doing a migration to "System B" or OpenCensus and should be able to have this done transparently without having to invasively instrument their code or take down their services. Some services like Prometheus will require us to pull from client applications while others like Stackdriver support pushing to an end point so can be intercepted.

Anyways this issue to have that discussion and set priorities and reminders for how we shall proceed.

songy23:
Marked as P2 since we need to finalize the OpenMetrics data models first.

odeke-em:
Alright, the metrics data model was finished as per https://github.com/census-instrumentation/opencensus-proto/pull/121 so I think this issue can now be tackled

Issue #99, opened 2018-10-16
metrics: define interceptor


odeke-em:
Discuss and propose the interceptor interface for metrics.

songy23:
I guess we can call this done for now. https://github.com/census-instrumentation/opencensus-service/pull/125 added the generic metric interceptor, I'll file a new issue to add the OC implementation of metric interceptor.

PR #125, opened 2018-10-22
interceptor: start adding metric receiver and interceptor.

songy23:
Updates https://github.com/census-instrumentation/opencensus-service/issues/99.
Fixes https://github.com/census-instrumentation/opencensus-service/issues/20.

From https://github.com/census-instrumentation/opencensus-go/issues/940#issuecomment-431929219, OC-Go is going to use the proto gen-go files for `Metrics` implementation, so I think we can start adding the metric receiver and interceptor interfaces now. Note that although the `Metrics` proto hasn't been finalized or released, any future changes will be backwards-compatible, so we should be able to use the `metricpb.Metric` for now.

Next step: add OpenCensus `metric_interceptor`, which depends on [metrics_service.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/agent/metrics/v1/metrics_service.proto).

### DIFF
--- a/interceptor/interceptor.go
+++ b/interceptor/interceptor.go
@@ -17,6 +17,7 @@ package interceptor
 import (
 	"context"
 
+	"github.com/census-instrumentation/opencensus-service/metricreceiver"
 	"github.com/census-instrumentation/opencensus-service/spanreceiver"
 )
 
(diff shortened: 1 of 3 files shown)


interceptor/interceptor.go
```diff
@@ -31,3 +32,15 @@ type TraceInterceptor interface {
...
+
+// A MetricsInterceptor is an "arbitrary data"-to-"metric proto" converter.
+// Its purpose is to translate data from the wild into metric proto accompanied
+// by a *commonpb.Node to uniquely identify where that data comes from.
+// MetricsInterceptor feeds a metricsreceiver.MetricsReceiver with data.
+//
+// For example it could be Prometheus data source which translates
+// Prometheus metrics into *metricpb.Metric-s.
+type MetricsInterceptor interface {
+	StartMetricInterception(ctx context.Context, destination metricsreceiver.MetricsReceiver) error
```
odeke-em: ditto here about Metrics and below too :)

songy23:
>  could you please make MetricReceiver --> MetricsReceiver. That is because Metric on its own will clash with the name Metric, the system of measurement but also we receive metrics in batches and so shall we do for Stats hence we'll also use StatsReceiver instead of StatReceiver. TraceReceiver works because a trace consists of multiple spans already

Make sense to me. I updated both receiver and interceptor to use plural (`Metrics`) instead of `Metric`. PTAL.

interceptor/interceptor.go
songy23: Overlooked this one - fixed.

PR #139, opened 2018-10-26
Add a minimal Makefile and set Travis to use it

pjanotti:
This is very minimalistic: it adds a Makefile and switches Travis
to use it. The Makefile tries to do the same things as the previous
script (golint is not enabled yet since it is breaking the build).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: go
 
 go:
   - 1.10.x
+  - 1.11.x
 
 install:
   - go get -t -v ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,5 @@ language: go
 go:
   - 1.10.x
 
-before_script:
-  - GO_FILES=$(find . -iname '*.go' | grep -v /vendor/)  # All the .go files, excluding vendor/ if any
-
-script:
-  - if [ -n "$(gofmt -s -l $GO_FILES)" ]; then echo "gofmt the following files:"; gofmt -s -l $GO_FILES; exit 1; fi
-  - go test -v -race ./...            # Run all the tests with the race detector enabled
-  - 'if [[ $TRAVIS_GO_VERSION = 1.8* ]]; then ! golint ./... | grep -vE "(_mock|_string|\.pb)\.go:"; fi'
+install:
+  - go get -t -v ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: go
 
 go:
   - 1.10.x
-  - 1.11.x
-
+ 
 install:
   - go get -t -v ./...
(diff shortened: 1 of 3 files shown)


pjanotti:
Marking as WIP so it can be validated and perhaps updated before people starting to review.

.travis.yml
```diff
@@ -1,12 +1,10 @@
 language: go
 
+go_import_path: github.com/census-instrumentation/opencensus-service
+
 go:
   - 1.10.x
-
-before_script:
```
odeke-em: I believe we want the Travis configuration to always run these scripts. Currently I don't see how they are run by Travis in combination with the Makefile

.travis.yml
```diff
@@ -1,12 +1,10 @@
...
-  - GO_FILES=$(find . -iname '*.go' | grep -v /vendor/)  # All the .go files, excluding vendor/ if any
-
-script:
-  - if [ -n "$(gofmt -s -l $GO_FILES)" ]; then echo "gofmt the following files:"; gofmt -s -l $GO_FILES; exit 1; fi
-  - go test -v -race ./...            # Run all the tests with the race detector enabled
-  - 'if [[ $TRAVIS_GO_VERSION = 1.8* ]]; then ! golint ./... | grep -vE "(_mock|_string|\.pb)\.go:"; fi'
+ 
+install:
+  - make install-tools
```
odeke-em: Could we explicitly trigger the test and fmt directives of the Makefile? `install-tools` only seems tailored to download `golint` which am not yet sold on disabling the build.

Makefile
```diff
@@ -0,0 +1,56 @@
...
+		echo "$(GOFMT) FAILED => gofmt the following files:\n"; \
+		echo "$$FMTOUT\n"; \
+		exit 1; \
+	fi
+
+.PHONY: lint
+lint:
+	@LINTOUT=`$(GOLINT) ./... 2>&1`; \
+	if [ "$$LINTOUT" ]; then \
+		echo "$(GOLINT) FAILED => clean the following lint errors:\n"; \
```
odeke-em: Am not sure that we want golint suggestions to fail the build as they are suggestions https://github.com/golang/lint#purpose I personally wouldn't want `mp[key]++` instead of `mp[key] += 1` which is more legible and easier to spot just because of a sweeping `+= 1` should ALWAYS be `++`

.travis.yml
pjanotti: In the presence of a Makefile Travis runs `make` so it goes to its default target. Right now it is "fmt test" so it runs gofmt and runs the tests. This step here was part of the "setup" and just needed if we wanted to run golint on Travis.

Makefile
pjanotti: I will remove golint related stuff from this PR. Let's resolve golint at #141

PR #144, opened 2018-10-29
README: slightly reword indefinite retries.

songy23:


README.md
```diff
@@ -124,13 +124,14 @@ Agent should drop the whole message and return an error to the client. In additi
...
 Identifier is no longer needed once the streams are established.
-3. On Library side, if connection to Agent failed, Library should retry indefintely. (Reason:
-consider environments where the running applications are already instrumented with OpenCensus
-Library but Agent is not deployed yet. Sometime in the future, we can simply roll out the Agent
-to those environments and Library would automatically connect to Agent with indefinite retries.
-Zero changes are required to the applications.) Depending on the language and implementation,
-retry can be done in either background or a daemon thread. Retry should be performed at a fixed
-frequency (rather than exponential backoff) to have a deterministic expected connect time.
+3. On Library side, if connection to Agent failed, Library should retry indefintely if possible,
+subject to available/configured memory buffer size. (Reason: consider environments where the
```
pjanotti: Sorry, I'm arriving late at this... it will be good to make explicit that what is dropped in the case the buffer is full: typically for monitoring data the older data is dropped, but this makes the implementation slightly harder and IIRC that is not what the go exporter does... perhaps document as it is and let's discuss if we should consider a change.

songy23: Good point, I created an issue https://github.com/census-instrumentation/opencensus-service/issues/146 for follow-up. In Agent we have (observability)[https://github.com/census-instrumentation/opencensus-service/blob/master/internal/observability.go] and (Acknowledgement)[https://github.com/census-instrumentation/opencensus-service/blob/master/spanreceiver/spanreceiver.go#L29] for this purpose. Maybe we should do the same for exporters.

PR #85, opened 2018-10-15
Minor readability improvements

rakyll:
- Shorter names for variables and types.
- Print flag values instead of pointer values.
- Use -config instead of -exporters-yaml because config contains more than exporters.

PR #145, opened 2018-10-29
interceptor, exporter/zipkin: own Trace exporter with per span attributes

### LINKED ISSUE #132: trace/interceptors: Zipkin loses remote endpoint
As part of the translation from Zipkin interceptor to OC proto to Zipkin exporter, the remote endpoint is lost. This is to track the issue.

odeke-em:
This change introduces our own copy of a modified Zipkin Trace
exporter from OpenCensus-Go. This modification takes into
account the fact that our Zipkin interceptor "intercepts"/proxies
spans while the OpenCensus-Go Zipkin exporter assumes a static
world in which there is only one localEndpoint and no remoteEndpoint.
However, localEndpoint and remoteEndpoint are per-span fields
hence the necessity for us to ensure that users of this project
can have an almost transparent proxying and processing of Zipkin spans.

This change also features an end-to-end test that takes intercepted
Zipkin spans e.g. from a Zipkin Java exporting application, intercepts
them to the AgentCore and then on export we ensure that our output
has all the endpoints "localEndpoint", "remoteEndpoint" as well
as an equivalence of all the other fields.

Fixes #132
Supersedes #959
Supersedes #960
Files #85


exporter/exporterparser/zipkin.go
```diff
@@ -106,89 +110,250 @@ func ZipkinExportersFromYAML(config []byte) (tes []exporter.TraceExporter, doneF
...
 		defaultLocalEndpointURI: defaultLocalEndpointURI,
-		serviceNameToExporter:   make(map[string]*zipkin.Exporter),
+		reporter:                reporter,
 	}
+	return zle, nil
+}
 
-	// Ensure that we add the default reporter's Close functions
-	zle.shutdownFns = append(zle.shutdownFns, reporter.Close)
+func zipkinEndpointFromNode(node *commonpb.Node, serviceName string, prefixFunc func(string) string) (*zipkinmodel.Endpoint, error) {
```
pjanotti: nit: seeing how the prefixFunc is used perhaps it will be better to use an array of literal strings to avoid allocations. 

Personally I would have 3 functions: one with a bool addRemotePrefix and two others like zipkinLocalEndpointFromNode and zipkinRemoteEndpointFromNode that just call the first with the proper value for addRemotePrefix. This way the allocations can be avoided and the prefix function is not needed.

exporter/exporterparser/zipkin.go
```diff
@@ -106,89 +110,250 @@ func ZipkinExportersFromYAML(config []byte) (tes []exporter.TraceExporter, doneF
...
+func remoteZipkinServiceNamePrefix(fieldName string) string {
+	return "zipkin.remoteEndpoint." + fieldName
 }
 
-func (ze *zipkinExporter) stop() error {
-	ze.mu.Lock()
-	defer ze.mu.Unlock()
+func (ze *zipkinExporter) serviceNameOrDefault(node *commonpb.Node) string {
+	ze.mu.RLock()
+	defaultServiceName := ze.defaultServiceName
```
pjanotti: Per conversation on previous PR the mutex is used here to protect the method during a call to stop(), however, the code as it is will give the impression that `ze.defaultServiceName` can actually change which is not the case (did I miss it somewhere?). If this is to prevent execution on reporter on the case of a call to stop() then using a RW lock with a int flag to signal when it is being closed (or already closed) seems reasonable.

exporter/exporterparser/zipkin.go
```diff
@@ -106,89 +110,250 @@ func ZipkinExportersFromYAML(config []byte) (tes []exporter.TraceExporter, doneF
...
+		// we need them enclosed in square brackets as per
+		//    https://golang.org/pkg/net/#SplitHostPort
+		endpointURI = fmt.Sprintf("[%s]", ipv6)
+		ipv6Selected = true
+	}
+
+	port := attributes[prefixFunc("port")]
+	if port != "" {
+		endpointURI += ":" + port
+	} else if ipv6Selected { // net.SplitHostPort requires a port of ipv6 values
```
basvanbeek: I would change this to:
```go
if port := attributes[prefixFunc("port")]; port == "" {
  endpointURI += ":0"
} else {
  endpointURI += ":"+port
}
return openzipkin.NewEndpoint(serviceName, endpointURI)
```
then you can also drop the ipv6Selected variable above as it is not needed anymore

exporter/exporterparser/zipkin_test.go
```diff
@@ -0,0 +1,228 @@
...
+)
+
+// This function tests that Zipkin spans that are intercepted then processed roundtrip
+// back to almost the same JSON with differences:
+// a) Go's net.IP.String intentional shortens 0s with "::" but also converts to hex values
+//    so
+//          "7::0.128.128.127"
+//    becomes
+//          "7::80:807f"
+// b) OpenZikpin-Go unconditionally creates a non-blank IPv4 if only IPv6 addresses were resolved
```
basvanbeek: this should be fixed per: https://github.com/openzipkin/zipkin-go/pull/86

odeke-em: Perfect, thank you @basvanbeek for the quick response and fix!

exporter/exporterparser/zipkin.go
odeke-em: Not really, even for ipv4 we need that port.

exporter/exporterparser/zipkin.go
```diff
@@ -106,89 +110,250 @@ func ZipkinExportersFromYAML(config []byte) (tes []exporter.TraceExporter, doneF
...
-	exporter := ze.exporterForNode(node)
-	return exportSpans(ctx, node, "zipkin", exporter, spandata)
+func (ze *zipkinExporter) zipkinSpan(node *commonpb.Node, s *trace.SpanData) zipkinmodel.SpanModel {
+	localEndpointServiceName := ze.serviceNameOrDefault(node)
+	localEndpoint, err := zipkinEndpointFromNode(node, localEndpointServiceName, func(s string) string { return s })
+	if err != nil {
+		panic(err)
+	}
+
+	remoteServiceName := node.Attributes[remoteZipkinServiceNamePrefix("serviceName")]
```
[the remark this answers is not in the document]
odeke-em: Actually I'll retract my acceptance, a literal will work just for remoteServiceName but then not for the other fields and enumerating all the other combinations just makes for lots of code repetition. Perhaps we can send a follow-up in another PR?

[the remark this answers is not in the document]
odeke-em: @pjanotti I've addressed the feedback to use literals in the commit https://github.com/census-instrumentation/opencensus-service/pull/145/commits/b95b453cf327c1a72cd9c6dd369cbd98e79b5944 but at the expense of a little more code. Thank you though for the suggestions!

exporter/exporterparser/zipkin.go
odeke-em: @pjanotti thanks for raising that, it was more of a defensibility for safety which as I mentioned let's try to ensure that for accesses we first protect them then perhaps later on remove locks. I've removed it and updated the code, PTAL and thanks again.

exporter/exporterparser/zipkin.go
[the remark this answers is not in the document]
odeke-em: Oh interesting, I missed that, elegant solution @pjanotti, please feel free to send it in the next cycle :)

internal/test_utils.go
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import "encoding/json"
+
+// AnagramicSignature is useful for comparing inputs such as roundtrip serialized
+// JSON or encodings whose iteration orders change as object are serialized and deserialized
```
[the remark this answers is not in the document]
odeke-em: Roger that, thanks.

odeke-em:
Thank you for the review @pjanotti! We can send smaller follow-up PRs.

PR #148, opened 2018-10-31
Remove extra functions and improve JSON check in tests

pjanotti:
These changes are a follow up to comments in a previous PR. It changes
the normalization process to compare JSON strings and remove some
extra functions by replacing them with an if/else statement and local
variables.

Original PR that triggered these changes is #145


exporter/exporterparser/zipkin.go
```diff
@@ -138,20 +138,27 @@ func zipkinEndpointFromNode(node *commonpb.Node, serviceName string, endpointTyp
 	// }
 	attributes := node.Attributes
 
+	var ipv4Key, ipv6Key, portKey string
+	if endpointType == isLocalEndpoint {
```
odeke-em: Elegant!

internal/testutils/test_utils.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// GenerateNormalizedJSON generates a normalized JSON from the string
+// given to the function. Useful to compare JSON contents that
+// may have differences due to formatting. It returns nil in case of
+// invalid JSON.
+func GenerateNormalizedJSON(j string) string {
+	var i interface{}
+	json.Unmarshal([]byte(j), &i)
+	n, _ := json.Marshal(i)
```
odeke-em: In the future perhaps use `blob` or something as `n` is for integers.

pjanotti:
Thanks for the review @odeke-em!

PR #155, opened 2018-11-01
interceptor/zipkin: add logic to parse from Protobuf

### LINKED ISSUE #154: interceptor/zipkin: consider implementing proto interception
Suggested in a review by @basvanbeek in https://github.com/census-instrumentation/opencensus-service/pull/111#discussion_r228709755

For maximum compatibility with Zipkin services, perhaps allow intercepting proto spans.

The next question then though is:
a) What is the service being implemented, gRPC or just sending proto over HTTP or some other transport?
b) What port

We are also privileged to be able to directly ping @adriancole @basvanbeek for help with some questions.

odeke-em:
interceptor/zipkin: add logic to parse from Protobuf
  
Can now intercept Zipkin Protobuf spans that are
sent to a Zipkin server.

The ZipkinInterceptor multiplexes on the "Content-Type"
header, checking if the value set is "application/x-protobuf"
otherwise fallback to try parsing as JSON.

Fixes #154
Spawned https://github.com/openzipkin/zipkin-go/issues/87
Waiting on https://github.com/openzipkin/zipkin-go/pull/88

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
 	github.com/jmespath/go-jmespath v0.0.0-20180206201540-c2b33e8439af // indirect
 	github.com/kr/pretty v0.1.0 // indirect
-	github.com/openzipkin/zipkin-go v0.1.2
+	github.com/openzipkin/zipkin-go v0.1.3
 	github.com/philhofer/fwd v1.0.0 // indirect
 	github.com/pkg/errors v0.8.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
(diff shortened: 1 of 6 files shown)


odeke-em:
Alright https://github.com/openzipkin/zipkin-go/pull/88 has been merged so it is now open season to review this PR

odeke-em:
Roger that, I'll update it

odeke-em:
@songy23 thank you for the review and approval. This change is basically receiving proto3 data instead of JSON by HTTP upload which @adriancole co-signed as per https://github.com/census-instrumentation/opencensus-service/issues/154#issuecomment-434876715

Am going to merge this to avoid blocking the updates of this scrum but I am doing so with a high probability that this is the right approach, we can always make updates.

PR #156, opened 2018-11-01
Add docker scratch for agent and collector

pjanotti:
* Adds the scratch docker for agent and collector
* Opportunistically change some close functions to close the server 
instead of the listener that they depend on (a somewhat cleaner shutdown),
the servers close take care of closing the listeners

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,6 @@ GOTEST=go test
 GOFMT=gofmt
 GOOS=$(shell go env GOOS)
 
-OUT_DIR?=./bin/
-
 GIT_SHA=$(shell git rev-parse --short HEAD)
 BUILD_INFO_IMPORT_PATH=github.com/census-instrumentation/opencensus-service/internal/version
 BUILD_INFO=-ldflags "-X $(BUILD_INFO_IMPORT_PATH).GitHash=$(GIT_SHA)"
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,11 @@
 ALL_SRC := $(shell find . -type f -name '*.go' -not -path "./vendor/*")
 
-GOTEST_OPT=-v -race
+GOTEST_OPT?=-v -race
 GOTEST=go test
 GOFMT=gofmt
 GOOS=$(shell go env GOOS)
+
+OUT_DIR?=./bin/
 
 GIT_SHA=$(shell git rev-parse --short HEAD)
 BUILD_INFO_IMPORT_PATH=github.com/census-instrumentation/opencensus-service/internal/version
@@ -13,10 +15,6 @@ BUILD_INFO=-ldflags "-X $(BUILD_INFO_IMPORT_PATH).GitHash=$(GIT_SHA)"
 
 .PHONY: default_goal
 default_goal: fmt test
-
-.PHONY: clean
-clean:
-	rm -rf bin/ 
 
 .PHONY: test
 test:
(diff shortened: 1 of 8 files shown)


pjanotti:
I suspect that the test failure is related to the wait below having completed before the expected response was ready.

https://github.com/census-instrumentation/opencensus-service/blob/72aa6578c371fdb4c98f40748b47f3f7e2044340/exporter/exporterparser/zipkin_test.go#L78-L84

Makefile
```diff
@@ -39,6 +35,28 @@ agent:
 collector:
 	CGO_ENABLED=0 go build -o ./bin/occollector_$(GOOS) $(BUILD_INFO) ./cmd/occollector
 
+.PHONY: docker-component # Not intended to be used directly
+docker-component: check-component
+	GOOS=linux $(MAKE) $(COMPONENT)
+	cp ./bin/oc$(COMPONENT)_linux ./cmd/oc$(COMPONENT)/
+	docker build -t oc$(COMPONENT) ./cmd/oc$(COMPONENT)/
+	rm ./cmd/oc$(COMPONENT)/oc$(COMPONENT)_linux
```
maynk: I would suggest to use -linux instead of _linux

cmd/ocagent/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# Use a helper to create an emtpy configuration since the agent requires such file
+FROM alpine:3.7 as helper
```
maynk: this is a bit strange
maybe we can get config from environment vars instead of mandating config file for startup (and fall back to defaults).

cmd/ocagent/main.go
```diff
@@ -121,14 +121,15 @@ func runZPages(port int) func() error {
 		log.Fatalf("Failed to bind to run zPages on %q: %v", addr, err)
 	}
 
+	srv := http.Server{Handler: zPagesMux}
 	go func() {
 		log.Printf("Running zPages at %q", addr)
-		if err := http.Serve(ln, zPagesMux); err != nil {
+		if err := srv.Serve(ln); err != nil && err.Error() != "http: Server closed" {
```
maynk: nit: is this really the best practice? isn't there a const defined somewhere?

cmd/ocagent/main.go
```diff
@@ -152,13 +153,14 @@ func runOCInterceptor(addr string, sr spanreceiver.SpanReceiver) (doneFn func()
 	agenttracepb.RegisterTraceServiceServer(srv, oci)
 	go func() {
 		log.Printf("Running OpenCensus interceptor as a gRPC service at %q", addr)
-
-		// Not using log.Fatalf(srv.Serve(ln)) because CTRL-C will close the listener as the user requested
-		// and that log.Fatalf will produce an unrelated but misleading error:
-		//     "Failed to run OpenCensus interceptor: accept tcp 127.0.0.1:55678: use of closed network connection"
-		_ = srv.Serve(ln)
+		if err := srv.Serve(ln); err != nil {
```
odeke-em: @pjanotti please take a look at that comment that you've deleted on why it was `_ = srv.Serve(ln)`

cmd/ocagent/Dockerfile
pjanotti: Yeah, right now the agent requires the file to be present (even an empty one). We should use something like Viper combined with Cobra to allow flexibility in this regard: no file is required by default but settings can be specified in a file or cli as convenient. Tracking #158

cmd/ocagent/main.go
pjanotti: Ah, thanks, that was what I found at first for golang in general but the package itself defines the proper constant. I will update.

cmd/ocagent/main.go
pjanotti: The thing was that it was closing the listener from which the server depended upon, the preferred way is to let the server itself close the listener. This way there is no error message from the gRPC server.

PR #161, opened 2018-11-03
go.mod: update all dependencies

odeke-em:
Updated all dependencies by using
```shell
GO111MODULE=on go get -u ./...
```
using Go1.11+

In the future, one can update dependencies by
```shell
GO111MODULE=on go get -u <package_path>
```
where `GO111MODULE=on` is only required if the
repository is checked out in your GOPATH


odeke-em:
Actually never mind me, am just peddling fake news in regards to
>We've always been able to test with Go1.11 before.

My apologies @pjanotti!

odeke-em:
For now, let's merge this PR in and when we need Go1.11 we can perform the procedure that you recommended.

Issue #136, opened 2018-10-25
interceptor/metrics: add OpenCensus metrics interceptor.


songy23:
Similar to OpenCensus [Trace interceptor](https://github.com/census-instrumentation/opencensus-service/blob/master/interceptor/opencensus/trace_interceptor.go).

This is currently blocked by:
1. https://github.com/census-instrumentation/opencensus-proto/pull/137
2. New release of opencensus-proto, which includes [metrics_service.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/agent/metrics/v1/metrics_service.proto).

songy23:
Going to mark this issue as done and create a new one for adding ocmetrics to commands.

PR #170, opened 2018-11-07
interceptor: rename ocinterceptor to octrace.

songy23:
Updates https://github.com/census-instrumentation/opencensus-service/issues/136.

Rename trace interceptor to differentiate from metrics interceptor.


interceptor/opencensustrace/doc.go
```diff
@@ -12,6 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package ocinterceptor is the logic for receiving OpenCensus trace protobuf defined spans from
+// Package octraceinterceptor is the logic for receiving OpenCensus trace protobuf defined spans from
 // already instrumented applications and then passing them onto a spanreceiver instance.
-package ocinterceptor
+package octraceinterceptor
```
odeke-em: Why don't we just name it `octrace`?

odeke-em:
Thank you @songy23 for working on this! I just have one request, could we instead just name it `octrace`  or `opencensustrace` since it is already a trace interceptor and also preserve the picked package name?

interceptor/opencensustrace/doc.go
songy23: `octrace` sounds good, updated.

songy23:
Thanks!

PR #177, opened 2018-11-07
Small README updates

pjanotti:
Just to make it match the current status of the repo. Some stuff there
is the direction that we want to go but it is not ready yet.


README.md
```diff
@@ -231,6 +232,8 @@ and then the Docker image `v1.0.0` of the agent can be started  by
 docker run -v $(pwd)/config.yaml:/config.yaml  -p 55678:55678  ocagent:v1.0.0
 ```
 
+A Docker scratch image can be built with make by targetting `docker-agent`.
```
odeke-em: s/targetting/targeting/g

README.md
```diff
@@ -260,55 +263,23 @@ agent/client health information/inventory metadata to downstream exporters.
...
-config.yaml:
-
+$ make docker-collector
+$ docker run --rm -it -p 55678:55678 occollector
 ```
-collector:
-  endpoint: "https://collector.local"
-```
-
-Run the example application that collects traces and exports
```
odeke-em: Why did you remove the stuff from here down below? They are necessary for folks to get a feel for what the agent does, they are unrelated to the collector.

pjanotti: Ops cut too much

pjanotti: Probably need to be moved to a section closer to the agent...

README.md
pjanotti: thanks

README.md
pjanotti: Actually, same it is already located on top I will do a small update to it.

odeke-em:
@flands I don't think this PR was complete, @pjanotti did you see my comments about reverting the instructions on how to run the agent that weren't part of the occollector?

README.md
odeke-em: Hey @pjanotti I didn't see an update to this content that was removed but it was marked resolved.

pjanotti:
@odeke-em yes, I saw your comment. The same text already existed on the top so I just updated it too match the one below.

pjanotti:
Here https://github.com/census-instrumentation/opencensus-service#running-an-end-to-end-exampledemo

README.md
pjanotti: The content at the bottom was removed since it was already present before it.

https://github.com/census-instrumentation/opencensus-service#running-an-end-to-end-exampledemo

odeke-em:
It doesn't show them the output from `ocagent` which is important for users to easily follow along as they build. It's alright for now, but in the future it'd be nice to keep or at least migrate such previous printouts.

Issue #151, opened 2018-10-31
discussion: rename interceptors to importers?


rakyll:
It seems like importers could have been named as "importers". I had to look for documentation to understand what an interceptor is, although "importer" would be more self descriptive. Importer is also symmetric to an exporter.

What do you think?

songy23:
I guess we copied the name "interceptor" from gRPC (e.g [intercetpor.go](https://github.com/grpc/grpc-go/blob/4f70de23e6d83e800fe1b15bdd6e3e11b9a1f27a/interceptor.go)). I proposed "adapter" initially though I don't have strong opinions towards this. "importer" sounds good to me too.

/cc @bogdandrutu @odeke-em

odeke-em:
@rakyll thank you for the request for clarification and the suggestion!

IMHO interceptor is used because it implies a sort of box-in-the-middle to capture then reroute input which is the intention here -- both passive and active capturing for example the Zipkin "interceptor" is just an HTTP server that receives spans sent to it.
Importer suggests an active fetch model where we seek out configurations/sources and pull from them which misses the part about the component passively waiting/being a sink. My opinion is to keep the package name as is.

However, I'll defer to others for the final decision on renaming.

/cc also @pjanotti @flands @maynk

flands:
Another vote for importer here.

pjanotti:
"Receiver" or "Listener" fit the actual role pretty well, and as mentioned this fits with the "passive" way that they act, instead of an active role implied by "importer" or "interceptor".

yurishkuro:
+1 for receiver.

flands:
Looks like "receiver" wins. Will open an issue.

flands:
Tracked via #166

bogdandrutu:
+1 for "receiver".

PR #178, opened 2018-11-07
all: rename interceptor to receiver

### LINKED ISSUE #166: Rename interceptors to receivers
Per #151 we agreed to rename interceptors to receivers. This is the issue to track the work item.

odeke-em:
As per the consensus from #151, rename interceptor to receiver
but also the related interfaces too and their methods.

Fixes #166


odeke-em:

PR #180, opened 2018-11-07
Output of ocagent

pjanotti:
Adding output of the agent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ and start it:
 
 ```shell
 $ ./bin/ocagent_$(go env GOOS)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -168,6 +168,7 @@ and start it:
 
 ```shell
 $ ./bin/ocagent_$(go env GOOS)
+$ 2018/10/08 21:38:00 Running OpenCensus interceptor as a gRPC service at "127.0.0.1:55678"
```
odeke-em: s/interceptor/receiver/g as it has just changed 2 PRs ago

pjanotti: good catch

PR #183, opened 2018-11-08
receiver/ocmetrics: start adding OCMetricsReceiver.

songy23:
Updates https://github.com/census-instrumentation/opencensus-service/issues/136.

Next step: add observability and tests, then update cmd, config and README about ocmetrics configuration.

### DIFF
--- a/receiver/opencensus/ocmetrics/opencensus.go
+++ b/receiver/opencensus/ocmetrics/opencensus.go
@@ -38,7 +38,7 @@ type Receiver struct {
 	metricBufferCount  int
 }
 
-// New creates a new opencensus.Receiver reference.
+// New creates a new ocmetrics.Receiver reference.
 func New(sr metricsink.Sink, opts ...Option) (*Receiver, error) {
 	if sr == nil {
 		return nil, errors.New("needs a non-nil metricsink.Sink")
--- a/receiver/opencensus/ocmetrics/opencensus.go
+++ b/receiver/opencensus/ocmetrics/opencensus.go
@@ -1,0 +1,157 @@
+// Copyright 2018, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ocmetrics
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"google.golang.org/api/support/bundler"
+
+	"go.opencensus.io/trace"
+
+	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
+	agentmetricspb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/metrics/v1"
+	metricspb "github.com/census-instrumentation/opencensus-proto/gen-go/metrics/v1"
+	resourcepb "github.com/census-instrumentation/opencensus-proto/gen-go/resource/v1"
+	"github.com/census-instrumentation/opencensus-service/internal"
+	"github.com/census-instrumentation/opencensus-service/metricsink"
+)
+
+// Receiver is the type used to handle metrics from OpenCensus exporters.
+type Receiver struct {
+	metricSink         metricsink.Sink
+	metricBufferPeriod time.Duration
+	metricBufferCount  int
+}
+
+// New creates a new opencensus.Receiver reference.
+func New(sr metricsink.Sink, opts ...Option) (*Receiver, error) {
+	if sr == nil {
+		return nil, errors.New("needs a non-nil metricsink.Sink")
+	}
+	ocr := &Receiver{metricSink: sr}
+	for _, opt := range opts {
+		opt.WithReceiver(ocr)
+	}
+	return ocr, nil
+}
+
+var _ agentmetricspb.MetricsServiceServer = (*Receiver)(nil)
+
+type bundledMetrics struct {
+	metrics  []*metricspb.Metric
+	node     *commonpb.Node
+	resource *resourcepb.Resource
+}
+
+var errMetricsExportProtocolViolation = errors.New("protocol violation: Export's first message must have a Node")
+
+const receiverName = "opencensus_metrics"
+
+// Export is the gRPC method that receives streamed metrics from
+// OpenCensus-metricproto compatible libraries/applications.
+func (ocr *Receiver) Export(mes agentmetricspb.MetricsService_ExportServer) error {
+	// The bundler will receive batches of metrics i.e. []*metricspb.Metric
+	// We need to ensure that it propagates the receiver name as a tag
+	ctxWithReceiverName := internal.ContextWithReceiverName(mes.Context(), receiverName)
+	metricsBundler := bundler.NewBundler((*bundledMetrics)(nil), func(payload interface{}) {
+		ocr.batchMetricExporting(ctxWithReceiverName, payload)
+	})
+
+	metricBufferPeriod := ocr.metricBufferPeriod
+	if metricBufferPeriod <= 0 {
+		metricBufferPeriod = 2 * time.Second // Arbitrary value
+	}
+	metricBufferCount := ocr.metricBufferCount
+	if metricBufferCount <= 0 {
+		// TODO: (@odeke-em) provide an option to disable any buffering
+		metricBufferCount = 50 // Arbitrary value
+	}
+
+	metricsBundler.DelayThreshold = metricBufferPeriod
+	metricsBundler.BundleCountThreshold = metricBufferCount
+
+	// Retrieve the first message. It MUST have a non-nil Node.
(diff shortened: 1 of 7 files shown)


receiver/ocmetrics/metrics_receiver.go
```diff
@@ -0,0 +1,117 @@
...
+	return &ocReceiverHandler{srvPort: port, receiverOptions: opts}, nil
+}
+
+// NewMetricsReceiverOnDefaultPort will create a handle that runs an OCMetricsReceiver at the default port.
+func NewMetricsReceiverOnDefaultPort(opts ...Option) (receiver.MetricsReceiver, error) {
+	return &ocReceiverHandler{srvPort: defaultOCMetricsReceiverPort, receiverOptions: opts}, nil
+}
+
+type ocReceiverHandler struct {
+	mu      sync.RWMutex
```
pjanotti: nit: since only Lock() and Unlock() are used with mu use a `sync.Mutex` instead.

receiver/ocmetrics/opencensus.go
```diff
@@ -0,0 +1,162 @@
...
+	metricBufferCount := ocr.metricBufferCount
+	if metricBufferCount <= 0 {
+		// TODO: (@odeke-em) provide an option to disable any buffering
+		metricBufferCount = 50 // Arbitrary value
+	}
+
+	metricsBundler.DelayThreshold = metricBufferPeriod
+	metricsBundler.BundleCountThreshold = metricBufferCount
+
+	// The first message MUST have a non-nil Node.
```
pjanotti: nit: the comment doesn't match the check below.

receiver/ocmetrics/opencensus.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Check the condition that the first message has a non-nil Node.
+	if recv.Node == nil {
+		return errMetricsExportProtocolViolation
+	}
+
+	// metricsMetricsFn := internal.NewReceivedMetricsRecorderStreaming(mes.Context(), receiverName)
```
pjanotti: nit: leftover?

receiver/ocmetrics/opencensus.go
```diff
@@ -0,0 +1,162 @@
...
+
+	processReceivedMetrics := func(ni *commonpb.Node, resource *resourcepb.Resource, metrics []*metricspb.Metric) {
```
pjanotti: Q.: why define this as a lambda?

pjanotti: nit: note that defined here it breaks the reading flow separating `recv.Node` from `lastNonNilNode` and its usage in the loop below.

receiver/ocmetrics/metrics_receiver.go
songy23: Done.

receiver/ocmetrics/opencensus.go
songy23: Updated the comment a bit.

receiver/ocmetrics/opencensus.go
songy23: This will be included in the next PR (adding observability to metrics receiver). Removed for now.

receiver/ocmetrics/opencensus.go
songy23: I think it's because this func is only used in this scope. /cc @odeke-em 

Anyway I've moved the func down.

odeke-em:
Thanks @songy23 for working on this!

I took sometime to think about this and have some thoughts, below, please take a look:

So I have a reservation about creating an entirely new server just for metrics. We should have the agent/opencensus-receiver running on a single port and just optionally registering those services as a gRPC server would do, nor should we have to create a separate outer package under receiver. OpenCensus should be its single package just Zipkin, Jaeger are and then Stackdriver will be, etc.

### How we got here
We unfortunately performed a rename and move in https://github.com/census-instrumentation/opencensus-service/pull/170 but had forgotten about the goal of them just belonging to one package

So initially we were to have a directory structure:
```shell
receiver
   |- opencensus
       |- trace-receiver-code
       |- metrics-receiver-code
   |- zipkin
   |- jaeger
   |- istio
   |- ...
```
and then the OpenCensus receiver could just either run as a Trace receiver or Metrics receiver or both since it was just implementing both services
but changed that with https://github.com/census-instrumentation/opencensus-service/pull/170 to make `octrace` and `ocmetrics` but that perhaps a variant of it.

### Proposition
Perhaps let's move `octrace` and `ocmetrics` into `receiver/opencensus` (a modification of the original structure) so
```
receiver
   |-opencensus
      |- ocmetrics
         |- <ocmetrics logic and source code>
      |- octrace
         |- <octrace logic and source code>
      |- opencensus.go
```
where in `opencensus.go` we shall have a struct say `ocReceiver`
```go
type ocReceiver struct {
      grpcServer *grpc.Server
      traceReceiver *octrace.Receiver
      metricsReceiver *ocmetrics.Receiver
}
```

with these methods
```go
func (ocr *ocReceiver) StartMetricsReception(ctx context.Context, ms metricsink.Sink) error {
     // Use a sync.Once here to ensure that this is ever invoke only once
     msvc := ocmetrics.NewReceiver(ctx, ...params)
     grpc.RegisterMetricsServiceServer(ocr.grpcServer, msvc)
    ocr.metricsReceiver = msvc
}

func (ocrh *ocReceiverHandler) StopMetricsReception(ctx context.Context) error {
     // Handle stop logic here or perhaps noop
}

func (ocr *ocReceiver) StartTraceReception(ctx context.Context, ms metricsink.Sink) error {
     // Use a sync.Once here to ensure that this is ever invoke only once
     trsvc := octrace.NewReceiver(ctx, ...params)
     grpc.RegisterTraceServiceServer(ocr.grpcServer, trsvc)
    ocr.traceReceiver = trsvc
}

func (ocrh *ocReceiverHandler) StopTraceReception(ctx context.Context) error {
     // Handle stop logic here or perhaps noop
}

func (ocrh *ocReceiverHandler) Stop(ctx context.Context) error {
     // Use a sync.Once here to ensure that the server is only ever stopped once
     ocrh.grpcServer.Stop()
}

// Start invokes both StartTraceReception and StartMetricsReception.
// To separately start the trace or metrics receivers, invoke the various methods separately instead
func (ocrh *ocReceiverHandler) Start(ctx context.Context) error {
     ocrh.StartTraceReception(ctx)
     ocrh.StartMetricsReception(ctx)
}
```

and constructor which just takes a gRPC server and returns both Trace and Metrics receivers
```go
// TraceMetricsReceiver is an interface composed of both Trace.Receiver and Metrics.Receiver
type TraceMetricsReceiver interface {
     Trace.Receiver
      Metrics.Receiver
}

func New(srv *grpc.Server) (TraceMetricsReceiver, error) {
     return &ocReceiver{grpcServer: srv}
}
```

The above follows the previous structure but also shields us from mixing octrace and ocmetrics which was the intention of the renames in #170 but also allows us to reuse the same server/port while also shielding users from complexity i.e. they'll just import `github.com/census-instrumentation/opencensus-service/opencensus` and then in the `cmd/ocagent/*` code we can just invoke `StartTraceReception` and `StartMetricsReception`. It'll also require removing the logic that starts gRPC servers for the various receivers, since the original goal was for them to exist in the same package and reuse that server.

Hope this helps and I believe it will alleviate any concerns

songy23:
The directory layout makes sense to me. I'd like to make a separate PR to refactor the current trace receiver first, then I'll rebase and update this PR.

odeke-em:
@songy23 I've mailed out https://github.com/census-instrumentation/opencensus-service/issues/185 and after it we can then trivially make the ocmetrics addition. Thanks!

songy23:
Rebased and updated, PATL @odeke-em.

Observability and tests will be in a following PR. Don't want to make this PR too large.

receiver/opencensus/ocmetrics/opencensus.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+// Receiver is the type used to handle metrics from OpenCensus exporters.
+type Receiver struct {
+	metricSink         metricsink.Sink
+	metricBufferPeriod time.Duration
+	metricBufferCount  int
+}
+
+// New creates a new opencensus.Receiver reference.
```
odeke-em: `// New creates a new opencensus.Receiver`
reference is not too correct :)

receiver/opencensus/opencensus.go
```diff
@@ -89,8 +93,14 @@ func (ocr *ocReceiver) StartTraceReception(ctx context.Context, ts spansink.Sink
 }
 
 func (ocr *ocReceiver) StartMetricsReception(ctx context.Context, ms metricsink.Sink) error {
-	// Currently a noop: TODO(@odeke-em, @songya): add the metrics receiver here
-	return nil
+	var err = errAlreadyStarted
+
+	ocr.startMetricsReceiverOnce.Do(func() {
+		ocr.metricsReceiver, err = ocmetrics.New(ms)
```
odeke-em: Please check if err == nil i.e.
```go
if err == nil {
   srv := ocr.grpcServer()
   agentmetricspb.RegisterMetricsServiceServer(srv, ocr.metricsReceiver)
}
```
and we can do the same for octrace

receiver/opencensus/ocmetrics/opencensus.go
songy23: Good catch, done.

receiver/opencensus/opencensus.go
songy23: SG, added.

songy23:
Thanks!

PR #184, opened 2018-11-09
Trace translator from Jaeger Thrift to OC-Proto

pjanotti:
Adds a trace translator from Jaeger Thrift to OC-proto. This should
be used by the agent and colllector receivers so putting the code
under translator.

/cc @odeke-em


translator/trace/jaegerthrift_to_protospan.go
```diff
@@ -0,0 +1,240 @@
...
+	return spanID
+}
+
+func jtagsToAttributes(tags []*jaeger.Tag) (tracepb.Span_SpanKind, *tracepb.Span_Attributes) {
+	if tags == nil {
+		return tracepb.Span_SPAN_KIND_UNSPECIFIED, nil
+	}
+
+	var sKind tracepb.Span_SpanKind
+	sAttribs := &tracepb.Span_Attributes{
```
odeke-em: Ditto, please make this `attrsMap` and at the end before return `&tracepb.Span_Attributes{AttributeMap: attrsMap}`

translator/trace/jaegerthrift_to_protospan.go
```diff
@@ -0,0 +1,240 @@
...
+
+		sAttribs.AttributeMap[tag.Key] = attrib
+	}
+
+	return sKind, sAttribs
+}
+
+// epochMicrosecondsAsTime converts microseconds since epoch to time.Time value.
+func epochMicrosecondsAsTime(ts uint64) time.Time {
+	seconds := ts / 1000000
```
odeke-em: Debugging and reading this is a lot easier if we use `1en` e.g. `seconds := us / 1e6`
so let's make the argument name `us`
```go
func epochMicrosecondsAsTime(ts uint64) time.Time {
   seconds := us / 1e6
   nanos := 1e3  * (us % 1e6)
   return
```
but also I believe the direct and cheaper solution is
```go
func epochMicrosecondsAsTime(us uint64) time.Time {
   return time.Unix(0, int64(us*1e3))
}
```
since there are 1e3 nanoseconds in a microsecond but please help me check my solution

translator/trace/jaegerthrift_to_protospan_test.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+func TestJaegerThriftBatchToOCProto(t *testing.T) {
+	const numOfFiles = 2
+	for i := 1; i <= 2; i++ {
+		thriftInFile := fmt.Sprintf("./testdata/thrift_batch_%02d.json", i)
+		jb := loadJaegerThriftBatchFromJSON(t, thriftInFile)
+		octrace, err := JaegerThriftBatchToOCProto(jb)
+		if err != nil {
+			t.Errorf("Failed to handled Jaeger Thrift Batch from %q. Error: %v", thriftInFile, err)
```
odeke-em: If this fails, below we'll have a nil dereference and a bunch of panics. Please add a continue after this error.

translator/trace/jaegerthrift_to_protospan_test.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		wantedSpanCount, receivedSpanCount := len(jb.Spans), len(octrace.Spans)
```
odeke-em: s/wantedSpanCount/wantSpanCount/g
s/receivedSpanCount/gotSpanCount/g

translator/trace/jaegerthrift_to_protospan_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func loadJaegerThriftBatchFromJSON(t *testing.T, file string) *jaeger.Batch {
+	r := &jaeger.Batch{}
+	loadFromJSON(t, file, &r)
+	return r
+}
+
+func loadFromJSON(t *testing.T, file string, obj interface{}) {
+	jsonReader := getFile(t, file)
```
odeke-em: This leaks file resources as `.Close` is never called

translator/trace/jaegerthrift_to_protospan_test.go
```diff
@@ -0,0 +1,82 @@
...
+	err := json.NewDecoder(jsonReader).Decode(obj)
+	if err != nil {
+		t.Errorf("Failed to decode json file %q into type %T", file, obj)
+	}
+}
+
+func getFile(t *testing.T, file string) *os.File {
```
odeke-em: I don't think we need this and as I mentioned above in `loadFromJSON` we are leaking file resources.

We can simply make it
```go
func loadFromJSON(t *testing.T, file string, obj interface{}) error {
   blob, err := ioutil.ReadFile(file)
   if err != nil {
      return err
   }
   return json.Unmarshal(blob, obj)
}
```

translator/trace/jaegerthrift_to_protospan.go
pjanotti: It didn't occur to me that it could take zero for the seconds, but makes sense that it does; nice.

translator/trace/jaegerthrift_to_protospan_test.go
pjanotti: Ops, my naive golang assumption that it stopped the test right there, I will have to sprinkle `continue` in some other ifs (no panic but no point in continuing the test in such cases).

translator/trace/jaegerthrift_to_protospan_test.go
pjanotti: Pft, I used ReadFile at the top but when here I just saw ReadAll ...

pjanotti:
Thanks for the review @odeke-em

translator/trace/jaegerthrift_to_protospan_test.go
```diff
@@ -0,0 +1,77 @@
...
+		gj, wj := testutils.GenerateNormalizedJSON(string(gb)), testutils.GenerateNormalizedJSON(string(wb))
+		if gj != wj {
+			t.Errorf("The roundtrip JSON doesn't match the JSON that we want\nGot:\n%s\nWant:\n%s", gj, wj)
+		}
+	}
+}
+
+func loadFromJSON(file string, obj interface{}) error {
+	blob, err := ioutil.ReadFile(file)
+	if err == nil {
```
odeke-em: Good handling but in the future, please do early returns as Go programmers prefer that :) i.e.
```go
blob, err := ioutil.ReadFile(file)
if err != nil {
   return err
}
return json.Unmarshal(blob, obj)
```

PR #186, opened 2018-11-10
receiver/opencensus: restructure to ensure Trace and Metrics reception

### LINKED ISSUE #185: receiver/opencensus: revert receiver/octrace move then renest to allow octrace and ocmetrics
Coming here from the suggestion that I made in https://github.com/census-instrumentation/opencensus-service/pull/183#issuecomment-437496371

Initially we had `receiver/opencensus` but mistakenly renamed that to `receiver/octrace` and then were going to add `receiver/ocmetrics`. However, the receiver opencensus should just have functionality for both interfaces
* receiver.TraceReceiver
* receiver.MetricsReceiver

and from the users' perspective i.e.(cmd/ocagent/main.go) just invoke the various trace receiver methods without having to separately create and invoke each receiver. I've proposed this structure
```shell
receiver
   |-opencensus
      |- ocmetrics
         |- <ocmetrics logic and source code>
      |- octrace
         |- <octrace logic and source code>
      |- opencensus.go
```

odeke-em:
The OpenCensus Agent should implement both Trace and Metrics reception
on the same server, but just as separate services. This change reverts
to the structure before PR #170 to have opencensus as the top level
receiver like we would do for Zipkin, Istio, Jaeger, Stackdriver etc.

With this structure we can now trivially separate octrace and ocmetrics
yet keep the API simple and can use the same gRPC server.

Fixes #185
Updates #183


odeke-em:
>Do we want to allow only starting either Trace or Metrics reception, but not both? IIUC with the current approach it's either both enabled or both disabled.

@songy23 that's not true, this PR does exactly as we normally would. It only starts either of those methods and not both. In the PR itself I only do TraceReception
screen shot 2018-11-11 at 3 47 53 pm

songy23:
> It only starts either of those methods and not both. In the PR itself I only do TraceReception

Make sense to me.

odeke-em:
Thanks for the review @songy23, after this PR you'll then be able to add ocmetrics.

PR #190, opened 2018-11-12
translator/trace/jaeger: conservative allocations for only existent fields

### LINKED ISSUE #187: translator/trace/jaeger: only allocate TraceID and SpanID only on nil
The code in
https://github.com/census-instrumentation/opencensus-service/blob/5d8c38a04e7d44a9810890c19380536d163c3e80/translator/trace/jaegerthrift_to_protospan.go#L171-L175

Always allocates an 8 byte slice. However, for every root span, this will mean 8 bytes wasted. Let's fix that.

### LINKED ISSUE #188: translator/trace/jaeger: retrieve status from "status.code", "status.message" and only allocate if set
The code in
https://github.com/census-instrumentation/opencensus-service/blob/5d8c38a04e7d44a9810890c19380536d163c3e80/translator/trace/jaegerthrift_to_protospan.go#L198-L203

only retrieves and converts the status code iff "http.status_code" is set. Some exporters such as OpenCensus-Go send the status code and message as "status.code" and "status.message" respectively. Let's also retrieve status from those tags.

Also let's only allocate a Status object iff either of statusMessage or statusCode are set.

### LINKED ISSUE #189: translator/trace/jaeger: avoid make(map, len(source)) lest we allocate unnecessarily
Currently while say retrieve attributes, we also use the map creation idiom with a hint `make(map[<T>]<V>, sizehint)` and then always assume that we'll create values i.e. a pointer.

https://github.com/census-instrumentation/opencensus-service/blob/5d8c38a04e7d44a9810890c19380536d163c3e80/translator/trace/jaegerthrift_to_protospan.go#L185
https://github.com/census-instrumentation/opencensus-service/blob/5d8c38a04e7d44a9810890c19380536d163c3e80/translator/trace/jaegerthrift_to_protospan.go#L236

and
https://github.com/census-instrumentation/opencensus-service/blob/5d8c38a04e7d44a9810890c19380536d163c3e80/translator/trace/jaegerthrift_to_protospan.go#L50
https://github.com/census-instrumentation/opencensus-service/blob/5d8c38a04e7d44a9810890c19380536d163c3e80/translator/trace/jaegerthrift_to_protospan.go#L79

However, there will be many spans without attributes and then we can no longer reliably detect if attributes were set or not. We should let the Go runtime handle map sizing but also only allocate iff attributes were actually set.

odeke-em:
translator/trace/jaeger: conservative allocations for only existent fields
  
* Only allocate TraceID and SpanID iff they are non-nil lest we'll
consume 8/16 bytes for every span which becomes quite expensive
* Also retrieve "status.code" and "status.message" as the Jaeger exporter
from OpenCensus sends those
* Avoid creating an attributes map if attributes don't exist and
let go of the `make(map[<K>]<V>, size_hint)` idiom as it prevents
us from being able to detect if values were added to the attributes map
* Added in-code conversion tests to ensure fidelity preservation
* Translate "message" in logs/tags to a description value

Noticed while rolling out the Jaeger receiver/interceptor.

Fixes #187
Fixes #188
Fixes #189
Fixes #190

### DIFF
--- a/internal/internal.go
+++ b/internal/internal.go
@@ -87,6 +87,9 @@ func guessUnixHomeDir() string {
 
 // TimeToTimestamp converts a time.Time to a timestamp.Timestamp pointer.
 func TimeToTimestamp(t time.Time) *timestamp.Timestamp {
+	if t.IsZero() {
+		return nil
+	}
 	nanoTime := t.UnixNano()
 	return &timestamp.Timestamp{
 		Seconds: nanoTime / 1e9,
--- a/translator/trace/jaegerthrift_to_protospan.go
+++ b/translator/trace/jaegerthrift_to_protospan.go
@@ -18,6 +18,7 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"fmt"
+	"reflect"
 	"strconv"
 	"time"
 
@@ -40,6 +41,10 @@ func JaegerThriftBatchToOCProto(jbatch *jaeger.Batch) (*agenttracepb.ExportTrace
 }
 
 func jProcessToOCProtoNode(p *jaeger.Process) *commonpb.Node {
+	if p == nil {
+		return nil
+	}
+
 	node := &commonpb.Node{
 		Identifier:  &commonpb.ProcessIdentifier{},
 		LibraryInfo: &commonpb.LibraryInfo{},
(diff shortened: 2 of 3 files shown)


odeke-em:

PR #194, opened 2018-11-13
receiver/octrace: make config a no-op instead of unimplemented.

songy23:
Reason: some Agent Exporters may expect both `export` and `config` services to be implemented, otherwise the connection will be dropped.

### DIFF
--- a/receiver/opencensus/octrace/opencensus.go
+++ b/receiver/opencensus/octrace/opencensus.go
@@ -51,12 +51,14 @@ func New(sr spansink.Sink, opts ...Option) (*Receiver, error) {
 
 var _ agenttracepb.TraceServiceServer = (*Receiver)(nil)
 
-var errUnimplemented = errors.New("unimplemented")
-
 // Config handles configuration messages.
 func (oci *Receiver) Config(tcs agenttracepb.TraceService_ConfigServer) error {
 	// TODO: Implement when we define the config receiver/sender.
-	return errUnimplemented
+	_, err := tcs.Recv()
+	if err != nil {
+		return err
+	}
+	return nil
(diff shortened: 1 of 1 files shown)


odeke-em:
Which exporters expect both to be implemented?

songy23:
Java. If `config` returns error the Exporter will drop the connection and retry.

odeke-em:
So that's an implementation with Java. Is there a way of making Java ignore these errors until we implement the Config service?

songy23:
> Is there a way of making Java ignore these errors until we implement the Config service?

That requires a lot of revision on the retry logic. What's the downside of having a no-op config service?

songy23:
> The downside is that clients will send messages that aren't replied to, so might think the protocol is operational. 

IMO that's less an issue. The way how `config` service works should be:
1.Exporter initiates the connection;
2-1. When Agent gets updated config from outside, push it to Exporters;
2-2. If there're no updated configs, Exporter should be held there waiting for messages. The connection should be kept open all the time.

With a no-op `config` service, from Exporter's perspective, it's like after initiation Exporter enters state 2-2 directly. Even with a real implementation, it's definitely possible that after initiation Exporter won't get messages for a long time.

songy23:
Thanks for reviewing. Just had a quick chat, since we mentioned `config` is optional maybe it's better to not correlate the `export` and `config` connections. For now let's hold off this PR and I'll update the Java Exporter.

bogdandrutu:
Any updates on this?

songy23:
This is not needed for now since we agreed that we shouldn't correlate `Export` and `Config` RPCs.

PR #213, opened 2018-11-19
receiver/zipkin: consolidate server starting to receiver

### LINKED ISSUE #212: receiver/zipkin: move Server starting logic to StartTraceReception
The Zipkin receiver currently just implements http.Handler so has method ServeHTTP and it is the duty of the caller to attach it to a server of their making. However, for ease of use we should let the Zipkin receiver start the server itself so that New takes in the binding address and on invoking StartTraceReception, the server is started.

odeke-em:
Moved the server starting logic to the receiver and updated
New to
  New(address string) (*ZipkinReceiver, error)
instead of
  New(spansink.Sink) (*ZipkinReceiver, error)

This change ensures that after invoking New all the user
has to do is then just invoke:
   zr.StartTraceReception(context.Context, spansink.Sink) error

and then for cleanup
    defer zr.StopTraceReception(context.Context)

Fixes #212

### DIFF
--- a/receiver/zipkin/trace_receiver_test.go
+++ b/receiver/zipkin/trace_receiver_test.go
@@ -156,11 +156,7 @@ func TestConversionRoundtrip(t *testing.T) {
   }
 }]`)
 
-	sink := new(noopSink)
-	zi, err := New(sink)
-	if err != nil {
-		t.Fatalf("Failed to create the Zipkin receiver: %v", err)
-	}
+	zi := &ZipkinReceiver{spanSink: new(noopSink)}
 	ereqs, err := zi.parseAndConvertToTraceSpans(receiverInputJSON, nil)
 	if err != nil {
 		t.Fatalf("Failed to parse and convert receiver JSON: %v", err)
--- a/cmd/ocagent/main.go
+++ b/cmd/ocagent/main.go
@@ -178,25 +178,16 @@ func runJaegerReceiver(collectorThriftPort, collectorHTTPPort int, sr receiver.T
 }
 
 func runZipkinReceiver(addr string, sr receiver.TraceReceiverSink) (doneFn func() error, err error) {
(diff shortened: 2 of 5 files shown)


odeke-em:
>receiver/zipkin/trace_receiver.go:112:1: receiver name zi should be consistent with previous receiver name zr for ZipkinReceiver

golint is ridiculous!

bogdandrutu:
@odeke-em I disagree that golint is ridiculous. I think consistency is very important for our codebase and ensuring that will help us long term.

receiver/zipkin/trace_receiver_test.go
```diff
@@ -156,11 +156,7 @@ func TestConversionRoundtrip(t *testing.T) {
   }
 }]`)
 
-	sink := new(noopSink)
-	zi, err := New(sink)
-	if err != nil {
-		t.Fatalf("Failed to create the Zipkin receiver: %v", err)
-	}
+	zi := &ZipkinReceiver{spanSink: new(noopSink)}
```
bogdandrutu: Why not zr?

odeke-em:
>@odeke-em I disagree that golint is ridiculous. I think consistency is very important for our codebase and ensuring that will help us long term.

@bogdandrutu yes and no, that code was in before this change and a vestige from the rename from ZipkinInterceptor to ZipkinReceiver. Really though, symbol names shouldn't break a build and IMHO enforcing that is just weird for writing code.

receiver/zipkin/trace_receiver_test.go
odeke-em: Sure, I can rename that, that code was just moved then `spanSink` added. Thanks.

odeke-em:
I'll mail a `zi` to `zr` PR separately as that's noisy for this PR.

Thank you @bogdandrutu and @pjanotti for the review.

PR #221, opened 2018-11-20
OC Proto to Jaeger Thrift translation

pjanotti:
Add a translator from OC-Proto to Jaeger Thrift.

As usual there are some questions about the translation,
they should be marked with TODOs but we may want to
follow up with some discussion about the general rules in
translation (e.g.: preserve key-value info even if it maps to
some property).

### DIFF
--- a/translator/trace/jaegerthrift_to_protospan.go
+++ b/translator/trace/jaegerthrift_to_protospan.go
@@ -163,7 +163,8 @@ func jReferencesToOCProtoLinks(jrefs []*jaeger.SpanRef) *tracepb.Span_Links {
 	for _, jref := range jrefs {
 		var linkType tracepb.Span_Link_Type
 		if jref.RefType == jaeger.SpanRefType_CHILD_OF {
-			linkType = tracepb.Span_Link_CHILD_LINKED_SPAN
+			// Wording on OC for Span_Link_PARENT_LINKED_SPAN: The linked span is a parent of the current span.
+			linkType = tracepb.Span_Link_PARENT_LINKED_SPAN
 		} else {
 			// TODO: SpanRefType_FOLLOWS_FROM doesn't map well to OC, so treat all other cases as unknown
 			linkType = tracepb.Span_Link_TYPE_UNSPECIFIED
@@ -227,15 +228,12 @@ func jtagsToAttributes(tags []*jaeger.Tag) (string, tracepb.Span_SpanKind, *trac
 		case "http.status_code", "status.code": // It is expected to be an int
 			statusCodePtr = new(int32)
 			*statusCodePtr = int32(tag.GetVLong())
-			continue
 
 		case "http.status_message", "status.message":
 			statusMessage = tag.GetVStr()
-			continue
 
 		case "message":
 			message = tag.GetVStr()
-			continue
 		}
 
 		attrib := &tracepb.AttributeValue{}
--- a/translator/trace/jaegerthrift_to_protospan_test.go
+++ b/translator/trace/jaegerthrift_to_protospan_test.go
@@ -118,12 +118,12 @@ func TestConservativeConversions(t *testing.T) {
 					Tags: []*jaeger.Tag{
 						{
 							Key:   "http.status_code",
-							VLong: func() *int64 { v := int64(5); return &v }(),
+							VLong: func() *int64 { v := int64(403); return &v }(),
 							VType: jaeger.TagType_LONG,
 						},
 						{
 							Key:   "http.status_message",
-							VStr:  func() *string { v := "cache miss"; return &v }(),
+							VStr:  func() *string { v := "Forbidden"; return &v }(),
 							VType: jaeger.TagType_STRING,
 						},
 					},
@@ -205,19 +205,19 @@ func TestConservativeConversions(t *testing.T) {
 					TraceId: []byte{0x00, 0x11, 0x12, 0x13, 0x14, 0x11, 0x11, 0x11, 0x01, 0x11, 0x11, 0x11, 0xFF, 0xFF, 0xFF, 0xFF},
 					// Ensure that the status code was properly translated
 					Status: &tracepb.Status{
-						Code:    5,
-						Message: "cache miss",
+						Code:    403,
+						Message: "Forbidden",
 					},
 					Attributes: &tracepb.Span_Attributes{
 						AttributeMap: map[string]*tracepb.AttributeValue{
 							"http.status_code": {
 								Value: &tracepb.AttributeValue_IntValue{
-									IntValue: 5,
+									IntValue: 403,
 								},
 							},
 							"http.status_message": {
 								Value: &tracepb.AttributeValue_StringValue{
-									StringValue: &tracepb.TruncatableString{Value: "cache miss"},
+									StringValue: &tracepb.TruncatableString{Value: "Forbidden"},
 								},
 							},
 						},
--- a/translator/trace/jaegerthrift_to_protospan_test.go
+++ b/translator/trace/jaegerthrift_to_protospan_test.go
@@ -22,12 +22,12 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/census-instrumentation/opencensus-service/internal/testutils"
+	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"
 
 	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
 	agenttracepb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1"
 	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
-	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"
+	"github.com/census-instrumentation/opencensus-service/internal/testutils"
 )
 
 func TestJaegerThriftBatchToOCProto(t *testing.T) {
@@ -208,6 +208,20 @@ func TestConservativeConversions(t *testing.T) {
 						Code:    5,
 						Message: "cache miss",
 					},
+					Attributes: &tracepb.Span_Attributes{
+						AttributeMap: map[string]*tracepb.AttributeValue{
+							"http.status_code": {
+								Value: &tracepb.AttributeValue_IntValue{
+									IntValue: 5,
+								},
+							},
+							"http.status_message": {
+								Value: &tracepb.AttributeValue_StringValue{
+									StringValue: &tracepb.TruncatableString{Value: "cache miss"},
+								},
+							},
+						},
+					},
 				},
 				{
 					Name:    &tracepb.TruncatableString{Value: "RPC call"},
@@ -216,6 +230,20 @@ func TestConservativeConversions(t *testing.T) {
 					Status: &tracepb.Status{
 						Code:    13,
 						Message: "proxy crashed",
+					},
+					Attributes: &tracepb.Span_Attributes{
+						AttributeMap: map[string]*tracepb.AttributeValue{
+							"status.code": {
+								Value: &tracepb.AttributeValue_IntValue{
+									IntValue: 13,
+								},
+							},
+							"status.message": {
+								Value: &tracepb.AttributeValue_StringValue{
+									StringValue: &tracepb.TruncatableString{Value: "proxy crashed"},
+								},
+							},
+						},
 					},
 				},
 			},
--- a/translator/trace/protospan_to_jaegerthrift.go
+++ b/translator/trace/protospan_to_jaegerthrift.go
@@ -18,6 +18,7 @@ import (
 	"encoding/binary"
 	"fmt"
 
+	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"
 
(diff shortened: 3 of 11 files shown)


translator/trace/jaegerthrift_to_protospan.go
```diff
@@ -227,15 +228,12 @@ func jtagsToAttributes(tags []*jaeger.Tag) (string, tracepb.Span_SpanKind, *trac
 		case "http.status_code", "status.code": // It is expected to be an int
 			statusCodePtr = new(int32)
 			*statusCodePtr = int32(tag.GetVLong())
-			continue
```
odeke-em: Why did you remove this? If we are capturing the status code as per the values declared in Lines 210 and 211 we don't need to add these as attributes.

translator/trace/jaegerthrift_to_protospan.go
```diff
@@ -227,15 +228,12 @@ func jtagsToAttributes(tags []*jaeger.Tag) (string, tracepb.Span_SpanKind, *trac
...
 
 		case "http.status_message", "status.message":
 			statusMessage = tag.GetVStr()
-			continue
```
odeke-em: Ditto about this continue being removed.

translator/trace/jaegerthrift_to_protospan_test.go
```diff
@@ -208,6 +208,20 @@ func TestConservativeConversions(t *testing.T) {
 						Code:    5,
 						Message: "cache miss",
 					},
+					Attributes: &tracepb.Span_Attributes{
+						AttributeMap: map[string]*tracepb.AttributeValue{
```
odeke-em: We want to capture the status code and not capture it as an attribute otherwise there is no point in setting the status above. What's your rationale for this change?

translator/trace/jaegerthrift_to_protospan_test.go
```diff
@@ -217,6 +231,20 @@ func TestConservativeConversions(t *testing.T) {
 						Code:    13,
 						Message: "proxy crashed",
 					},
+					Attributes: &tracepb.Span_Attributes{
+						AttributeMap: map[string]*tracepb.AttributeValue{
+							"status.code": {
```
odeke-em: I don't think we want these attributes as they are, they should have been translated in status.

translator/trace/protospan_to_jaegerthrift.go
```diff
@@ -0,0 +1,321 @@
...
+	// Pre-allocate assuming that few, if any spans, are nil.
+	jSpans := make([]*jaeger.Span, 0, len(ocSpans))
+	for _, ocSpan := range ocSpans {
+		startTime := timestampToEpochMicroseconds(ocSpan.StartTime)
+		jSpan := &jaeger.Span{
+			SpanId:        ocIDBytesToJaegerID(ocSpan.SpanId),
+			ParentSpanId:  ocIDBytesToJaegerID(ocSpan.ParentSpanId),
+			OperationName: truncableStringToStr(ocSpan.Name),
+			References:    ocLinksToJaegerReferences(ocSpan.Links),
+			// Flags: TODO (@pjanotti) at first nothing matches to it.
```
odeke-em: I don't understand this comment.

translator/trace/protospan_to_jaegerthrift.go
```diff
@@ -0,0 +1,321 @@
...
+func truncableStringToStr(ts *tracepb.TruncatableString) string {
+	if ts == nil {
+		return ""
+	}
+	return ts.Value
+}
+
+func traceIDBytesToLowAndHigh(traceID []byte) (traceIDLow, traceIDHigh int64) {
+	if len(traceID) != 16 {
+		return
```
odeke-em: It would be useful to return an error instead of silently passing as TraceID is a very important field.

translator/trace/protospan_to_jaegerthrift.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	traceIDHigh = int64(binary.BigEndian.Uint64(traceID[:8]))
+	traceIDLow = int64(binary.BigEndian.Uint64(traceID[8:]))
+	return
+}
+
+func ocIDBytesToJaegerID(b []byte) (id int64) {
+	if len(b) != 8 {
+		return
```
odeke-em: Same here it'd be nice to return an error.

translator/trace/protospan_to_jaegerthrift_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
+	agenttracepb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1"
+	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
+	"github.com/census-instrumentation/opencensus-service/internal/testutils"
+	"github.com/golang/protobuf/ptypes/timestamp"
+	"github.com/jaegertracing/jaeger/thrift-gen/jaeger"
+)
+
+func TestTraceIDRoundTripJaegerOCProtoJaeger(t *testing.T) {
```
odeke-em: Perhaps make the prefix `TestJaegerFromOCProto` otherwise the test's name becomes hard to parse since the reader has to examine the suffix instead of prefix which humans find easier scanning.

translator/trace/protospan_to_jaegerthrift_test.go
```diff
@@ -0,0 +1,245 @@
...
+				"a.binary": "AQIDBAMCAQ==",
+				"a.bool":   "true",
+				"a.double": "1234.56789",
+				"a.long":   "123456789",
+				"ip":       "10.53.69.61",
+			},
+		},
+		Spans: []*tracepb.Span{
+			{
+				TraceId:      base64ToBytes("AAAAAAAAAABSlpqJVVcaPw=="),
```
odeke-em: Why are we using base64? I think just plain bytes are easier to parse and easy to examine as a human and for example this is 24 bytes which becomes impossible to eyeball and catch conversion bugs. Please take a look at other tests for how we use plain `[]byte{0x01,....}`.

translator/trace/protospan_to_jaegerthrift_test.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		gjson, _ := json.Marshal(gotJBatch)
+		wjson, _ := json.Marshal(wantJBatch)
+		gjsonStr := testutils.GenerateNormalizedJSON(string(gjson))
+		wjsonStr := testutils.GenerateNormalizedJSON(string(wjson))
+		if gjsonStr != wjsonStr {
+			t.Errorf("OC Proto to Jaeger Thrift failed.\nGot:\n%s\nWant:\n%s\n", gjsonStr, wjsonStr)
+			t.Errorf("Saving formated jsons at ./testdata/ to help with further investigation")
+			testutils.SaveAsFormattedJSON(fmt.Sprintf("./testdata/failure_oc_to_jaeger_got_%02d.json", i), gotJBatch)
```
odeke-em: I don't think we want tests writing arbitrary JSON to file systems. Please remove this unless we clear these files on exit but this might not be pleasant to users who clone our project and get a bug.

translator/trace/jaegerthrift_to_protospan.go
pjanotti: The issue is that the OC spec indicates that we should be actually mapping the values to the OC ones (see https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#mapping-from-http-status-codes-to-trace-status-codes), with that it is not possible to restore the original keys from OC -> Jaeger, so dropping the key-value pairs is losing information that some systems in the backend may be relying upon.

At this stage it seems that a better general approach is: it is ok/desirable to lift information from known key-value pairs but we should preserve original pairs since in the backend there may be systems relying on these tags "http.status_code" seems a very likely one for this case.

translator/trace/protospan_to_jaegerthrift.go
pjanotti: Jaeger uses this as a bit flags but I didn't find the related/equivalent in OC proto, so it seems that the info that they carry is lost. I will reword it.

translator/trace/protospan_to_jaegerthrift_test.go
pjanotti: If a change breaks the test this should speedup the fixing it. It will be visible as untracked file together with the other test files.

translator/trace/jaegerthrift_to_protospan.go
odeke-em: Oh okay I see, thanks for the references. As a minor point, in that case we need to then separate `http.status_code` and translate them to OpenCensus status codes.

translator/trace/protospan_to_jaegerthrift_test.go
odeke-em: I am uncomfortable with writing arbitrary files and leaving them on user's disks. In many projects I've encountered this as a discomfort and actually last week, I helped clean up such in the Go project from a stray write https://github.com/golang/go/issues/28781. We can print them to stdout.

translator/trace/protospan_to_jaegerthrift.go
```diff
@@ -0,0 +1,360 @@
...
+		hostTag := &jaeger.Tag{
+			Key:   "hostname",
+			VType: jaeger.TagType_STRING,
+			VStr:  &node.Identifier.HostName,
+		}
+		jTags = append(jTags, hostTag)
+	}
+
+	// Add OpenCensus library information as tags if available
+	ocLib := node.LibraryInfo
```
odeke-em: Nice!

translator/trace/protospan_to_jaegerthrift.go
```diff
@@ -0,0 +1,360 @@
...
+			jTag := &jaeger.Tag{
+				Key:   k,
+				VType: jaeger.TagType_STRING,
+				VStr:  &str,
+			}
+			jTags = append(jTags, jTag)
+		}
+	}
+
+	if node.Identifier != nil && node.Identifier.HostName != "" {
```
odeke-em: Cool! How about Pid and StartTimestamp?

Field|Reference
---|---
Pid|[ProcessIdentifier.Pid](https://godoc.org/github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1#ProcessIdentifier.Pid)
StartTimestamp|[ProcessIdentifier.StartTimestamp](https://godoc.org/github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1#ProcessIdentifier.StartTimestamp)

translator/trace/jaegerthrift_to_protospan_test.go
```diff
@@ -208,6 +208,20 @@ func TestConservativeConversions(t *testing.T) {
 						Code:    5,
 						Message: "cache miss",
 					},
+					Attributes: &tracepb.Span_Attributes{
+						AttributeMap: map[string]*tracepb.AttributeValue{
+							"http.status_code": {
+								Value: &tracepb.AttributeValue_IntValue{
+									IntValue: 5,
```
odeke-em: In that case then let's use real status codes i.e. >=1XX

translator/trace/jaegerthrift_to_protospan.go
pjanotti: >  in that case we need to then separate http.status_code and translate them to OpenCensus status codes.

Agreed, created #238 to track it.

translator/trace/protospan_to_jaegerthrift_test.go
pjanotti: Think of these files as penalty for someone breaking the tests 😃 , but that is fine, I can quickly add this code if I want to compare the files. I will remove it and the related function.

translator/trace/protospan_to_jaegerthrift.go
pjanotti: Added.

translator/trace/jaegerthrift_to_protospan_test.go
pjanotti: Done.

translator/trace/protospan_to_jaegerthrift_test.go
odeke-em: LOLOL

translator/trace/protospan_to_jaegerthrift.go
```diff
@@ -0,0 +1,381 @@
...
+
+	return ocSpanAttributesToJaegerTags(annotation.Attributes)
+}
+
+func ocMessageEventToJaegerTags(msgEvent *tracepb.Span_TimeEvent_MessageEvent) []*jaeger.Tag {
+	if msgEvent == nil {
+		return nil
+	}
+
+	// TODO: (@pjanotti) Not clear how to map those to Jaeger, perhaps some OpenTracing tags...
```
odeke-em: I believe those map as OpenTracing "Span Logs", please perhaps file an issue otherwise we'll be losing some critical information :)

translator/trace/protospan_to_jaegerthrift.go
```diff
@@ -0,0 +1,381 @@
...
+
+	return jLogs
+}
+
+func ocAnnotationToJagerTags(annotation *tracepb.Span_TimeEvent_Annotation) []*jaeger.Tag {
+	if annotation == nil {
+		return nil
+	}
+
+	// TODO: (@pjanotti) what about Description? Does it fit as another tag?
```
odeke-em: Please see my comment about MessageEventToJaegerTags being "Span Logs" for Jaeger

translator/trace/protospan_to_jaegerthrift.go
```diff
@@ -0,0 +1,381 @@
...
+		}
+		jReferences, err := ocLinksToJaegerReferences(ocSpan.Links)
+		if err != nil {
+			return nil, fmt.Errorf("Error converting OC links to Jaeger references: %v", err)
+		}
+		spanID, err := ocIDBytesToJaegerID(ocSpan.SpanId)
+		if err != nil {
+			return nil, fmt.Errorf("OC span has invalid span ID: %v", err)
+		}
+		parentSpanID, err := ocIDBytesToJaegerID(ocSpan.ParentSpanId)
```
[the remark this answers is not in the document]
pjanotti: Since there will be special cases for the IDs on the function that understand the difference between ParentSpanId, SpanId and so on, I will do a bit different but overall with same intended effect.

translator/trace/protospan_to_jaegerthrift.go
odeke-em: Perfect, thanks!

PR #234, opened 2018-11-27
cmd/ocagent: fix Zipkin logical conflict detection typo

### LINKED ISSUE #233: cmd/ocagent: wrong Zipkin receiver vs exporter logical conflict detection
Currently with a configuration such as
```yaml
receivers:
    zipkin:
        address: "localhost:9410"

exporters:
    zipkin:
```

and then running either cmd/ocagent or cmd/occollector will result in
```shell
2018/11/26 18:40:53 Configuration logical error: ZipkinExporter address ("localhost:9411") aka (localhost on port 9411)
is the same as the receiver address ("localhost:9410") aka (localhost on port 9411)
```
which is clearly a wrong error "localhost:9410" is not "localhost:9411" but even worse

```shell
is the same as the receiver address ("localhost:9410") aka (localhost on port 9411)
```
Which is a consequence of a typographic error in https://github.com/census-instrumentation/opencensus-service/blob/5bd378023295ee43a9c6c0812a857d9022b7b7a6/cmd/ocagent/config.go#L186

which was reported by @pjanotti in PR https://github.com/census-instrumentation/opencensus-service/pull/229

I am filing this issue separately to make a focused fix and keep changes minimal and testable but to also allow for future regression isolation.

I'll mail a PR shortly.

odeke-em:
The Zipkin logical conflict detection had 2 errors:
a) By mistake, it resolved the receiver{Host, Port}
from the exporterHostPort address
b) During IP resolution to check say if "localhost" and "//zipkin"
were the same, if the results were the same, the ports weren't
checked for equivalence hence if we had:
    "localhost:9410" and "127.0.0.1:9411"
it would incorrectly report a logical conflict.

Added a regression test as well.

Fixes #233

### DIFF
--- a/cmd/ocagent/config.go
+++ b/cmd/ocagent/config.go
@@ -183,7 +183,7 @@ func (c *config) checkLogicalConflicts(blob []byte) error {
 			zExporterHostPort, zReceiverHostPort)
 	}
 	zExpHost, zExpPort, _ := net.SplitHostPort(zExporterHostPort)
-	zReceiverHost, zReceiverPort, _ := net.SplitHostPort(zExporterHostPort)
+	zReceiverHost, zReceiverPort, _ := net.SplitHostPort(zReceiverHostPort)
 	if eqHosts(zExpHost, zReceiverHost) && zExpPort == zReceiverPort {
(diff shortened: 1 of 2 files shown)


odeke-em:
Thank you @pjanotti for the review!

Issue #181, opened 2018-11-08
PR Travis run failure: data race on TestZipkinExportersFromYAML_roundtripJSON


pjanotti:
This was hit during a PR Travis run but didn't repro consistently, details: https://travis-ci.org/census-instrumentation/opencensus-service/builds/452159703#L871

It is related to the buffer `buf` used with the http test server, see https://github.com/census-instrumentation/opencensus-service/blob/c99676e914a01e576f6054b038c6cc02fc9fe548/exporter/exporterparser/zipkin_test.go#L43-L49 and when it is read later at https://github.com/census-instrumentation/opencensus-service/blob/c99676e914a01e576f6054b038c6cc02fc9fe548/exporter/exporterparser/zipkin_test.go#L112-L113

odeke-em:
Thanks for filing this @pjanotti! This race was introduced by https://github.com/census-instrumentation/opencensus-service/pull/156  particularly https://github.com/census-instrumentation/opencensus-service/pull/156/files#diff-925aacb58ff0d51b1eee965a21f0c5a4L44 which changed from a concurrent buffer to a plain buffer

The fix perhaps is to revert that part of the PR(which seemed like an unrelated change to add a Docker build) and to keep such buffers concurrency safe.

pjanotti:
That definitely will fix the issue, but the race seems because the server is still running, so if we close it we should be good (one line clean change). The problem is that to validate my theory I need a consistent way to repro it... let me see if I can get one way to do it...

odeke-em:
A guaranteed solution that will never race would instead to make that channel `make(chan []byte)` and inside that server do
```go
 cst := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { 
 	blob, _ := ioutil.ReadAll(r.Body)
 	r.Body.Close() 
 	responseCh <- blob
 })) 
 defer cst.Close() 
```

odeke-em:
or a variant of it that parses even the error
```go
type reqBodyAndErr{
   body []byte
   err error
}

 buf := new(bytes.Buffer) 
 cst := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { 
 	blob, err := ioutil.ReadAll(r.Body)
 	r.Body.Close() 
 	responseCh <- &reqBodyAndErr{body: blob, err: err}
 })) 
 defer cst.Close() 
```

pjanotti:
[...]

PR #235, opened 2018-11-27
exporterparser/zipkin: safely retrieve remoteEndpoint serviceName

### LINKED ISSUE #230: exporter/zipkin: missing endpoints in payload cause a panic
Steps to reproduce:

* Download latest collector
* Configure to receive OC agent (--receiver-oc-trace)
* Configure debug-tracer with log level set to debug
* Configure to send to a Zipkin exporter
* Run example/main.go pointed at collector

Expected: Traces are received by Zipkin exporter

Actual: crash seen in OC collector logs

odeke-em:
Safely retrieve the servicename for the remoteEndpoint if no
endpoints were sent in the payload hence a nil Node.

A reduction of PR #229 and originally @pjanotti's endeavor.

Fixes #230

### DIFF
--- a/exporter/exporterparser/zipkin.go
+++ b/exporter/exporterparser/zipkin.go
@@ -132,6 +132,13 @@ func newZipkinExporter(finalEndpointURI, defaultServiceName, defaultLocalEndpoin
 	return zle, nil
 }
 
+func lookupAttribute(node *commonpb.Node, key string) string {
+	if node == nil {
+		return ""
+	}
+	return node.Attributes[key]
+}
+
 func zipkinEndpointFromNode(node *commonpb.Node, serviceName string, endpointType zipkinDirection) (*zipkinmodel.Endpoint, error) {
 	if node == nil {
 		return nil, nil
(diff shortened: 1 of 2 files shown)


exporter/exporterparser/zipkin.go
```diff
@@ -132,6 +132,13 @@ func newZipkinExporter(finalEndpointURI, defaultServiceName, defaultLocalEndpoin
 	return zle, nil
 }
 
+func lookupAttribute(node *commonpb.Node, key string) string {
```
pjanotti: personal preference: nothing against small functions (on the contrary) but I tend to add one that is used just in one place, but I am fine as it is.

exporter/exporterparser/zipkin_test.go
```diff
@@ -112,6 +112,15 @@ exporters:
     {"timestamp": 1472470996403000,"value": "bar"}
   ],
   "tags": {"http.path": "/api","clnt/finagle.version": "6.45.0"}
+},
```
pjanotti: side note: in principle this represents a slightly increase in the probability of the sporadic test failure that we saw in  #181

PR #237, opened 2018-11-27
cmd/ocagent: remove vestigial code

odeke-em:
The code for protoToSpanData comes from
the day of a prototype that used this repository
but has since diverged since we reused it for
the OpenCensus Agent/Service.


odeke-em:
Thank you @pjanotti and @bogdandrutu for the reviews!

PR #239, opened 2018-11-28
cmd/ocagent: add OpenCensus Metrics receiver + noopMetricsSink

### LINKED ISSUE #216: cmd: add OpenCensus metrics receiver.
As a follow-up of https://github.com/census-instrumentation/opencensus-service/issues/136. Currently blocked by https://github.com/census-instrumentation/opencensus-service/issues/100.

odeke-em:
For now before any OpenCensus Metrics exporter is created, just
hook it up. This will allow applications such as the OpenCensus-PHP
stats backend to send content.

Fixes #216

### DIFF
--- a/cmd/ocagent/main.go
+++ b/cmd/ocagent/main.go
@@ -171,24 +171,14 @@ func runOCReceiver(acfg *config.Config, sr receiver.TraceReceiverSink, mr receiv
 
 	ctx := context.Background()
 
-	switch {
-	case acfg.CanRunAllOpenCensusReceivers():
-		if err := ocr.StartTraceReception(ctx, sr); err != nil {
-			return nil, fmt.Errorf("Failed to start OpenCensus Trace receiver: %v", err)
-		}
-		if err := ocr.StartMetricsReception(ctx, mr); err != nil {
-			ocr.Stop()
-			return nil, fmt.Errorf("Failed to start OpenCensus Metrics receiver: %v", err)
-		}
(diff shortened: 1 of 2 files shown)


odeke-em:
Kindly pinging @pjanotti!

cmd/ocagent/main.go
```diff
@@ -157,10 +169,32 @@ func runOCReceiver(addr string, sr receiver.TraceReceiverSink) (doneFn func() er
 		return nil, fmt.Errorf("Failed to register ocgrpc.DefaultServerViews: %v", err)
 	}
 
-	if err := ocr.StartTraceReception(context.Background(), sr); err != nil {
-		return nil, fmt.Errorf("Failed to start TraceReceiver: %v", err)
+	ctx := context.Background()
+
+	switch {
+	case acfg.CanRunAllOpenCensusReceivers():
```
pjanotti: Why have the CanRunAllOpenCensusReceivers as a separate function? At first glance the only thing different in this case is the call to ocr.Stop if starting metrics fail and traces were started. It seems that if statements will avoid the the very similar code between the cases. The single log message can also be handled (although I don't think it is worth the trouble).

odeke-em: Aha you read my mind, I now see it isn't worth it making CanRunAllOpenCensusReceivers, I'll revert that, thank you!

odeke-em:

odeke-em:
@basvanbeek you can now start sending metrics to the agent for OpenCensus PHP stats.

PR #244, opened 2018-12-03
receiver/opencensus: provide a Start method for both Trace and Metrics

### LINKED ISSUE #243: cmd/ocagent: OpenCensus receiver gRPC server error for registering after (*grpc).Server.Serve
Currently running ocagent with this YAML configuration
```yaml
receivers:
    opencensus:
        address: "127.0.0.1:55678"

exporters:
    zipkin:
```
gives 
```shell
$ ./bin/ocagent_darwin --config t.yaml 
2018/12/03 00:29:12 Running OpenCensus Trace receiver as a gRPC service at "127.0.0.1:55678"
FATAL: 2018/12/03 00:29:12 grpc: Server.RegisterService after Server.Serve for "opencensus.proto.agent.metrics.v1.MetricsService"
```

where the error starts from

https://github.com/census-instrumentation/opencensus-service/blob/b093b60e9ad4929a2ed5b22f8fa5f0399af5b69e/cmd/ocagent/main.go#L174-L186

of which the root cause is that when invoking either of StartTraceReception or StartMetricsReception, we invoke .Serve on the gRPC server which requires that all services be registered before invoking .Serve.

### Proposed fix
Now I recall my previous train of thought, for which the fix should be:
a) If both trace and metrics are enabled, then invoke a Start method on the receiver that registers both services before finally invoking (*grpc.Server).Serve
b) Register either of those services and then finally a method that invokes (*grpc.Server).Serve

odeke-em:
Provide a Start method on the OpenCensus receiver to allow registering
all services before running the gRPC server.

Unfortunately gRPC's RegisterService doesn't return an error so
it has no choice but to Fatalf as per
https://github.com/grpc/grpc-go/blob/ca62c6b92c334f52c7e48b7cbf624f3b877fb092/server.go#L408-L410

By design, in Go we cannot catch a .Fatalf that invokes os.Exit(code)
so even a test is worthless to test this.

Fixes #243

### DIFF
--- a/cmd/ocagent/main.go
+++ b/cmd/ocagent/main.go
@@ -171,14 +171,20 @@ func runOCReceiver(acfg *config.Config, sr receiver.TraceReceiverSink, mr receiv
 
 	ctx := context.Background()
 
-	if acfg.CanRunOpenCensusTraceReceiver() {
+	switch {
+	case acfg.CanRunOpenCensusTraceReceiver() && acfg.CanRunOpenCensusMetricsReceiver():
+		if err := ocr.Start(ctx, sr, mr); err != nil {
+			return nil, fmt.Errorf("Failed to start Trace and Metrics Receivers: %v", err)
+		}
+		log.Printf("Running OpenCensus Trace and Metrics receivers as a gRPC service at %q", addr)
+
+	case acfg.CanRunOpenCensusTraceReceiver():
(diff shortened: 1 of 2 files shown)


odeke-em:
Thank you for the review @songy23 for the review!

>LGTM, please fix golint.

Yes, I fixed it, I just had to add some comments.

PR #246, opened 2018-12-04
receiver/jaeger: Jaeger Agent running on UDP ports

odeke-em:
* Update receiver/jaeger.New to take in a Configuration
that now allows specifying the Jaeger agent UDP ports
* Separated out startCollector and startAgent helpers
for the Jaeger receiver
* Agent now runs on UDP ports:
+ Thrift Compact: by default 6831
+ Thrift Binary: by default 6832
+ Agent server: by default 5778

### DIFF
--- a/cmd/ocagent/main.go
+++ b/cmd/ocagent/main.go
@@ -196,7 +196,14 @@ func runOCReceiver(acfg *config.Config, sr receiver.TraceReceiverSink, mr receiv
 }
 
 func runJaegerReceiver(collectorThriftPort, collectorHTTPPort int, sr receiver.TraceReceiverSink) (doneFn func() error, err error) {
-	jtr, err := jaeger.New(context.Background(), collectorThriftPort, collectorHTTPPort)
+	jtr, err := jaeger.New(context.Background(), &jaeger.Configuration{
+		CollectorThriftPort: collectorThriftPort,
+		CollectorHTTPPort:   collectorHTTPPort,
+
+		// TODO: (@odeke-em, @pjanotti) send a change
+		// to dynamically retrieve the Jaeger Agent's ports
+		// and not use their defaults of 5778, 6831, 6832
+	})
 	if err != nil {
 		return nil, fmt.Errorf("Failed to create new Jaeger receiver: %v", err)
 	}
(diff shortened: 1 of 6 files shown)


odeke-em:
The compact Thrift processor works great but unfortunately there is a seemingly elusive problem with the binary Thrift agent processor. That is
```
Bad version in ReadMessageBegin
```
when we try to send using the following test program
```go
package main

import (
        "context"
        "math/rand"
        "time"

        "go.opencensus.io/exporter/jaeger"
        "go.opencensus.io/trace"
)

func main() {
        je, _ := jaeger.NewExporter(jaeger.Options{
                AgentEndpoint: "localhost:6832",
        })
        trace.RegisterExporter(je)
        trace.ApplyConfig(trace.Config{
                DefaultSampler: trace.AlwaysSample(),
        })

        rng := rand.New(rand.NewSource(time.Now().UnixNano()))
        for {
                _, span := trace.StartSpan(context.Background(), "Goose")
                time.Sleep(time.Duration(rng.Int63n(1000)) * time.Millisecond)
                span.End()
        }
}
```

I've ensured that `go.opencensus.io/exporter/jaeger` and the ocagent binary are using the same version of Thrift. Hmm.

odeke-em:
After countless hours and ripping through every Jaeger dependency, turns out that perhaps the version of the Thrift library that is vendored by the Go library is what's tripping things up and causing that "Bad version in ReadMessageBegin". I'll file a bug on OpenCensus-Go but sheesh for now I'll disable the Thrift compact test. I can verify that things work with OpenCensus-Python's Jaeger program at
```python
#!/usr/bin/env python

from opencensus.trace.exporters.jaeger_exporter import JaegerExporter
from opencensus.trace.tracer import Tracer

def main():
    je = JaegerExporter(service_name="service-b",
                        host_name='localhost',
                        agent_port=6831,
                        endpoint='/api/traces')

    tracer = Tracer(exporter=je)
    with tracer.span(name='doingWork') as span:
        for i in range(10):
            continue

if __name__ == "__main__":
    main()
```
which produces 
screen shot 2018-12-05 at 9 51 51 pm

odeke-em:

@flands please go ahead with the release. I opened https://github.com/census-instrumentation/opencensus-service/issues/262 to track and resolve the Thrift dependency problems that we'll coordinate with OpenCensus-Go and Uber's Jaeger project.

Issue #218, opened 2018-11-20
Unable to build repo


akhil-karanth:
I am trying to build the ocagent using the followign command as specified in the documentation

```
./build_binaries.sh binaries
```

I am getting the following error:

```# github.com/uber/tchannel-go/thrift/gen-go/meta
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:79:20: not enough arguments in call to oprot.Flush
	have ()
	want (context.Context)
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:103:15: assignment mismatch: 2 variables but 1 values
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:152:20: not enough arguments in call to oprot.Flush
	have ()
	want (context.Context)
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:176:15: assignment mismatch: 2 variables but 1 values
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:225:20: not enough arguments in call to oprot.Flush
	have ()
	want (context.Context)
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:249:15: assignment mismatch: 2 variables but 1 values
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:295:31: cannot use metaProcessorHealth literal (type *metaProcessorHealth) as type thrift.TProcessorFunction in assignment:
	*metaProcessorHealth does not implement thrift.TProcessorFunction (wrong type for Process method)
		have Process(int32, thrift.TProtocol, thrift.TProtocol) (bool, thrift.TException)
		want Process(context.Context, int32, thrift.TProtocol, thrift.TProtocol) (bool, thrift.TException)
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:296:34: cannot use metaProcessorThriftIDL literal (type *metaProcessorThriftIDL) as type thrift.TProcessorFunction in assignment:
	*metaProcessorThriftIDL does not implement thrift.TProcessorFunction (wrong type for Process method)
		have Process(int32, thrift.TProtocol, thrift.TProtocol) (bool, thrift.TException)
		want Process(context.Context, int32, thrift.TProtocol, thrift.TProtocol) (bool, thrift.TException)
../../uber/tchannel-go/thrift/gen-go/meta/meta.go:297:36: cannot use metaProcessorVersionInfo literal (type *metaProcessorVersionInfo) as type thrift.TProcessorFunction in assignment:
[...]

PR #247, opened 2018-12-04
README: Update usage commands.

songy23:
Updates https://github.com/census-instrumentation/opencensus-service/issues/218.

Set env var `GO111MODULE` to on before make agent/collector. Otherwise users will get errors like https://github.com/census-instrumentation/opencensus-service/issues/218.

Also reported offline by @rghetia.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,8 @@ Architecture amd64
 
 ### <a name="agent-usage"></a>Usage
 
+The minimum Go version required for this project is Go1.11.
+
 First, install ocagent if you haven't.
 
 ```shell
@@ -296,6 +298,8 @@ agent/client health information/inventory metadata to downstream exporters.
 
 The collector is in its initial development stages. It can be run directly
 from sources, binary, or a Docker image.
+
+The minimum Go version required for this project is Go1.11.
 
 1. Run from sources:
 ```shell
--- a/README.md
+++ b/README.md
@@ -104,6 +104,12 @@ First, install ocagent if you haven't.
 $ go get github.com/census-instrumentation/opencensus-service/cmd/ocagent
 ```
 
+Alternatively you can build ocagent from binary (from the root of your repo):
+
+```shell
+$ export GO111MODULE=on && make agent
+```
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -104,6 +104,12 @@ First, install ocagent if you haven't.
 $ go get github.com/census-instrumentation/opencensus-service/cmd/ocagent
 ```
 
+Alternatively you can build ocagent from binary (from the root of your repo):
+
+```shell
+$ export GO111MODULE=on && make agent
```
odeke-em: Setting `GO111MODULE=on` can trip out some builds so perhaps
```shell
$ GO111MODULE=on make agent
```
which is the same remedy I had recommended for him offline.

README.md
```diff
@@ -297,10 +303,10 @@ $ go run github.com/census-instrumentation/opencensus-service/cmd/occollector
 ```
 2. Run from binary (from the root of your repo):
 ```shell
-$ make collector
+$ export GO111MODULE=on && make collector
```
odeke-em: ditto

README.md
songy23: I see, I moved setting `GO111MODULE` to Makefile instead, which should work the same way.

odeke-em:
Oh @songy23 one last one, could you also state somewhere in the README that the Go minimum version is Go1.11?

README.md
```diff
@@ -104,6 +104,12 @@ First, install ocagent if you haven't.
 $ go get github.com/census-instrumentation/opencensus-service/cmd/ocagent
 ```
 
+Alternatively you can build ocagent from binary (from the root of your repo):
+
+```shell
+$ make agent
```
odeke-em: So even if they use `go get` without `GO111MODULE=on` it'll trip out unfortunately.

songy23:
> could you also state somewhere in the README that the Go minimum version is Go1.11?

Sure, added.

> So even if they use go get without GO111MODULE=on it'll trip out unfortunately.

Also add to `go get` commands.

songy23:
Thanks!

pjanotti:
Thanks @songy23!

PR #253, opened 2018-12-05
Receive jaeger-agent tchannel on port 14267

### LINKED ISSUE #245: cmd/occollector: TChannel used by jaeger-agent (default port: 14267) to send spans is not working
The endpoint used by the jaeger-agent to send to send spans in jaeger.thrift format is not working. The fix itself is trivial, just need to properly name the channel (change it to "jaeger-collector" and register a handler for the jaeger spans on the TChannel server - writing proper tests shouldn't be hard.

``` golang
		tch, terr := tchannel.NewChannel("jaeger-collector", new(tchannel.ChannelOptions)) // <- changed the name of the channel
		if terr != nil {
			err = fmt.Errorf("Failed to create NewTChannel: %v", terr)
			return
		}

		// Add the handler to the Jaeger batches.
		server := thrift.NewServer(tch)
		server.Register(jaeger.NewTChanCollectorServer(jr))
```

The jaeger-agent also can send Zipkin data via this channel. If we want that capability we should do it as a separate issue.

/cc @flands

pjanotti:
Enable the TChannel receiver for Jaeger.

@odeke-em per our earlier conversation I don't have bandwidth rightnow to write a test for it. I also see that is going to conflict with the PR that you have open, either one can go first since the fix is simple. It will be good if we get this in before we cut another version.

Fixes #245

### DIFF
--- a/receiver/jaeger/trace_receiver.go
+++ b/receiver/jaeger/trace_receiver.go
@@ -94,11 +94,14 @@ func (jr *jReceiver) StartTraceReception(ctx context.Context, spanSink receiver.
 
 	var err = errAlreadyStarted
 	jr.startOnce.Do(func() {
-		tch, terr := tchannel.NewChannel("recv", new(tchannel.ChannelOptions))
+		tch, terr := tchannel.NewChannel("jaeger-collector", new(tchannel.ChannelOptions))
 		if terr != nil {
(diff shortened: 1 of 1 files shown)


pjanotti:
Thanks for reviewing it @odeke-em!

PR #269, opened 2018-12-10
Change Jaeger exporter from collector-endpoint to endpoint

flands:
All other exporters are using endpoint as the configuration parameter.
To standardize, update Jaeger to follow the same pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ exporters:
     endpoint: "http://localhost:10001"
 
   jaeger:
-    collector_endpoint: "http://localhost:14268/api/traces"
+    endpoint: "http://localhost:14268/api/traces"
 
   kafka:
     brokers: ["localhost:9092"]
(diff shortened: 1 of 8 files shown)


odeke-em:
Jaeger has both an agent and a collector, that's why that configuration has that name "collector_endpoint"

odeke-em:
>@odeke-em this is true, but not sure I follow why it requires a dedicated configuration parameter. Whether agent or collector you need to specify a destination. If sending to agent then not sending to collector and vice versa. Also, the exporter only supports the collector today.

@flands the Jaeger exporter can send via any of them: with this current change we can't disambiguate between agent and collector if we used "endpoint". I think trying to unify options for exporters might not be appropriate as in this case and would complicate the configuration. Even worse than that we can have binary vs compact thrift UDP endpoints.

flands:
Superseded by #331

PR #270, opened 2018-12-12
Add grpc-gateway for OpenCensus trace receiver

draffensperger:
This will allow clients that do not support gRPC (e.g. browsers) to write traces to the agent.

This uses the `cmux` package to auto-detect whether a request is gRPC or HTTP/JSON and so reuses the same `address` for both. This also updates `opencensus-proto` to be pinned to a git commit to bring in the new generated code for the grpc-gateway.

To test this change, run the agent locally with an exporter configured and send it a HTTP-JSON request to its HTTP port e.g.:

```bash
START="$(date -u +"%Y-%m-%dT%H:%M"):00Z"
END="$(date -u +"%Y-%m-%dT%H:%M"):01Z"
read -r -d '' REQUEST_JSON << EOF
{
   "node":{"identifier":{"hostName":"test"}},
   "spans":[
      {
          "traceId":"W47/95gDgQPSabYzgT/GDA==",
          "spanId":"7uGbfsPBsXM=",
          "name":{"value":"test6"},
          "startTime":"$START",
          "endTime":"$END"
      }
   ]
}
EOF
curl -H "Content-Type: application/json" --data "$REQUEST_JSON" localhost:55681/v1/trace
```

Note that the `traceId` and `spanId` fields are base64 encoded, which matches the [standard proto JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) for bytes that grpc-gateway follows.


songy23:
Please fix the build.

draffensperger:
OK, build is fixed and I also added an end to end test for the trace grpc-gateway functionality. That involved refactoring out a `testhelper` package from the Jaeger tests.

internal/config/config.go
```diff
@@ -113,6 +114,21 @@ func (c *Config) OpenCensusReceiverAddress() string {
 	return inCfg.OpenCensus.Address
 }
 
+// OpenCensusReceiverHTTPPort is a helper to safely retrieve the port
+// that the OpenCensus receiver grpc-gateway will be bound to.
+// If Config is nil or the OpenCensus receiver's configuration is nil, it
+// will be set to the default value of 55680.
```
pjanotti: nit default is 55681

receiver/opencensus/opencensus.go
```diff
@@ -53,18 +61,27 @@ var (
...
+	grpcLn, err := net.Listen("tcp", grpcAddr)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to bind to gRPC address %q: error: %v", grpcAddr, err)
+	}
+
+	httpAddr := fmt.Sprintf(":%d", httpPort)
+	httpLn, err := net.Listen("tcp", httpAddr)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to bind to address %q: error: %v", addr, err)
+		return nil, fmt.Errorf("Failed to bind to HTTP port %q: error: %v", httpAddr, err)
```
pjanotti: nit: it was already there but the "error: " is redundant, same on the error handling above.

receiver/opencensus/opencensus.go
```diff
@@ -53,18 +61,27 @@ var (
...
+	}
+
+	ocr := &Receiver{grpcLn: grpcLn, httpLn: httpLn}
+
+	for _, opt := range opts {
+		opt.WithReceiver(ocr)
```
pjanotti: Per discussion in other PR the `WithReceiver` should be private (this shouldn't cause problems and achieves the desired encapsulation).

pjanotti:
Thanks for this PR @draffensperger!

> This reuses the collector_http_port option that was added for the Jaeger receiver, which is typed to int. I can modify this to add a new option, e.g. address_http typed to string that parallels the address option if that's better.

I think this will cause some confusion as it contains "collector". As you suggested "address" seems to better match the pattern already in use.

> This also updates opencensus-proto to be pinned to a git commit. Should this wait for a new release of opencensus-proto first?

It will be better if we can have a minor release vouched by the maintainers of the repo.

internal/config/config.go
draffensperger: Thanks, fixed.

receiver/opencensus/opencensus.go
draffensperger: Good point, fixed.

receiver/opencensus/opencensus.go
draffensperger: Done, makes sense.

draffensperger:
Thanks for the review @pjanotti! I've applied fixes per your comments including making the HTTP address called `address_http` in the config and updated the type to `string` and the Go variables to match.

I will coordinate with the maintainers of the `opencensus-proto` repo to see if we can get a new minor release and then update this PR again once that's ready.

draffensperger:
> Currently this change goes to the v0.0.3 tag which seems lower than the highest tag of v0.1.0
> However, you can just do GO111MODULE=on go get github.com/census-instrumentation/opencensus-proto@master

When I do that, I still get the seemlying older (but actually newer) tag. Is this because the `master` branch of `opencensus-proto` is a different branch from the tagged release branch for `v0.1.x`?

> I don't think that we should make a new port to send to the agent. 

Good thought! I experimented some with using an HTTP handler that selects the appropriate handler based on the protocol, but I ran into errors with that and according to this [blog post](https://medium.com/@thatcher/why-choose-between-grpc-and-rest-bc0d351f2f84), it requires TLS setup to work that way. A suggested workaround is to use the [cmux](https://github.com/soheilhy/cmux) library, which splits the traffic by examining the first few bytes of the TCP connection, and that worked fine for me. I've updated it with the change to use `cmux`.

Would you be OK with this approach?

odeke-em:
>When I do that, I still get the seemlying older (but actually newer) tag. Is this because the master branch of opencensus-proto is a different branch from the tagged release branch for v0.1.x?

Perhaps you need to use the `-u` flag because there is a v0.1.0 tagged release https://github.com/census-instrumentation/opencensus-proto/releases/tag/v0.1.0
so please make it
```shell
GO111MODULE=on go get -u -v github.com/census-instrumentation/opencensus-proto@master
```

>but I ran into errors with that and according to this blog post, it requires TLS setup to work that way. 

Not really, the recommendation to use 
```go
if r.ProtoMajor == 2 && strings.HasPrefix(
	r.Header.Get("Content-Type"), "application/grpc") {
	grpcServer.ServeHTTP(w, r)
} else {
	yourMux.ServeHTTP(w, r)
}
```

surely will fail for cases where we can't muck around with the client making the request,
but the caveat here is that we have control over the frontend and just need to make sure that the frontend sends that distinguishing attribute, be it Content-Type or perhaps some sort of header. 

> suggested workaround is to use the cmux library, which splits the traffic by examining the first few bytes of the TCP connection, and that worked fine for me. I've updated it with the change to use cmux.

Sounds interesting and the pragmatic approach that we can use to get this functionality in. Please don't worry about my handler recommendation for now, I'll also experiment with it and later if see a need for code reduction, we'll apply it.

go.mod
```diff
@@ -12,14 +12,15 @@ require (
 	github.com/apache/thrift v0.0.0-20161221203622-b2a4d4ae21c7
 	github.com/aws/aws-sdk-go v1.15.68 // indirect
 	github.com/bmizerany/perks v0.0.0-20141205001514-d9a9656a3a4b // indirect
-	github.com/census-instrumentation/opencensus-proto v0.1.0
+	github.com/census-instrumentation/opencensus-proto v0.0.3-0.20181214143942-ba49f56771b8
```
odeke-em: Please don't forget to run the go get -u for the proto repository, as this proto version should be `v0.1.x`

draffensperger: It turns out that the `GO111MODULE=on go get -u -v github.com/census-instrumentation/opencensus-proto@master` still didn't make it `v0.1.0-...`, but I was able to manually update it to `v0.1.0-0.20181214143942-ba49f56771b8` and it worked.

receiver/opencensus/opencensus.go
```diff
@@ -28,22 +29,29 @@ import (
 	"github.com/census-instrumentation/opencensus-service/receiver"
 	"github.com/census-instrumentation/opencensus-service/receiver/opencensus/ocmetrics"
 	"github.com/census-instrumentation/opencensus-service/receiver/opencensus/octrace"
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
```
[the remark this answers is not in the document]
draffensperger: Done, good thought.

receiver/opencensus/opencensus.go
```diff
@@ -173,18 +184,69 @@ func (ocr *Receiver) Stop() error {
...
-		ocr.server.Stop()
+		ocr.serverGRPC.Stop()
+		_ = ocr.serverHTTP.Close()
 		_ = ocr.ln.Close()
 	})
 	return err
 }
 
-func (ocr *Receiver) startGRPCServer() error {
+type agentRouter struct {
```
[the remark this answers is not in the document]
draffensperger: Removed.

receiver/opencensus/opencensus.go
```diff
@@ -173,18 +184,69 @@ func (ocr *Receiver) Stop() error {
...
+	grpcHandler        *grpc.Server
+	grpcGatewayHandler http.Handler
+}
+
+func (agr *agentRouter) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	fmt.Printf("Content type: %v\n", r.Header.Get("Content-Type"))
```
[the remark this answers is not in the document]
draffensperger: Sure, removed this whole function since `agentRouter` is no longer needed.

receiver/opencensus/octrace/opencensus.go
```diff
@@ -102,6 +104,7 @@ func (oci *Receiver) Export(tes agenttracepb.TraceService_ExportServer) error {
 	spansMetricsFn := internal.NewReceivedSpansRecorderStreaming(tes.Context(), receiverName)
 
 	processReceivedSpans := func(ni *commonpb.Node, resource *resourcepb.Resource, spans []*tracepb.Span) {
+		fmt.Printf("Received spans: %v", spans)
```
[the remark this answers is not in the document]
draffensperger: Thanks, removed!

receiver/opencensus/opencensus_test.go
```diff
@@ -13,3 +13,133 @@
...
+       ]
+    }`)
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(traceJSON))
+	if err != nil {
+		t.Fatalf("Error creating trace POST request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	client := &http.Client{}
+	fmt.Println("Sending request...")
```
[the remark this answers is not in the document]
draffensperger: Removed

receiver/opencensus/opencensus_test.go
```diff
@@ -13,3 +13,133 @@
...
+	addr := ":35990"
+
+	// Set the buffer count to 1 to make it flush the test span immediately.
+	ocr, err := opencensus.New(addr, opencensus.WithTraceReceiverOptions(octrace.WithSpanBufferCount(1)))
+	if err != nil {
+		t.Fatalf("Failed to create trace receiver: %v", err)
+	}
+	defer ocr.StopTraceReception(context.Background())
+
+	fmt.Printf("Starting trace receiption")
```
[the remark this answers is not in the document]
draffensperger: Removed

receiver/opencensus/opencensus_test.go
```diff
@@ -13,3 +13,133 @@
...
+	sink := new(testhelper.ConcurrentSpanSink)
+
+	go func() {
+		if err := ocr.StartTraceReception(context.Background(), sink); err != nil {
+			t.Fatalf("Failed to start trace receiver: %v", err)
+		}
+	}()
+
+	fmt.Printf("Started trace receiption")
```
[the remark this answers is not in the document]
draffensperger: Done, removed. Sorry, forgot to remove these debugging log statements!

receiver/opencensus/octrace/opencensus.go
```diff
@@ -131,6 +134,11 @@ func (oci *Receiver) Export(tes agenttracepb.TraceService_ExportServer) error {
 
 		recv, err = tes.Recv()
 		if err != nil {
+			if err == io.EOF {
+				// Do not return EOF as an error so that grpc-gateway calls get an empty
```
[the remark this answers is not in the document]
draffensperger: Good point - I'll plan to add CORS support in a follow up PR.

internal/config/config.go
```diff
@@ -46,8 +46,9 @@ import (
 //      port: 55679
 
 const (
-	defaultOCReceiverAddress = ":55678"
-	defaultZPagesPort        = 55679
+	defaultOCReceiverAddress     = ":55678"
+	defaultZPagesPort            = 55679
+	defaultOCReceiverAddressHTTP = ":55681"
```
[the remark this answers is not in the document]
draffensperger: Done.

receiver/opencensus/opencensus.go
```diff
@@ -173,18 +184,51 @@ func (ocr *Receiver) Stop() error {
...
 	err := errAlreadyStarted
 	ocr.startServerOnce.Do(func() {
 		errChan := make(chan error, 1)
 		go func() {
-			errChan <- ocr.server.Serve(ocr.ln)
+			// Register the grpc-gateway on the HTTP server mux
+			c := context.Background()
+			opts := []grpc.DialOption{grpc.WithInsecure()}
+			endpoint := ocr.ln.Addr().String()
+			mux := ocr.httpServer().Handler.(*gatewayruntime.ServeMux)
```
yurishkuro: is the gateway's http server going to make a recursive local connection to gRPC, or is it going to invoke gRPC handler directly without an extra network hop?

receiver/opencensus/opencensus.go
```diff
@@ -173,18 +184,51 @@ func (ocr *Receiver) Stop() error {
 	var err = errAlreadyStopped
 	ocr.stopOnce.Do(func() {
 		// TODO: (@odeke-em) should we instead do (*grpc.Server).GracefulStop?
-		ocr.server.Stop()
+		ocr.serverGRPC.Stop()
+		_ = ocr.serverHTTP.Close()
 		_ = ocr.ln.Close()
 	})
 	return err
```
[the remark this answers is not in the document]
draffensperger: @odeke-em,  should this have some extra logic to return `nil` if it has not been stopped yet? Seems to have been added in https://github.com/census-instrumentation/opencensus-service/commit/164cdbf9ea2fac490b6c5e8c796994e317dfb507

receiver/opencensus/opencensus.go
draffensperger: It does make a recursive local connection to gRPC - not the most efficient, but to my knowledge the [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) project does not support that mode where you have it just call the gRPC service handler directly.

We could manually write a REST endpoint that converts the JSON to proto using [jsonpb](https://godoc.org/github.com/golang/protobuf/jsonpb) and then calls the service handler - or even one that just parses the JSON and executes the span receiver logic based on the JSON values. But either of those would require maintaining at least some extra code. My suggestion would be to hold off doing those optimizations for now, but I'm open if you think it should be done sooner.

go.mod
bcmills: Note that manual edits of this sort will no longer work in Go 1.13.
(And users may have to apply a `replace` directive to correct manually-edited pseudo-versions in the wild.)

Going forward, a more robust fix for this sort of problem is to add an explicit pre-release tag upstream (such as `v0.1.0-dev`) so that the `go` command can deduce the correct relationship between the release branch and the master branch.

PR #281, opened 2018-12-19
Better config naming for collector queue

pjanotti:
We still have to do further improvements on this area, this is a
incremental improvement trying to make things more clear.


cmd/occollector/app/builder/testdata/processor_config.yaml
```diff
@@ -1,18 +1,18 @@
-processors:
+ queued-exporters:
```
sjkaris: Should we rename the file to match the `processors` -> `queued-exporters` rename?

PR #282, opened 2018-12-20
Adds Batch Queue of Ids (for tail-sampling)

pjanotti:
Adding one of the types used to implement tail-based sampling.

This type will be used to create the batches that are periodically
processed while running tail-based sampling. The idea is that
receivers will enqueue ids concurrently in a batch while from time
to time a batch will be dequeued for sampling policy evaluation.

The queue of batches behaves in FIFO fashion and the owner of the type
is in charge of closing each batch by calling dequeue (which returns the
oldest batch in the queue while also starting a new batch).

### DIFF
--- a/internal/collector/processor/idbatchqueue/id_batch_queue.go
+++ b/internal/collector/processor/idbatchqueue/id_batch_queue.go
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package idbatchqueue defines a FIFO queue of fixed size in which the
-// elements are batches of ids defined as byte slices.
+// Package idbatchqueue defines a pipeline of fixed size in which the
+// elements are batches of ids.
 package idbatchqueue
 
 import (
@@ -34,22 +34,26 @@ type ID []byte
 // Batch is the type of batches held by the Batcher.
 type Batch []ID
 
-// Batcher behaves like a FIFO queue of a fixed number of batches, in which
-// items can be enqueued concurrently. The caller is in control of when a batch is
-// completed and a new one should be started.
+// Batcher behaves like a pipeline of batches that has a fixed number of batches in the pipe
+// and a new batch being built outside of the pipe. Items can be concurrently added to the batch
+// currently being built. When the batch being built is closed, the oldest batch in the pipe
+// is pushed out so the one just closed can be put on the end of the pipe (this is done as an
+// atomic operation). The caller is in control of when a batch is completed and a new one should
+// be started.
 type Batcher interface {
-	// Enqueue puts the given id on the batch being currently built. The client is in charge
+	// AddToCurrentBatch puts the given id on the batch being currently built. The client is in charge
 	// of limiting the growth of the current batch if appropriate for its scenario. It can
-	// either call Dequeue earlier or stop adding new items depending on what is required by
-	// the scenario.
-	Enqueue(id ID)
-	// Dequeue takes the batch at the front of the queue, and moves the current batch to
-	// the end of the queue, creating a new batch to receive new enqueued items.
-	// It returns the batch that was in front of the queue and a boolean
-	// that if true indicates that there are more batches to be retrieved.
-	Dequeue() (Batch, bool)
-	// Stop informs that no more items are going to be enqueued. After
-	// this method is called attempts to enqueue new items will panic.
+	// either call CloseCurrentAndTakeFirstBatch earlier or stop adding new items depending on what is
+	// required by the scenario.
+	AddToCurrentBatch(id ID)
+	// CloseCurrentAndTakeFirstBatch takes the batch at the front of the pipe, and moves the current
+	// batch to the end of the pipe, creating a new batch to receive new items. This operation should
+	// be atomic.
+	// It returns the batch that was in front of the pipe and a boolean that if true indicates that
+	// there are more batches to be retrieved.
+	CloseCurrentAndTakeFirstBatch() (Batch, bool)
+	// Stop informs that no more items are going to be batched and the pipeline can be read until it
+	// is empty. After this method is called attempts to enqueue new items will panic.
 	Stop()
 }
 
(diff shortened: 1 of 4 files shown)


internal/collector/processor/idbatchqueue/id_batch_queue.go
```diff
@@ -0,0 +1,149 @@
...
+	// of limiting the growth of the current batch if appropriate for its scenario. It can
+	// either call Dequeue earlier or stop adding new items depending on what is required by
+	// the scenario.
+	Enqueue(id ID)
+	// Dequeue takes the batch at the front of the queue, and moves the current batch to
+	// the end of the queue, creating a new batch to receive new enqueued items.
+	// It returns the batch that was in front of the queue and a boolean
+	// that if true indicates that there are more batches to be retrieved.
+	Dequeue() (Batch, bool)
+	// Stop informs that no more items are going to be enqueued. After
```
maynk: methods like start/stop are usually implementation details and do not belong to interface

internal/collector/processor/idbatchqueue/id_batch_queue.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// New creates a Batcher that will hold numBatches in its FIFO queue, having a channel with
+// batchChannelSize to receive new items. New batches will be created with capacity set to
+// newBatchesInitialCapacity.
+func New(numBatches, newBatchesInitialCapacity, batchChannelSize uint64) (Batcher, error) {
+	if numBatches < 1 {
+		return nil, ErrInvalidNumBatches
+	}
+	if batchChannelSize < 1 {
```
maynk: why not similar check for newBatchesInitialCapacity?

internal/collector/processor/idbatchqueue/id_batch_queue.go
```diff
@@ -0,0 +1,149 @@
...
+
+	batcher := &batcher{
+		pendingIds:                make(chan ID, batchChannelSize),
+		batches:                   batches,
+		currentBatch:              make(Batch, 0, newBatchesInitialCapacity),
+		newBatchesInitialCapacity: newBatchesInitialCapacity,
+		stopchan:                  make(chan bool),
+	}
+
+	// Single goroutine that keeps filling the current batch, contention is expected only
```
maynk: can't you do this lazily, when Dequeue is called, rather than a goroutine that runs all the time and contends for a mutex? that way the implementation can be lock-free

internal/collector/processor/idbatchqueue/id_batch_queue.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Enqueue puts the given id on the batch being currently built. The client is in charge
+// of limiting the growth of the current batch if appropriate for its scenario. It can
+// either call Dequeue earlier or stop adding new items depending on what is required by
+// the scenario.
+func (b *batcher) Enqueue(id ID) {
+	b.pendingIds <- id
+}
+
+// Dequeue takes the batch at the front of the queue, and moves the current batch to
```
maynk: the comment is misleading. It should be: Dequeue takes the "active" batch, closes it, and puts it at the back of the queue.
Active batch is logically not at front of queue.

internal/collector/processor/idbatchqueue/id_batch_queue_test.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+			if got != nil {
+				got.Stop()
+			}
+		})
+	}
+}
+
+func TestTypicalConfig(t *testing.T) {
+	concurrencyTest(t, 10, 100, uint64(4*runtime.NumCPU()))
```
maynk: why is the batchChannelSize expected to typically be 4*runtime.NumCPU()? isn't this too low? perhaps the thread dequeueing from channel doesn't get scheduled for 20-50ms and in the meanwhile a lot of IDs arrive then they would have to block.

pjanotti:
I liked the idea that you described, however, the performance was worst according to the benchmark:

pjanotti:
I liked your suggestion but the performance was worst for the high concurrent benchmark. My assumption is that channels will need some synchronization in the end too (although hidden from us):
```
BenchmarkOld-8        5000000           325 ns/op         139 B/op           0 allocs/op
BenchmarkNew-8        3000000           486 ns/op         136 B/op           0 allocs/op
```

Changed the comments to avoid confusion with queue, using the term pipeline instead.

internal/collector/processor/idbatchqueue/id_batch_queue_test.go
pjanotti: This number came from finding one with reasonable performance from the benchmark. Any goroutine trying to send to the channel will be preempted when the channel is full this should give an opportunity for the one receiving for the channel to run, and it should run until the channel is empty or a batch it switched (contention on the mutex). That said the risk here is if there are too many goroutines running but that will be a general problem not only for this go routine.

internal/collector/processor/idbatchqueue/id_batch_queue.go
pjanotti: Removed the queue terminology to avoid confusion, using "pipeline" and more descriptive names to the methods.

internal/collector/processor/idbatchqueue/id_batch_queue.go
pjanotti: That was a good idea, I gave it a go but in the end it requires synchronization anyway, i.e.: whatever is receiving the new elements to the current batch needs to be stopped and swapped when closing the current batch but new items still can be added concurrently.

internal/collector/processor/idbatchqueue/id_batch_queue.go
pjanotti: newBatchesInitialCapacity is to allow some guess about the typical size of batches in order to reduce re-allocations when the batches are growing and the code works even if it is zero. The other two can cause deadlock if their value is zero.

internal/collector/processor/idbatchqueue/id_batch_queue.go
pjanotti: I'm not sure this applies in this case since the pipe will keep a constant number of batches until it knows that no more batches will be added. After calling Stop eventually there should be no more data in the pipeline before that you should keep getting empty batches.

maynk: let's go with current interface, we can revisit when rest of the code is in place.

PR #283, opened 2018-12-21
Remove broken checksums now caught by go 1.11.4

pjanotti:
I was in contact with someone trying to build the repo and his golang was a new install (via brew). So he hit the same issue as Travis w/ go 1.11.4 (see https://github.com/census-instrumentation/opencensus-service/pull/272): invalid checksum. 

The problem for us is that the recommendation for this issue is to remove the old checksum, this fixes the build for go 1.11.4. However, any previous go version is going to re-introduce the incorrect checksums (I tested only with go 1.11.1). So if we merge this change and don't upgrade go to 1.11.4 on our boxes we are going to keep getting a modified go.sum (re-introducing the wrong checksums), which is fine as long as people are aware of the issue and don't commit the go.sum - in other words if this PR is merged any changes to go.mod/go.sum need to be done using go 1.11.4 and above.

See https://github.com/golang/go/issues/29278 for more info about the change in golang.

Creating as DO NOT MERGE so we can discuss and perhaps adopt some other alternative, however, notice that someone setting up a brand new go env is likely to hit the issue and be unable to build the repo. 

PS.: I remember in the past using brew to switch between go 1.10 and 1.11 but I am not sure if we can easily do the same for minor versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: go
 go_import_path: github.com/census-instrumentation/opencensus-service
 
 go:
-  - 1.11.2
+  - 1.11.x
  
 env:
   global:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,7 +56,7 @@ $ git push fork feature
 
 ## General Notes
 
-This project uses go 1.11 and Travis for CI.
+This project uses Go 1.11.4 and Travis for CI.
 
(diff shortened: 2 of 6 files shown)


songy23:
We ran into a similar issue when building with go1.11.4. See https://github.com/census-ecosystem/opencensus-experiments/pull/138.

/cc @rghetia

PR #284, opened 2018-12-27
Add basic telemetry to collector

pjanotti:
This change adds a Prometheus endpoint using OC to provide metrics
to the collector. There are settings to control the level and port
in which the metrics are exposed. The main metrics added for now
are in the queued exporters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,13 +319,15 @@ Usage:
   occollector [flags]
 
 Flags:
-      --config string      Path to the config file
-      --debug-processor    Flag to add a debug processor (combine with log level DEBUG to log incoming spans)
-  -h, --help               Help for occollector
-      --log-level string   Output level of logs (TRACE, DEBUG, INFO, WARN, ERROR, FATAL) (default "INFO")
-      --receive-jaeger     Flag to run the Jaeger receiver (i.e.: Jaeger Collector), default settings: {ThriftTChannelPort:14267 ThriftHTTPPort:14268}
-      --receive-oc-trace   Flag to run the OpenCensus trace receiver, default settings: {Port:55678}
-      --receive-zipkin     Flag to run the Zipkin receiver, default settings: {Port:9411}
+      --config string          Path to the config file
+      --debug-processor        Flag to add a debug processor (combine with log level DEBUG to log incoming spans)
+  -h, --help                   Help for occollector
+      --log-level string       Output level of logs (TRACE, DEBUG, INFO, WARN, ERROR, FATAL) (default "INFO")
+      --metrics-level string   Output level of telemetry metrics (NONE, BASIC, NORMAL, DETAILED) (default "BASIC")
+      --metrics-port uint16    Port exposing collector telemetry. (default 8888)
+      --receive-jaeger         Flag to run the Jaeger receiver (i.e.: Jaeger Collector), default settings: {ThriftTChannelPort:14267 ThriftHTTPPort:14268}
+      --receive-oc-trace       Flag to run the OpenCensus trace receiver, default settings: {Port:55678}
+      --receive-zipkin         Flag to run the Zipkin receiver, default settings: {Port:9411}
 ```
 
 Sample configuration file:
--- a/cmd/occollector/app/collector/collector.go
+++ b/cmd/occollector/app/collector/collector.go
@@ -181,10 +181,6 @@ func execute() {
 	receiversCloseFns := createReceivers(spanProcessor)
 	closeFns = append(closeFns, receiversCloseFns...)
 
-	view.Register(processor.MetricViews(telemetryLevel)...)
-	view.Register(processor.QueuedProcessorMetricViews(telemetryLevel)...)
-
-	// All metric views should have been loaded at this point.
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -319,13 +319,15 @@ Usage:
...
-      --log-level string   Output level of logs (TRACE, DEBUG, INFO, WARN, ERROR, FATAL) (default "INFO")
-      --receive-jaeger     Flag to run the Jaeger receiver (i.e.: Jaeger Collector), default settings: {ThriftTChannelPort:14267 ThriftHTTPPort:14268}
-      --receive-oc-trace   Flag to run the OpenCensus trace receiver, default settings: {Port:55678}
-      --receive-zipkin     Flag to run the Zipkin receiver, default settings: {Port:9411}
+      --config string          Path to the config file
+      --debug-processor        Flag to add a debug processor (combine with log level DEBUG to log incoming spans)
+  -h, --help                   Help for occollector
+      --log-level string       Output level of logs (TRACE, DEBUG, INFO, WARN, ERROR, FATAL) (default "INFO")
+      --metrics-level string   Output level of telemetry metrics (NONE, BASIC, NORMAL, DETAILED) (default "BASIC")
+      --metrics-port uint16    Port exposing collector telemetry. (default 8888)
```
maynk: do we need a --metrics-format (default to PROMETHEUS) or that is for future?

cmd/occollector/app/collector/collector.go
```diff
@@ -161,9 +181,19 @@ func execute() {
...
+	case err = <-asyncErrorChannel:
+		logger.Error("Asynchronous error received, terminating process", zap.Error(err))
+	case <-signalsChannel:
+		logger.Info("Received kill signal from OS")
+	}
 
-	<-signalsChannel
 	logger.Info("Starting shutdown...")
 
 	// TODO: orderly shutdown: first receivers, then flushing pipelines giving
```
maynk: at this point, shouldn't you also gracefully shutdown the http server that is serving metrics?

maynk: (shutdown would ideally include waiting the http server to exit before exiting the process)

internal/collector/processor/processor.go
```diff
@@ -37,11 +43,109 @@ func (sp *debugSpanProcessor) ProcessSpans(batch *agenttracepb.ExportTraceServic
...
 	return &debugSpanProcessor{logger: logger}
 }
+
+// Keys and stats for telemetry.
+var (
+	tagSourceFormatKey, _ = tag.NewKey("format")
+	tagServiceNameKey, _  = tag.NewKey("service")
+	tagExporterNameKey, _ = tag.NewKey("exporter")
+
+	statBatchCount        = stats.Int64("batches_received", "counts the number of batches received", stats.UnitDimensionless)
```
maynk: statReceivedBatchCount would be clearer

internal/collector/processor/processor.go
```diff
@@ -37,11 +43,109 @@ func (sp *debugSpanProcessor) ProcessSpans(batch *agenttracepb.ExportTraceServic
...
+	statDroppedBatchCount = stats.Int64("batches_dropped", "counts the number of batches dropped", stats.UnitDimensionless)
+
+	statSpanCount        = stats.Int64("spans_received", "counts the number of spans received", stats.UnitDimensionless)
```
maynk: statReceivedSpanCount would be clearer

internal/collector/processor/processor.go
```diff
@@ -37,11 +43,109 @@ func (sp *debugSpanProcessor) ProcessSpans(batch *agenttracepb.ExportTraceServic
...
+		Description: "The number of spans dropped.",
+		TagKeys:     tagKeys,
+		Aggregation: view.Sum(),
+	}
+
+	return []*view.View{receivedBatchesView, droppedBatchesView, receivedSpansView, droppedSpansView}
+}
+
+func statsTagsForBatch(processorName string, batch *agenttracepb.ExportTraceServiceRequest, spanFormat string) []tag.Mutator {
+	var serviceName string
```
maynk: really nitpicky comment, but I would suggest to break out the logic below into a separate utility function "serviceNameForBatch" and pass in a serviceName param to this method similar to processorName instead of the batch itself

README.md
pjanotti: Future

cmd/occollector/app/collector/collector.go
pjanotti: To gracefully shutdown here we will need to start having a timeout and so on, just calling Close is a rude shutdown (ie.: underlying listener is closed). This is desirable but we didn't do it yet for the pipes sending data to the backend. I think that adding it now for the handler of Prometheus metrics is an overkill (right now the worst that can happen is a GET to be cut in the middle and Prometheus losing one last scrape of data, but that was already in a race).

internal/collector/processor/processor.go
pjanotti: :+1:

internal/collector/processor/processor.go
pjanotti: Good idea, I like it :)

PR #285, opened 2018-12-28
Opencensus trace receiver should be enabled by default

### LINKED ISSUE #259: cmd/occollector: opencensus receiver should be enabled by default

maynk:
Previously all supported trace receivers (opencensus, zipkin, jaeger)
were disabled by default. This change enables the opencensus trace
receiver by default.

Fixes #259


maynk:
I signed it!

PR #293, opened 2019-01-05
cmd/ocagent: started Stackdriver Metrics exporter

odeke-em:
First fitting of a Metrics exporter with Stackdriver Metrics.
Depends on the completion of
    https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/69


example/main.go
```diff
@@ -31,14 +35,47 @@ func main() {
...
 	}
 	trace.RegisterExporter(oce)
+	view.RegisterExporter(oce)
+
 	trace.ApplyConfig(trace.Config{
 		DefaultSampler: trace.AlwaysSample(),
 	})
 
+	// Some stats
+	view.SetReportingPeriod(61 * time.Second)
```
pjanotti: It seems the same code as the other PR except for the period being 61 here, 4 on the Prometheus one. Why the different values in each case?

exporter/exporterparser/stackdriver.go
```diff
@@ -84,3 +88,12 @@ func (sde *stackdriverExporter) ExportSpans(ctx context.Context, td data.TraceDa
 	// upload can use the context and information from the Node.
 	return exportSpans(ctx, "stackdriver", sde.exporter, td)
 }
+
+var _ exporter.MetricsExporter = (*stackdriverExporter)(nil)
+
+func (sde *stackdriverExporter) ExportMetricsData(ctx context.Context, md data.MetricsData) error {
+	for _, metric := range md.Metrics {
+		_ = sde.exporter.ExportMetric(ctx, md.Node, md.Resource, metric)
```
pjanotti: I guess that the ExportMetric was defined to take only proto and stdlib types so no dependencies but it seems better to pass a slice of metrics instead of looping over them for every metrics exporter. Perhaps to be done separately.

exporter/metrics_exporter.go
```diff
@@ -29,16 +30,31 @@ type MetricsExporter interface {
...
+
 // ExportMetricsData exports the MetricsData to all exporters wrapped by the current one.
 func (mes metricsExporters) ExportMetricsData(ctx context.Context, md data.MetricsData) error {
 	for _, me := range mes {
 		_ = me.ExportMetricsData(ctx, md)
 	}
 	return nil
 }
+
+// ReceiveTraceData receives the span data in the protobuf format, translates it, and forwards the transformed
```
pjanotti: nit: our friend golint is not gone like that the comment doesn't match with the function name :)

example/main.go
odeke-em: @pjanotti, Stackdriver requires a minimum of 60 seconds when reporting metrics.

exporter/metrics_exporter.go
odeke-em: Haha, gotcha.

exporter/exporterparser/stackdriver.go
odeke-em: In deed, we can revisit that later after shipping this exporter out :)

odeke-em:
Thanks for the approval @songy23!
>Please fix the build.

There is some weird issue with go get with modules where it is downloading a version of the modules but checksums are failing
```shell
go: verifying github.com/apache/thrift@v0.0.0-20161221203622-b2a4d4ae21c7: checksum mismatch
	downloaded: h1:Fv9bK1Q+ly/ROk4aJsVMeuIwPel4bEnD8EPiI91nZMg=
	go.sum:     h1:CZI8h5fmYwCCvd2RMSsjLqHN6OqABlWJweFKxz4vdEs=
The command "go get -t -v ./..." failed and exited with 1 during .
```

songy23:
I believe that's because Go in different versions expect different checksum. On Travis CI it's using Go1.11.4, may on your local machine you're using Go1.11.2?

odeke-em:
>I believe that's because Go in different versions expect different checksum. On Travis CI it's using Go1.11.4, may on your local machine you're using Go1.11.2?

Interesting! I just manually copied and pasted the go.mod and go.sum files from the current master branch and now there is a bit of progress.

Issue #301, opened 2019-01-11
Readme exporter/receiver inaccuracies


sjkaris:
First, exporters:
It seems as though the agent will not export opencensus if the endpoint has a prefix as it does in the README, so 
```yaml
  opencensus:
    endpoint: "http://localhost:10001"
```
should be
```yaml
  opencensus:
    endpoint: "localhost:10001"
```

Receivers:
Zipkin appears to only take `port` as a configuration, not address, so 
```yaml
  zipkin:
    address: "localhost:9411"
```
should be
```yaml
  zipkin:
    port: 9411
```

pjanotti:
I started a pass on the docs and will finish this afternoon, I will double check these.

PR #302, opened 2019-01-11
Updated README for beta

flands:
While testing receivers and exporters issues with the documentation were
found. Addressing them for beta. We need to revisit for beta2 to address
the inconsistencies.

Addresses #301

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,12 +127,11 @@ receivers:
   opencensus:
     address: "localhost:55678"
 
-  zipkin:
-    port: 9411
-
   jaeger:
     collector_thrift_port: 14267
     collector_http_port: 14268
+
+  zipkin:
 ```
 
 ### <a name="config-exporters"></a>Exporters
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,7 +145,7 @@ please see the [exporter README.md](exporter/exporterparser/README.md).
 ```yaml
 exporters:
   opencensus:
-    endpoint: "http://localhost:10001"
+    endpoint: "localhost:10001"
```
pjanotti: :+1: valid syntax is defined here https://github.com/grpc/grpc/blob/master/doc/naming.md

README.md
```diff
@@ -128,7 +128,7 @@ receivers:
     address: "localhost:55678"
 
   zipkin:
-    address: "localhost:9411"
```
pjanotti: This depends: if the collector then port is correct, if agent address is correct... we have an issue open to fix that.

PR #308, opened 2019-01-12
exporter/prometheus: fix nil deference when no Prometheus exporter

odeke-em:
An embarassing crash would occur if in YAML, exporters were specified
and perhaps the Prometheus clause was started but had no content e.g.
```yaml
exporters:
   prometheus:
```

This change fixes that by explicitly returning nil early, but also
added a test to ensure we never regress.


odeke-em:
Thank you for the review @draffensperger!

PR #317, opened 2019-01-15
Add config based attributes

### LINKED ISSUE #296: cmd/occollector: add tags to all spans
Tags are a powerful ways of querying/aggregating tracing data. Requests have been made to be able to tag all spans received by a given collector (e.g. datacenter specific tagging). Other "global" configuration parameters may be desired in the future so this should be considered in the design. A proposal configuration may look something like the following:

```
global:
  tags:
  - key1: value1
  - key2: value2
```

An open question is how to handle keys defined in this way that already exist in the span. The proposal is:
* v1: override (collector wins)
* v2: provide a configuration option (either collector always wins or spans always wins for all tags)
* v3: provide a configuration option per key

Acceptance criteria:
* Ability to define one or more key/value pairs
* Defined key/value pairs will be attached to every span received by a given collector and forwarded to configured destination(s)
* If a span already has a defined key then it is overridden
* Stretch goal: configuration option to globally override or not override when a tag exists in a span

sjkaris:
fixes #296 by allowing for a configuration that will add attributes to all spans passing through the collector service.


cmd/occollector/app/collector/collector.go
```diff
@@ -192,7 +192,19 @@ func execute() {
 	}
 
 	// Wraps processors in a single one to be connected to all enabled receivers.
-	spanProcessor := processor.NewMultiSpanProcessor(spanProcessors...)
+	var processorOptions []processor.MultiProcessorOption
+	fmt.Printf("global %+v and global attrs %+v\n", multiProcessorCfg.Global, multiProcessorCfg.Global.Attributes)
```
pjanotti: Remove debug printf

pjanotti: Consider a logging statement

cmd/occollector/app/collector/collector.go
```diff
@@ -192,7 +192,19 @@ func execute() {
...
+	if multiProcessorCfg.Global != nil && multiProcessorCfg.Global.Attributes != nil {
+		fmt.Println("appending processor options")
```
pjanotti: same here

internal/collector/processor/multi_processor.go
```diff
@@ -18,28 +18,99 @@ import (
...
+			if len(attributes) == 0 {
+				return
+			}
+			for _, span := range batch.Spans {
+				if span == nil {
+					// We will not create nil spans with just attributes on them
+					continue
+				}
+				if span.Attributes == nil {
+					span.Attributes = &tracepb.Span_Attributes{}
```
pjanotti: nit: set capacity to len(attributes)

pjanotti:
gofmt -s -w ./internal/collector/processor/multi_processor_test.go 😄

README.md
```diff
@@ -162,6 +162,22 @@ exporters:
...
+
+The collector also takes some global configurations that modify its behavior for all receivers / exporters. One of the configurations
+available is to add Attributes or Tags to all spans passing through this collector. These additional tags can be configured to either overwrite
+attributes if they already exists on the span, or respect the original values. An example of this is provided below.
+```yaml
+global:
+  attributes:
+    overwrite: true
+    values:
+      some_string: "hello world"
```
pjanotti: Add a comment to indicate they are Key:Value pairs

PR #328, opened 2019-01-17
Add memory and CPU telemetry to prometheus stats

sjkaris:


internal/collector/telemetry/process_telemetry.go
```diff
@@ -0,0 +1,105 @@
...
+	done  chan struct{}
+}
+
+var mRuntimeAllocMem = stats.Int64("oc.io/process/memory_alloc", "Number of bytes currently allocated in use", "By")
+var viewAllocMem = &view.View{
+	Name:        mRuntimeAllocMem.Name(),
+	Description: mRuntimeAllocMem.Description(),
+	Measure:     mRuntimeAllocMem,
+	Aggregation: view.LastValue(),
+	TagKeys:     []tag.Key{},
```
pjanotti: nit: doesn't it work with the default value (nil) since it is an empty slice anyway? same below.

internal/collector/telemetry/process_telemetry.go
```diff
@@ -0,0 +1,105 @@
...
+	"go.opencensus.io/tag"
+)
+
+// ProcessMetricsViews is a struct that contains views related to process metrics (cpu, mem, etc)
+type ProcessMetricsViews struct {
+	views []*view.View
+	done  chan struct{}
+}
+
+var mRuntimeAllocMem = stats.Int64("oc.io/process/memory_alloc", "Number of bytes currently allocated in use", "By")
```
pjanotti: nit: "B" instead of "By"

sjkaris: According to the documentation, it should be By (for bytes) http://unitsofmeasure.org/ucum.html#para-48
https://opencensus.io/stats/measure/

internal/collector/telemetry/process_telemetry.go
sjkaris: Done.

internal/collector/telemetry/process_telemetry.go
```diff
@@ -0,0 +1,105 @@
...
+	Aggregation: view.LastValue(),
+	TagKeys:     []tag.Key{},
+}
+
+var mRuntimeTotalAllocMem = stats.Int64("oc.io/process/total_memory_alloc", "Number of bytes allocated in total", "By")
+var viewTotalAllocMem = &view.View{
+	Name:        mRuntimeTotalAllocMem.Name(),
+	Description: mRuntimeTotalAllocMem.Description(),
+	Measure:     mRuntimeTotalAllocMem,
+	Aggregation: view.Count(),
```
pjanotti: is this really a count? if it is a count, total number of allocation calls, the descrption for the measure is wrong.

internal/collector/telemetry/process_telemetry.go
```diff
@@ -0,0 +1,105 @@
...
+	TagKeys:     []tag.Key{},
+}
+
+var mRuntimeSysMem = stats.Int64("oc.io/process/sys_memory_alloc", "Number of bytes given to the process in total", "By")
```
pjanotti: memory of the system?

internal/collector/telemetry/process_telemetry.go
sjkaris: fixed.

internal/collector/telemetry/process_telemetry.go
sjkaris: Confusingly named by golang, I believe this is actually the amount of memory given to the process that it could allocate

internal/collector/telemetry/process_telemetry.go
```diff
@@ -0,0 +1,104 @@
...
+func (pmv *ProcessMetricsViews) StartCollection() {
+	ticker := time.NewTicker(5 * time.Second)
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				pmv.updateViews()
+			case <-pmv.done:
+				return
+			}
```
pjanotti: maybe is the formatting on GH but next bracket seems off

sjkaris: idk, gofmt seems to want it like this

PR #330, opened 2019-01-18
Valid OC proto traceID from 64bit ZipkinV2 traceID

### LINKED ISSUE #321: Zipkin traceID of 8 bytes generates invalid OC Proto traceID
Zipkin traceID can be 8 or 16 bytes, however, OC proto requires it to be 16-byte: https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace.proto#L38-L39

The conversion from Zipkin to OC proto can generate OC proto traceIDs with 8 bytes if the original Zipkin traceID was only 8 bytes. It should generate them with proper filling zeros.

Note this can impact Zipkin v1 and v2.

pjanotti:
Fixes #321


go.mod
```diff
@@ -49,7 +49,7 @@ require (
 	github.com/uber/jaeger-lib v1.5.0
 	github.com/uber/tchannel-go v1.10.0
 	github.com/yancl/opencensus-go-exporter-kafka v0.0.0-20181029030031-9c471c1bfbeb
-	go.opencensus.io v0.18.0
+	go.opencensus.io v0.18.1-0.20181204023538-aab39bd6a98b
```
sjkaris: is this intentional?

pjanotti: Ops, you are right I didn't do anything in this change to affect go.mod or go.sum ...

pjanotti:
Thanks for the review guys!

PR #337, opened 2019-01-23
 Refactor cmd/collector package

sjkaris:
The cmd codebase was getting cluttered, and this aims to simplify and
restructure how collector startup occurred.

Testing Done: no functionality changed, tested manually.


sjkaris:
Ah totally forgot about that haha, added :)

cmd/occollector/app/collector/telemetry.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2018, OpenCensus Authors
```
songy23: nit: should be 2019 instead of 2018.

PR #347, opened 2019-01-24
 Add compression to opencensus receiver/exporter

sjkaris:
Additionally enables the previously unintentionally disabled opencensus exporter in the collector.

### DIFF
--- a/internal/compression/grpc/grpc.go
+++ b/internal/compression/grpc/grpc.go
@@ -30,11 +30,11 @@ var (
 )
 
 // GetGRPCCompressionKey returns the grpc registered compression key if the
-// passed in compression key is supported, and returns nil otherwise.
-func GetGRPCCompressionKey(compression string) *string {
-	compressionKey := strings.ToLower(compression)
+// passed in compression key is supported, and Unsupported otherwise
+func GetGRPCCompressionKey(compressionType string) string {
+	compressionKey := strings.ToLower(compressionType)
 	if encodingKey, ok := grpcCompressionKeyMap[compressionKey]; ok {
-		return &encodingKey
+		return encodingKey
 	}
-	return nil
+	return compression.Unsupported
 }
--- a/internal/compression/grpc/grpc.go
+++ b/internal/compression/grpc/grpc.go
@@ -1,0 +1,40 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package grpc
+
+import (
+	"strings"
+
+	"google.golang.org/grpc/encoding/gzip"
+
+	"github.com/census-instrumentation/opencensus-service/internal/compression"
+)
+
+var (
+	// Map of opencensus compression types to grpc registered compression types
+	grpcCompressionKeyMap = map[string]string{
+		compression.Gzip: gzip.Name,
+	}
+)
+
+// GetGRPCCompressionKey returns the grpc registered compression key if the
+// passed in compression key is supported, and returns nil otherwise.
+func GetGRPCCompressionKey(compression string) *string {
+	compressionKey := strings.ToLower(compression)
+	if encodingKey, ok := grpcCompressionKeyMap[compressionKey]; ok {
+		return &encodingKey
+	}
+	return nil
+}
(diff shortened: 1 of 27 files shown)


internal/compression/types.go
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compression
+
+var (
+	supportedCompressionTypes = map[string]bool{
```
odeke-em: Perhaps we could provide more context as to what compression is for.

How about explicitly saying `supportedGRPCCompressionTypes`
so
```go
// Add more supported gRPC compression types here.
var supportedGRPCCompressionTypes = map[string]bool {
    "gzip": true,
}
```
and then

```go
func SupportedGRPCCompression(compressionType string) bool {
     supported, ok := supportedCompressionTypes[compressionType]
    return ok && supported
}
```

internal/compression/types_test.go
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package compression
+
+import (
+	"testing"
+)
+
+func TestIsSupportedCompressionType(t *testing.T) {
+	if !IsSupportedCompressionType("gzip") {
```
odeke-em: It would be nice to also pass in mis-capitalized "gzip" so "Gzip"

internal/config/config.go
```diff
@@ -299,7 +300,7 @@ func eqLocalHost(host string) bool {
 //  + kafka
 //  + opencensus
 //  + prometheus
-func ExportersFromYAMLConfig(config []byte) (traceExporters []exporter.TraceExporter, metricsExporters []exporter.MetricsExporter, doneFns []func() error, err error) {
+func ExportersFromYAMLConfig(logger *zap.Logger, config []byte) ([]exporter.TraceExporter, []exporter.MetricsExporter, []func() error, error) {
```
odeke-em: Am not too fond of passing in external loggers to libraries that are used by different packages, but anyways we are using zap allover so that's okay.

internal/compression/gzip.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compression
+
+import _ "google.golang.org/grpc/encoding/gzip"
```
odeke-em: You and I know that it is for gRPC compression but just the name of the package "compression" loses context and it could be compression for payloads. We use more than 1 serialization format, e.g. gRPC and Thrift, but also compression could be for other things. I think we should just place this code in exporterparser/opencensus.go instead of otherwise it is going to be very confusing.

internal/compression/types.go
odeke-em: Please see my comment about not making this a package due to ambiguity.

internal/compression/gzip.go
sjkaris: This is compression for both the collector and receiver, and therefore doesn't make sense in exporterparser. The initial plan was to have all compression be implemented here (be it for grpc, thrift, etc). But if we want to break out what is for grpc, thrift, http, etc, we could make separate packages within compression, where we would leave `types.go` where it is (to have functions for what is supported by various encodings) and then a compression/grpc package that contains the grpc specific compression, compression/thrift for thrift, etc.

sjkaris:
@odeke-em re-did the packaging, please take another look

internal/compression/grpc/grpc.go
```diff
@@ -0,0 +1,40 @@
...
+		compression.Gzip: gzip.Name,
+	}
+)
+
+// GetGRPCCompressionKey returns the grpc registered compression key if the
+// passed in compression key is supported, and returns nil otherwise.
+func GetGRPCCompressionKey(compression string) *string {
+	compressionKey := strings.ToLower(compression)
+	if encodingKey, ok := grpcCompressionKeyMap[compressionKey]; ok {
+		return &encodingKey
```
odeke-em: I am not sure that we need a string pointer. Perhaps just a string?

internal/compression/gzip.go
odeke-em: Roger that! Thanks for the explanation.

internal/compression/keys.go
```diff
@@ -0,0 +1,21 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compression
+
+// Compression keys for supported compression types within opencensus collector
+const (
+	Unsupported = "gzip"
```
odeke-em: Oh oh, was this a copy+paste mistake? Making Unsupported be "gzip"?

sjkaris: :o idk how i did that. Definitely a typo

sjkaris: its supposed to be empty string

sjkaris: good catch

odeke-em: Cool, good catch!

PR #350, opened 2019-01-25
Add zipkinv1 thrift 01

### LINKED ISSUE #348: zipkinV1: add support to "application/x-thrift"
Some applications sending zipkinv1, e.g.: traefik, are using content type as "application/x-thrift"

pjanotti:
Fixes #348 

Leverages the code already referenced from Jaeger to decode zipkin v1 thrift spans.

### DIFF
--- a/translator/trace/jaegerthrift_to_protospan.go
+++ b/translator/trace/jaegerthrift_to_protospan.go
@@ -106,10 +106,10 @@ func jSpansToOCProtoSpans(jspans []*jaeger.Span) []*tracepb.Span {
 		startTime := epochMicrosecondsAsTime(uint64(jspan.StartTime))
 		_, sKind, sStatus, sAttributes := jtagsToAttributes(jspan.Tags)
 		span := &tracepb.Span{
-			TraceId: jTraceIDToOCProtoTraceID(jspan.TraceIdHigh, jspan.TraceIdLow),
-			SpanId:  jSpanIDToOCProtoSpanID(jspan.SpanId),
+			TraceId: traceIDToOCProtoTraceID(jspan.TraceIdHigh, jspan.TraceIdLow),
+			SpanId:  spanIDToOCProtoSpanID(jspan.SpanId),
 			// TODO: Tracestate: Check RFC status and if is applicable,
-			ParentSpanId: jSpanIDToOCProtoSpanID(jspan.ParentSpanId),
+			ParentSpanId: spanIDToOCProtoSpanID(jspan.ParentSpanId),
 			Name:         strToTruncatableString(jspan.OperationName),
 			Kind:         sKind,
 			StartTime:    internal.TimeToTimestamp(startTime),
@@ -171,8 +171,8 @@ func jReferencesToOCProtoLinks(jrefs []*jaeger.SpanRef) *tracepb.Span_Links {
 		}
 
 		link := &tracepb.Span_Link{
-			TraceId: jTraceIDToOCProtoTraceID(jref.TraceIdHigh, jref.TraceIdLow),
-			SpanId:  jSpanIDToOCProtoSpanID(jref.SpanId),
+			TraceId: traceIDToOCProtoTraceID(jref.TraceIdHigh, jref.TraceIdLow),
+			SpanId:  spanIDToOCProtoSpanID(jref.SpanId),
 			Type:    linkType,
 		}
 		links = append(links, link)
@@ -181,7 +181,7 @@ func jReferencesToOCProtoLinks(jrefs []*jaeger.SpanRef) *tracepb.Span_Links {
 	return &tracepb.Span_Links{Link: links}
 }
 
-func jTraceIDToOCProtoTraceID(high, low int64) []byte {
+func traceIDToOCProtoTraceID(high, low int64) []byte {
 	if high == 0 && low == 0 {
 		return nil
 	}
@@ -191,7 +191,7 @@ func jTraceIDToOCProtoTraceID(high, low int64) []byte {
 	return traceID
 }
 
-func jSpanIDToOCProtoSpanID(id int64) []byte {
+func spanIDToOCProtoSpanID(id int64) []byte {
 	if id == 0 {
 		return nil
 	}
--- a/translator/trace/jaegerthrift_to_protospan.go
+++ b/translator/trace/jaegerthrift_to_protospan.go
@@ -106,10 +106,10 @@ func jSpansToOCProtoSpans(jspans []*jaeger.Span) []*tracepb.Span {
 		startTime := epochMicrosecondsAsTime(uint64(jspan.StartTime))
 		_, sKind, sStatus, sAttributes := jtagsToAttributes(jspan.Tags)
 		span := &tracepb.Span{
-			TraceId: traceIDToOCProtoTraceID(jspan.TraceIdHigh, jspan.TraceIdLow),
-			SpanId:  spanIDToOCProtoSpanID(jspan.SpanId),
+			TraceId: jTraceIDToOCProtoTraceID(jspan.TraceIdHigh, jspan.TraceIdLow),
+			SpanId:  jSpanIDToOCProtoSpanID(jspan.SpanId),
 			// TODO: Tracestate: Check RFC status and if is applicable,
-			ParentSpanId: spanIDToOCProtoSpanID(jspan.ParentSpanId),
+			ParentSpanId: jSpanIDToOCProtoSpanID(jspan.ParentSpanId),
 			Name:         strToTruncatableString(jspan.OperationName),
 			Kind:         sKind,
 			StartTime:    internal.TimeToTimestamp(startTime),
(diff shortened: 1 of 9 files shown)


translator/trace/zipkinv1_thrift_to_protospan.go
```diff
@@ -0,0 +1,211 @@
...
+		case zipkincore.AnnotationType_STRING:
+			pbAttrib.Value = &tracepb.AttributeValue_StringValue{
+				StringValue: &tracepb.TruncatableString{Value: string(binaryAnnotation.Value)}}
+		default:
+			err := fmt.Errorf("unknown zipkin v1 binary annotation type (%d)", int(binAnnotationType))
+			pbAttrib.Value = strAttributeForError(err)
+		}
+
+		key := binaryAnnotation.Key
+		if key == zipkincore.LOCAL_COMPONENT {
```
pjanotti: Opened https://github.com/census-instrumentation/opencensus-service/issues/352 so the same behavior regarding LOCAL_COMPONENT is also added to zipkin v1.

pjanotti:
Same appveyor failure: https://ci.appveyor.com/project/OpenCensusService/opencensus-service/builds/21895779

translator/trace/zipkinv1_thrift_to_protospan.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	attributes = &tracepb.Span_Attributes{
+		AttributeMap: attributeMap,
+	}
+	return attributes, localComponent
+}
+
+func bytesToNumber(b []byte, number interface{}) error {
+	buf := bytes.NewReader(b)
```
odeke-em: This seems quite expensive for just the flexibility of writing to various types. I think we can make this simple by
```go
func toUint64WithPadding(b []byte) (uint64, error) {
        if len(b) == 0 {
            return 0, errNoBytes
        }
	bs := b[:]
	if diff := 8 - len(bs); diff > 0 {
		// Left pad the byte slice to maxSizeOf(uint64AsBytes) i.e. 0x7F FF FF FF FF FF FF FF.
		bs = append(bytes.Repeat([]byte{0x00}, diff), bs...)
	}
	return binary.BigEndian.Uint64(bs), nil
}
```

and then casting to the various values values e.g.
```go
      if i, err := toUint64WithPadding(binaryAnnotation.Value); err == nil {
           pAttrib.Value = &tracepb.AttributeValue_IntValue{IntValue: int64(i)}
      } else {
           pbAttrib.Value = strAttributeForError(err)
```

translator/trace/zipkinv1_to_protospan.go
```diff
@@ -88,11 +90,24 @@ func ZipkinV1JSONBatchToOCProto(blob []byte) ([]*agenttracepb.ExportTraceService
...
+	return zipkinToOCProto(func(i int) (*tracepb.Span, *annotationParseResult, error) {
+		if i >= len(zSpans) {
+			return nil, nil, errNoMoreSpans
+		}
+
+		return zipkinV1ToOCSpan(zSpans[i])
+	})
+}
+
+func zipkinToOCProto(toOCSpan func(i int) (*tracepb.Span, *annotationParseResult, error)) ([]*agenttracepb.ExportTraceServiceRequest, error) {
```
odeke-em: Why is this function's idiom an iterator on an index instead of just iterating per span?
This new change makes for a non-pure-like function. I understand the need for specialized
processing with the toOCSpan but how about
```go
func zipkinToOCProto(zSpans []*zipkin.Span, toOCSpan func(zs *zipkincore.Span) (*tracepb.Span, *annotationParseResult, error)) ([]*agenttracepb.ExportTraceServiceRequest, error) {
       for _, zSpan := range zSpans {
             ocSpan, parsedAnnotations, err := zipkinV1ToOCSpan(zSpan)
             ...
      }
}
```

and that when used will just be simple matter of
```go
return zipkinToOCProto(zSpans, zipkinV1ToOCSpan)
```
instead of the repeated `if i >= len(zSpans)`

translator/trace/jaegerthrift_to_protospan.go
```diff
@@ -106,10 +106,10 @@ func jSpansToOCProtoSpans(jspans []*jaeger.Span) []*tracepb.Span {
 		startTime := epochMicrosecondsAsTime(uint64(jspan.StartTime))
 		_, sKind, sStatus, sAttributes := jtagsToAttributes(jspan.Tags)
 		span := &tracepb.Span{
-			TraceId: jTraceIDToOCProtoTraceID(jspan.TraceIdHigh, jspan.TraceIdLow),
-			SpanId:  jSpanIDToOCProtoSpanID(jspan.SpanId),
+			TraceId: traceIDToOCProtoTraceID(jspan.TraceIdHigh, jspan.TraceIdLow),
```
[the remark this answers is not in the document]
pjanotti: I found strange using j on z but they use the same convention, for me it is fine either way, it is more a matter of us being aware of the convention, will get it back to j prefix.

translator/trace/zipkinv1_thrift_to_protospan.go
```diff
@@ -0,0 +1,211 @@
...
+
+	agenttracepb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1"
+	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
+	"github.com/golang/protobuf/ptypes/timestamp"
+	"github.com/jaegertracing/jaeger/thrift-gen/zipkincore"
+)
+
+// ZipkinV1ThriftBatchToOCProto converts Zipkin v1 spans to OC Proto.
+func ZipkinV1ThriftBatchToOCProto(zSpans []*zipkincore.Span) ([]*agenttracepb.ExportTraceServiceRequest, error) {
+	return zipkinToOCProto(func(i int) (*tracepb.Span, *annotationParseResult, error) {
```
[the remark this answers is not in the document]
pjanotti: The loop in `zipkinToOCProto` runs looping over the triplet `(*tracepb.Span, *annotationParseResult, error)` but `ZipkinV1ThriftBatchToOCProto` and `ZipkinV1JSONBatchToOCProto` are slightly different: the former produces the pair for the loop from `zipkincore.Span` and the second from `zipkinV1Span`.  This is a lazy way of generating the triplet, the alternative to that is to store the results first on a slice and then loop over the results. Let me see how it looks like.

translator/trace/zipkinv1_to_protospan.go
pjanotti: In order to do that there is a conversion to zipkincore.Span that the JSON path doesn't need (per related comment above). The alternative seems to eager generate the slice with oc span and respective parsed annotation.

pjanotti: or a function that instead takes an interface, and do the cast inside...

pjanotti: in the end I wasn't satisfied with the two alternatives

odeke-em: >In order to do that there is a conversion to zipkincore.Span that the JSON path doesn't need (per related comment above). The alternative seems to eager generate the slice with oc span and respective parsed annotation.

if you have `toOCSpan` as specialized function that just takes in a `zipkincore.Span` that's the equivalent of taking an index and then looking it up and it even saves us from having to index on our own and then compare against the length of the slice then return `errNoMoreSpans`.
I still don't get the rationale here for the `for i := 0; ; i++` and then a mapping function that takes in an index yet we could simply iterate over each span.

pjanotti: `for i := 0; ; i++` and the `errNoMoreSpans` will be removed (something that ended in between two ideas). 

> if you have toOCSpan as specialized function that just takes in a zipkincore.Span 

The usage of `toOCSpan` with an index is to avoid converting from `zipkinV1Span` to `zipkincore.Span` (or vice-versa). The simplest alternative to get rid of it is to do the all transformations (to the triples) before going to the "re-used" loop, i.e.: `zipkinToOCProto` it will allocate an extra array per call and will have very similar loops for v1 json and thrift functions but likely very minimal impact on performance overall (the function itself should be much more expensive than the extra array).

pjanotti: confirmed: 1 extra allocation and ~1% longer per benchmark, but the overall time measured is around 10ms. It is acceptable.

odeke-em: >The usage of toOCSpan with an index is to avoid converting from zipkinV1Span to zipkincore.Span (or vice-versa). The simplest alternative to get rid of it is to do the all transformations (to the triplets) before going to the "re-used" loop, i.e.: zipkinToOCProto. It will allocate an extra array per call and will have very similar loops for v1 json and thrift functions but likely very minimal impact on performance overall (the function itself should be much more expensive than the extra array).

Currently I think let's go for clarity before optimization and then can optimize as needed. Thank you for making the changes for clearer code :)

odeke-em:
>Fix #348

@pjanotti if you use `Fixes #348` instead of `Fix #348` in your commit message, Github will auto-close the corresponding issue once this PR is merged.

odeke-em:
@pjanotti please try force pushing to retrigger CI as https://github.com/census-instrumentation/opencensus-service/issues/356 seems like a flake.

PR #359, opened 2019-02-04
Add a "zipkin-scribe" receiver

### LINKED ISSUE #338: receiver/zipkin: Add scribe protocol support
The Zipkin receiver accepts HTTP today, it should also accept scribe.

pjanotti:
Fix #338 

Adds support to zipkin-scribe. While this is zipkin it uses a different protocol (other are all over http), so this becomes a different receiver "zipkin-scribe" that can be enabled separately from normal zipkin. It adds a reference to a repo with the thrift generated files for Scribe protocol (the project in question uses the same version of thrift currently used in oc-service, an old one to match usage in some Jaeger references), the repo is https://github.com/Omnition/scribe-go we can move it later if we want (we can't simply add the files to this repo because them we need to subject the generated files to the same gofmt rules, etc).

### DIFF
--- a/receiver/zipkin/scribe/scribe_receiver.go
+++ b/receiver/zipkin/scribe/scribe_receiver.go
@@ -1,0 +1,170 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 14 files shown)


receiver/zipkin/scribe/scribe_receiver.go
```diff
@@ -0,0 +1,170 @@
...
+		zSpans = append(zSpans, zs)
+	}
+
+	if len(zSpans) == 0 {
+		return scribe.ResultCode_OK, nil
+	}
+
+	ocBatches, err := tracetranslator.ZipkinV1ThriftBatchToOCProto(zSpans)
+	if err != nil {
+		return scribe.ResultCode_OK, err
```
odeke-em: This looks out of place, `scribe.ResultCode_OK` yet we encountered an error?

pjanotti: I had the same feeling, the thing is that the interface only defines ResultCode_OK and ResultCode_TRY_LATER, so in this case the ResultCode_OK paired with the error seems the appropriate response.

pjanotti:
Thanks for the review @odeke-em!

PR #361, opened 2019-02-05
Add batching to queued sender implementations

sjkaris:
Adds batching (as a feature of queuing) to the current queued sender implementations.


sjkaris:
@songy23 Makes sense, split out the exporter changes (this will just add to the current queued senders), and will make a followup PR containing the changes to allow the other exporters to be batched/queued. Ultimately with the discussion on friday, I can see if we just want to forgo any of these smaller config changes, and just refactor the entire configuration scheme in one go

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+}
+
+func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource, spanFormat string) string {
+	h := md5.New()
+	if node != nil {
+		nodeKey, err := proto.Marshal(node)
+		if err != nil {
+			b.logger.Error("Error marshalling node to batcher mapkey.", zap.Error(err))
+		}
+		h.Write(nodeKey)
```
pjanotti: Do you want to call h.Write if err != nil?

pjanotti: same below

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+// batches
+func (b *batcher) ProcessSpans(request *agenttracepb.ExportTraceServiceRequest, spanFormat string) (uint64, error) {
+	bucketID := b.genBucketID(request.Node, request.Resource, spanFormat)
+	bucket := b.getOrAddBucket(bucketID, request.Node, request.Resource, spanFormat)
+	bucket.add(request.Spans)
+	return 0, nil
+}
+
+func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource, spanFormat string) string {
+	h := md5.New()
```
pjanotti: I'm curious about the numbers that you get in `BenchmarkGenBucketID`

cmd/occollector/app/builder/processor_builder.go
```diff
@@ -78,6 +106,9 @@ type QueuedSpanProcessorCfg struct {
 	// SenderType indicates the type of sender to instantiate
 	SenderType   SenderType `mapstructure:"sender-type"`
 	SenderConfig interface{}
+	// BatchingConfig sets config parameters related to batching
+	BatchingConfig BatchingConfig `mapstructure:"batching"`
+	RawConfig      *viper.Viper
```
pjanotti: Why this needs to be added?

go.mod
```diff
@@ -71,6 +71,7 @@ require (
 	go.uber.org/zap v1.9.1
 	golang.org/x/crypto v0.0.0-20190131182504-b8fe1690c613 // indirect
 	golang.org/x/oauth2 v0.0.0-20181102170140-232e45548389 // indirect
+	golang.org/x/sys v0.0.0-20181031143558-9b800f95dbbc // indirect
```
pjanotti: I don't think you need go.* changes...

cmd/occollector/app/builder/processor_builder.go
```diff
@@ -63,6 +67,30 @@ func NewJaegerThriftHTTPSenderCfg() *JaegerThriftHTTPSenderCfg {
 	return opts
 }
 
+// BatchingConfig contains configuration around the queueing batching.
+// It contains some advanced configurations, which should not be used
+// by a typical user, but are provided as advanced features to increase
+// scalability.
+type BatchingConfig struct {
+	// Enabled marks batching as enabled or not
+	Enabled bool `mapstructure:"enabled"`
```
odeke-em: Perhaps `Enable` as `Enabled` implies a past tense and the equivalent of having already been applied.

cmd/occollector/app/collector/processors.go
```diff
@@ -93,21 +94,53 @@ func buildQueuedSpanProcessor(logger *zap.Logger, opts *builder.QueuedSpanProces
 			logger,
 			sender.HTTPTimeout(thriftHTTPSenderOpts.Timeout),
 		)
-	default:
-		logger.Fatal("Unrecognized sender type configured")
+	}
+
+	if spanSender == nil {
+		logger.Fatal("Unrecognized sender type or no exporters configured")
```
odeke-em: It would be useful for this Fatal to be a Fatalf and include the opts.SenderType.

cmd/occollector/app/collector/processors.go
```diff
@@ -63,7 +64,7 @@ func createExporters(v *viper.Viper, logger *zap.Logger) ([]func(), []exporter.T
 	return wrappedDoneFns, traceExporters, metricsExporters
 }
 
-func buildQueuedSpanProcessor(logger *zap.Logger, opts *builder.QueuedSpanProcessorCfg) (processor.SpanProcessor, error) {
+func buildQueuedSpanProcessor(logger *zap.Logger, opts *builder.QueuedSpanProcessorCfg) ([]func(), processor.SpanProcessor, error) {
```
odeke-em: It would be useful if you could name this return parameters. With the addition `[]func()` I can't tell what this is and below in the good case you return `return nil, queuedSpanProcessor, nil` which is not intuitive.

internal/collector/processor/nodebatcher/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+
+	"github.com/census-instrumentation/opencensus-service/internal/collector/processor"
+	"github.com/census-instrumentation/opencensus-service/internal/collector/telemetry"
+)
+
+var (
+	statBatchSize         = stats.Int64("batch_size", "Size of batches sent from the batcher (in span)", stats.UnitDimensionless)
+	statAddNodeBatches    = stats.Int64("add_node_batches", "Count of nodes that have been added to batching.", stats.UnitDimensionless)
```
odeke-em: `statNodesAddedToBatches` -- `Number of nodes that have been batched`

internal/collector/processor/nodebatcher/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+	statRemoveNodeBatches = stats.Int64("remove_node_batches", "Number of nodes that have been removed from batching.", stats.UnitDimensionless)
```
odeke-em: `statsNodesRemovedFromBatches`

internal/collector/processor/nodebatcher/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+
+	statSendByBatchSize     = stats.Int64("batch_size_trigger_send", "Number of times the batch was sent due to a size trigger", stats.UnitDimensionless)
```
odeke-em: `statCountOfBatchingBySizeTrigger` -- `Number of times that batches were sent because of a size trigger`

internal/collector/processor/nodebatcher/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+	statSendByTimeout       = stats.Int64("timeout_trigger_send", "Number of times the batch was sent due to a timeout trigger", stats.UnitDimensionless)
```
odeke-em: `statCountOfBatchingByTimeout`

internal/collector/processor/nodebatcher/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+	statAddOnDeadNodeBucket = stats.Int64("removed_node_send", "Number of times the batch was sent due to spans being added for a no longer active node", stats.UnitDimensionless)
```
odeke-em: `statCountOfBatchingByDeadNode`

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
+	"github.com/golang/protobuf/proto"
+	"go.opencensus.io/stats"
+	"go.uber.org/zap"
+
+	"github.com/census-instrumentation/opencensus-service/internal/collector/processor"
+)
+
+const (
+	initBatchCap             = uint32(1024)
```
odeke-em: `initialBatchCapacity`

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+	nodeDead                 = uint32(1)
```
odeke-em: Was it intended that `nodeDead` and `batchClosed` have the same values? Those seem to me to be mutually exclusive states.

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris:
```
BenchmarkGenBucketID/composite-md5-bigInput-12         	  500000	      3229 ns/op
BenchmarkGenBucketID/composite-md5-smallInput-12       	 3000000	       517 ns/op
```

cmd/occollector/app/builder/processor_builder.go
sjkaris: ah, it was needed for the followup which changes the other exporters to allow for batching. I must have missed its removal -- removed.

internal/collector/processor/nodebatcher/node_batcher.go
odeke-em: @pjanotti hash.Hash.Write will never return an error https://golang.org/pkg/hash/#Hash

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+// ProcessSpans implements batcher as a SpanProcessor and takes the provided spans and adds them to
+// batches
+func (b *batcher) ProcessSpans(request *agenttracepb.ExportTraceServiceRequest, spanFormat string) (uint64, error) {
+	bucketID := b.genBucketID(request.Node, request.Resource, spanFormat)
+	bucket := b.getOrAddBucket(bucketID, request.Node, request.Resource, spanFormat)
+	bucket.add(request.Spans)
+	return 0, nil
+}
+
+func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource, spanFormat string) string {
```
odeke-em: This method seems quite expensive with all the proto Marshalling. Perhaps it would be useful to compare proto.Marshal-ing vs just stringifying

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+	if ok {
+		return bucket.(*nodeBatcher)
+	}
+	return nil
+}
+
+func (b *batcher) getOrAddBucket(
+	bucketID string, node *commonpb.Node, resource *resourcepb.Resource, spanFormat string,
+) *nodeBatcher {
+	bucket, loaded := b.buckets.LoadOrStore(
```
odeke-em: s/loaded/alreadyStored/g or `alreadyAdded`. Loaded isn't intuitive and speaks to people who are used to atomics.

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+	stats.Record(context.Background(), statRemoveNodeBatches.M(1))
+	b.buckets.Delete(bucketID)
+}
+
+type nodeBatcher struct {
+	timeout       time.Duration
+	sendBatchSize uint32
+	currBatch     *batch
+	node          *commonpb.Node
+	resource      *resourcepb.Resource // TODO(skaris) remove when resource added to span
```
odeke-em: `when resource is added`

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+		parent:          parent,
+	}
+
+	return nb
+}
+
+func (nb *nodeBatcher) add(spans []*tracepb.Span) {
+	// Reset cycles untouched to 0
+	atomic.StoreUint32(&nb.cyclesUntouched, 0)
+	// Add will always add these spans to the current batch, if ok is false it means that
```
odeke-em: What does `cut` mean?

go.mod
sjkaris: removed

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,467 @@
...
+	// we need to cut
+	var b *batch
+	closed := true
+	cut := false
+	for closed {
+		// atomic.LoadPointer only takes unsafe.Pointer interfaces. We do not use unsafe
+		// to skirt around the golang type system.
+		b = (*batch)(atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&nb.currBatch))))
+		cut, closed = b.add(spans)
```
odeke-em: So am assuming `cut` here means `performedBatching`? If so, please use a clearer term like `performedBatching`.

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: Nope, fixed.

internal/collector/processor/nodebatcher/node_batcher.go
pjanotti: The result for smallinput is good (comparable to other small functions that we have), the biginput doesn't seem worth any investment at this point since the code is going to change soon when we fully switch to resource instead of node.

internal/collector/processor/processor.go
```diff
@@ -129,25 +130,32 @@ func MetricViews(level telemetry.Level) []*view.View {
...
-		serviceName = batch.Node.ServiceInfo.Name
+		serviceName = node.ServiceInfo.Name
 	}
 	return serviceName
 }
 
-func statsTagsForBatch(processorName, serviceName, spanFormat string) []tag.Mutator {
+// ServiceNameForBatch gets the service name for a specified batch. Used for metrics.
+func ServiceNameForBatch(batch *agenttracepb.ExportTraceServiceRequest) string {
+	return ServiceNameForNode(batch.Node)
```
odeke-em: Please be defensive about this
```go
if batch == nil {
    return ""
}
return ServiceNameForNode(batch.Node)
```

internal/collector/processor/nodebatcher/node_batcher.go
odeke-em: @pjanotti how are you getting those comparisons of worthiness? Seems like we don't yet have a reference benchmark to gauge against. I made a suggestion to benchmark perhaps stringifying vs md5 hashes.

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: Did the comparison, proto marshalling was actually about 2x-3x faster. 

```
BenchmarkGenBucketID/string-smallInput-12         	 1000000	      1730 ns/op
BenchmarkGenBucketID/string-bigInput-12           	  200000	      6902 ns/op
BenchmarkGenBucketID/md5-smallInput-12            	 1000000	      1036 ns/op
BenchmarkGenBucketID/md5-bigInput-12              	  300000	      5626 ns/op
BenchmarkGenBucketID/composite-md5-smallInput-12       	 3000000	       517 ns/op
BenchmarkGenBucketID/composite-md5-bigInput-12         	  500000	      3229 ns/op
```

sjkaris: (currently the composite-md5 is the function that is still implemented)

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: used `cutBatch` instead (mirroring the function name) lmk what you think

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: I think both are important -- the proto marshalling actually does equally well with the big and small -- its perf difference is the same multiple (2-3x) for both big and small vs stringify

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: They are states for different things -- nodeDead is a nodeBucket state, batchClosed is a batch state. Perhaps rename to `nodeBucket_nodeDead` and `batch_closed` ?

internal/collector/processor/nodebatcher/node_batcher.go
pjanotti: @odeke-em yeah, we should add some more benchmarks and publish them as reference. I currently get ~63K for unmarshaling a zipkinv1 JSON batch , and around ~10Kns for a similar batch in zipkinv1 thrift. For comparison that [tiny optimization that was merged recently](https://github.com/census-instrumentation/opencensus-service/pull/360) had the function with worst case before optimization at around 360ns.

internal/collector/processor/nodebatcher/node_batcher.go
odeke-em: Having `_<suffix>` is iffy, perhaps:
```go
nodeStatusDead = uint32(1)
batchStatusClode = uint32(1)
```

sjkaris: done.

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -0,0 +1,469 @@
...
+
+type batch struct {
+	items         atomic.Value // []*tracepb.Span
+	currCap       uint32
+	nextEmptyItem uint32
+	sendItemsSize uint32
+
+	closed  uint32
+	growMu  sync.Mutex
+	pending int32
```
odeke-em: Using atomics might fail due to the arrangement of these fields due to potential alignment issue

Field|64 bit size(bytes)|32 bit size(bytes)
---|---|---
items|16|8
currCap|4|4
nextEmptyItem|4|4
sendItemsSize|4|4
close|4|4
growMu|8|8
pending|4|4

Please examine and adjust accordingly, you might want to ensure that all the mutex is first, then integers and then items.

odeke-em: or rather, integers first, then items and then lastly the mutex but just make sure that anything that gets loaded or stored is either on a 4 or 8 alignment boundary.

sjkaris: Discussed offline -- using 32 bit ints is safe for both 32 and 64 bit systems since golang keeps stuff word-aligned. Moved `lastSent` to top for alignment since it is 64 bits and therefore needs to be 64-bit aligned for 32 bit systems.

sjkaris:
travis sometimes hangs on testing exporters, created #375

cmd/occollector/app/builder/processor_builder.go
```diff
@@ -63,6 +67,30 @@ func NewJaegerThriftHTTPSenderCfg() *JaegerThriftHTTPSenderCfg {
...
+
+	// NumTickers sets the number of tickers to use to divide the work of looping
+	// over batch buckets. This is an advanced configuration option.
+	NumTickers int `mapstructure:"num-tickers,omitempty"`
+	// TickTime sets time interval at which the tickers tick. This is an advanced
+	// configuration option.
+	TickTime *time.Duration `mapstructure:"tick-time,omitempty"`
+	// RemoveAfterTicks is the number of ticks that must pass without a span arriving
+	// from a node after which the batcher for that node will be deleted. This is an
+	// advanved configuration option.
```
flands: advanced

PR #362, opened 2019-02-06
receiver: implemented Prometheus receiver

### LINKED ISSUE #137: receiver/metrics: add Prometheus metrics receiver.

odeke-em:
This Prometheus receiver requires adding yaml
definitions to the configuration file under "receivers"

```yaml
receivers:
   prometheus:
     - job_name: 'ocjdbc'
        scrape_interval: 5s
        static_configs:
          - targets: ['localhost:9988']

     - job_name: 'kafka_metrics'
         scrape_interval: 4s
         static_configs:
           - targets: ['localhost:8887']

     buffer_period: 500ms
     buffer_count: 2
```

Fixes #137


receiver/prometheus/metrics_receiver.go
```diff
@@ -15,36 +15,90 @@
...
 import (
+	"context"
 	"errors"
+	"sync"
 	"time"
 
+	agentmetricspb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/metrics/v1"
+	"github.com/census-instrumentation/opencensus-service/data"
 	"github.com/census-instrumentation/opencensus-service/receiver"
+	"github.com/orijtech/promreceiver"
```
songy23: Do we have a plan to move the receiver to this repo?

receiver/prometheus/metrics_receiver.go
```diff
@@ -15,36 +15,90 @@
...
+		tms := &promMetricsReceiverToOpenCensusMetricsReceiver{mrs: mrs}
+		cfg := pr.cfg
+		pr.recv, err = promreceiver.ReceiverFromConfig(
+			context.Background(),
+			tms,
+			cfg.ScrapeConfig,
+			promreceiver.WithBufferPeriod(cfg.BufferPeriod),
+			promreceiver.WithBufferCount(cfg.BufferCount))
+	})
+	return err
```
sjkaris: I believe this will always return this error (even when StartMetricsReception was called for the first time)

receiver/prometheus/metrics_receiver.go
```diff
@@ -15,36 +15,90 @@
...
+}
+
+func (pr *preceiver) StopMetricsReception(ctx context.Context) error {
+	pr.recv.Flush()
+	<-time.After(2 * time.Second)
```
sjkaris: I think we shouldn't block the caller. Could start a small goroutine to cancel after the timeout instead.

receiver/prometheus/metrics_receiver.go
odeke-em: I wouldn't advise doing that right now, but it can be done. The Prometheus receiver code is about ~1K lines long because we have to translate them on notification from Prometheus listeners, from their labels to our metrics.

That repo is licensed for OpenCensus Authors and preferably, I'd rather put it under census-ecosystem so that it can be worked on without blocking work on this repo.

internal/config/config.go
```diff
@@ -99,6 +102,16 @@ type ReceiverConfig struct {
 	DisableTracing bool `yaml:"disable_tracing"`
 	// DisableMetrics disables metrics receiving and is only applicable to metrics receivers.
 	DisableMetrics bool `yaml:"disable_metrics"`
+
+	// Configuration is an inlined attribute that defines Prometheus configurations.
```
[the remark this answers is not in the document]
odeke-em: Roger that, thanks! I'll update it.

receiver/prometheus/metrics_receiver.go
odeke-em: I'll take it out sure, that time.After isn't to cancel though, that's do give it at least 2 seconds to shut down.

receiver/prometheus/metrics_receiver.go
odeke-em: No it won't. startOnce is only ever invoked once and when invoked on line 69, we set it
screen shot 2019-02-06 at 6 51 50 pm

odeke-em:
For some reason integrating go-modules with that minimal code is causing trouble and I spent all day trying to work on that, so perhaps I have to go through all our dependencies to see what's tripping out.

songy23:
Which Go version are you using? `go.sum` generated by Go1.11.2 doesn't work with Go1.11.4+, and Travis and Appveyor are using Go1.11.5.

receiver/prometheus/metrics_receiver.go
sjkaris: Ah idk how i missed that.... this LGTM!

odeke-em:
This PR is heavily stalled by problems with go mod and not matching dependencies despite us specifying each dependency. I have spent days just fiddling with it that I am inclined to stop wasting time and vendor these dependencies by hand -- more time spent fighting the system than writing code and tests, unfortunately. It is ready for review though.

receiver/prometheus/metrics_receiver.go
odeke-em: @sjkaris I've removed the `<-time.After`, PTAL.

odeke-em:
I took a stab at solving the dependency problems this and pleased that all is now well and we now have the Prometheus receiver at long last!

PR #368, opened 2019-02-07
receiver/opencensus: multiplex application/grpc+proto for gRPC based receiver

### LINKED ISSUE #365: Receiving HTTP 1.1 400 response from agent
I've been trying to get a client for the agent working for the Erlang opencensus library. Oddly the client is able to publish to the collector if configured to point to it but when connecting to the agent and attempting to start the export stream I'm getting an HTTP 1.1 response:

```
400 Bad Request\r\nContent-Type: text/plain; charset=utf-8\r\nX-Content-Type-Options: 
nosniff\r\nDate: Thu, 07 Feb 2019 02:11:51 GMT\r\nContent-Length: 12\r\nConnection: 
close\r\n\r\nBad Request\n
```

Any idea how the collector and agent might be running differently?

### LINKED ISSUE #366: receiver/opencensus: grpc-gateway mux should accept all gRPC content-types
Found the root cause of @tsloughter's issue reported in #365 by recalling that the receiver multiplexes on content from the web-grpc-gateway. His Erlang client was sending content-type of "application/grpc+proto" but instead matching to the web-gateway route.

This issue serves to us accepting all possible gRPC content-types as per https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
screen shot 2019-02-06 at 8 30 15 pm

or at least support "application/grpc+proto" since that's what we primarily support currently. We can support "application/grpc+json" later on as a nice to have for maximum compatibility.

odeke-em:
The registered grpc-gateway HTTP/2 content-type header field
was "application/grpc" but some clients in the wild send
"application/grpc+proto" and as per the gRPC spec, it is a descriptive
content length that should be handled by Proto (un)marshalers.

This change removed the use of

    (*grpc.Server).Stop()

which painfully took 20+s even without the "-race" go start option.
Sure, that method adds some sort of internal shutdown mechanism but
it is very painful to wait for 20+s for a server with zero connections
to stop yet all we need to do is stop the actual net.Listener by

     (net.Listener).Stop()

This change also adds a test which should work in a perfect world.
However, we don't have control over writes on the wire with the gRPC
connection hence our expectations for results to assert against are
blissfully optimistic and will fail most of the time as those writes
take sometime. The test has been included but skipped until we can
figure out what's going on. In the meantime though, users should be
able to use the OpenCensus receiver without trouble. Also internally
in the receivers, we've got data bundling which is controlled
separately despite options.

Fixes #365
Fixes #366

### DIFF
--- a/receiver/opencensus/opencensus_test.go
+++ b/receiver/opencensus/opencensus_test.go
@@ -24,14 +24,17 @@ import (
 	"testing"
 	"time"
 
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+
 	"github.com/census-instrumentation/opencensus-service/receiver/opencensus"
+	"github.com/census-instrumentation/opencensus-service/receiver/opencensus/octrace"
 	"github.com/census-instrumentation/opencensus-service/receiver/testhelper"
 
 	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
 	agenttracepb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1"
 	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
 	"github.com/census-instrumentation/opencensus-service/internal"
-	"github.com/census-instrumentation/opencensus-service/receiver/opencensus/octrace"
 )
 
 func TestGrpcGateway_endToEnd(t *testing.T) {
@@ -153,7 +156,7 @@ func TestGrpcGatewayCors_endToEnd(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to create trace receiver: %v", err)
 	}
-	defer ocr.StopTraceReception(context.Background())
+	defer ocr.Stop()
 
 	sink := new(testhelper.ConcurrentSpanSink)
 	go func() {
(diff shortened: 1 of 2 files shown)


receiver/opencensus/opencensus_test.go
```diff
@@ -174,6 +177,101 @@ func TestGrpcGatewayCors_endToEnd(t *testing.T) {
 	verifyCorsResp(t, url, "disallowed-origin.com", 200, false)
 }
 
+// As per Issue https://github.com/census-instrumentation/opencensus-service/issues/366
+// the agent's mux should be able to accept all Proto affiliated content-types and not
+// redirect them to the web-grpc-gateway endpoint.
+func TestAcceptAllGRPCProtoAffiliatedContentTypes(t *testing.T) {
+	t.Skip("Currently a flaky test as we need a way to flush all written traces")
```
draffensperger: If you add a short wait for the server to start like the test above for the grpc-gateway, will that work?

The way that the test above at least attempts to flush all written traces is by using the `octrace.WithSpanBufferCount(1)`. Does that not work here?

receiver/opencensus/opencensus_test.go
```diff
@@ -174,6 +177,101 @@ func TestGrpcGatewayCors_endToEnd(t *testing.T) {
...
+	protoAffiliatedContentTypes := []string{"application/grpc", "application/grpc+proto"}
+	for _, contentType := range protoAffiliatedContentTypes {
+		if err := runContentTypeTests(addr, asContentType, contentType); err != nil {
+			t.Errorf("%q Content-type failed to send proto: %v", contentType, err)
+		}
+	}
+
+	// Before we exit we have to verify that we got exactly 4 TraceService requests.
+	wantLen := len(protoAffiliatedContentSubTypes) + len(protoAffiliatedContentTypes)
+	gotReqs := cbts.AllTraces()
```
draffensperger: Would adding a short delay before this give enough time for the local grpc writes to complete?

receiver/opencensus/opencensus_test.go
odeke-em: @draffensperger, yes I did that if you check in the code below.

receiver/opencensus/opencensus_test.go
odeke-em: I tried a bit of that but no dice, which is why I ended up with the t.Skip* for further investigation.

odeke-em:
>LGTM overall, please try putting a short wait in the unit tests to see if it works, as @draffensperger suggested.

Thanks for the review @songy23 and @draffensperger. @songy23 I tried all those but in vain, hence the t.Skip*

PR #373, opened 2019-02-12
ZipkinV1 JSON Receiver: handle LOCAL_COMPONENT binary annotation

### LINKED ISSUE #352: zipkinv1 json: map LOCAL_COMPONENT to "component"
While testing the fix for #348 with Traefik and various combinations of sending format and receivers it became clear that we should map LOCAL_COMPONENT to OpenTracing tag "component". This was done on the [fix](https://github.com/census-instrumentation/opencensus-service/pull/350) for #348 but need to be ported back to zipkin v1 json.

pjanotti:
The LOCAL_COMPONENT key is used to express the "component or namespace of a local span" in Zipkin V1. This should be used as a fallback for service name if other information is not available for the span.

The LOCAL_COMPONENT, if available, is one of the binary annotations on the span. This change tries to capture it when processing the binary annotations, but, only uses it if none of the other sources for service name is available.

Opportunistically adding equivalent test to Zipkin V1 Thrift that already handles LOCAL_COMPONENT correctly.

Fix #352

### DIFF
--- a/translator/trace/testdata/zipkin_v1_local_component.json
+++ b/translator/trace/testdata/zipkin_v1_local_component.json
@@ -1,0 +1,17 @@
+[
+    {
+        "traceId": "0ed2e63cbe71f5a8",
+        "name": "checkStock",
+        "id": "fe351a053fbcac1f",
+        "parentId": "0ed2e63cbe71f5a8",
+        "timestamp": 1544805927453923,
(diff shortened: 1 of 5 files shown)


pjanotti:
Thanks for the review @songy23 and @sjkaris

PR #376, opened 2019-02-12
Allow all exporters to be batched and queued

sjkaris:
Adds the ability for all exporters to be batched and queued (part 2 of the original batching PR #361) 

Testing Done: unit tests


cmd/occollector/app/collector/processors.go
```diff
@@ -97,11 +83,24 @@ func buildQueuedSpanProcessor(
...
+	if spanSender != nil {
+		allSendersAndExporters = append(allSendersAndExporters, spanSender)
+	}
+	if len(traceExporters) > 0 {
+		allSendersAndExporters = append(
+			allSendersAndExporters, processor.NewTraceExporterProcessor(traceExporters...),
+		)
+	}
+
+	sender := processor.NewMultiSpanProcessor(allSendersAndExporters)
```
pjanotti: I think we need one instance of the queue per destination instead, otherwise failure to one destination will cause retry of the same data on all destinations.

cmd/occollector/app/collector/processors.go
```diff
@@ -101,9 +89,20 @@ func buildQueuedSpanProcessor(
 			sender.HTTPTimeout(thriftHTTPSenderOpts.Timeout),
 		)
 	}
+	doneFns, traceExporters, _ := createExporters(opts.RawConfig, logger)
+
+	if spanSender == nil && len(traceExporters) == 0 {
+		logger.Fatal("Unrecognized sender type or no exporters configured")
```
songy23: minor nit: may be good to differentiate the two cases.

exporter/exporterparser/datadog.go
```diff
@@ -18,50 +18,46 @@ import (
...
 	exporter *datadog.Exporter
 }
 
-// DatadogTraceExportersFromYAML parses the yaml bytes and returns an exporter.TraceExporter targeting
+// DatadogTraceExportersFromViper unmarshals the viper and returns an exporter.TraceExporter targeting
 // Datadog according to the configuration settings.
-func DatadogTraceExportersFromYAML(config []byte) (tes []exporter.TraceExporter, mes []exporter.MetricsExporter, doneFns []func() error, err error) {
+func DatadogTraceExportersFromViper(v *viper.Viper) (tes []exporter.TraceExporter, mes []exporter.MetricsExporter, doneFns []func() error, err error) {
 	var cfg struct {
-		Exporters *struct {
```
odeke-em: Does this radically change what the DataDog exporter will look like in the YAML file?

exporter/exporterparser/prometheus_test.go
```diff
@@ -33,22 +36,24 @@ func TestPrometheusExporter(t *testing.T) {
 	}{
 		{
 			config: `
-exporters:
```
odeke-em: Changing this is a little worrying, does this mean that we shall no longer have sections?

odeke-em:
Thank you for working on this @sjkaris!

I have a bit of a concern:
a) Is this PR just changing from YAML to Viper?
b) How come in the code and in tests we are switching from
```yaml
exporters:
   <exporter_name>: 
      <exporter_config>
```

to

```yaml
<exporter_name>:
   <exporter_config>
```

This change seems more of a Viper change than batching, but if I am not mistaken, please send that change separately as it'll make it easier to review but also to catch bugs and keep batching for exporters focused too.

Thank you.

exporter/exporterparser/datadog.go
sjkaris: No, it keeps it the same

sjkaris:
This PR doesnt change the way the config is written at all. 

Basically, with this change, the `prometheus` exporter package no longer cares about its parent sections. Instead, `internal/config/config.go` and `cmd/occollector/app/collector/processors.go` are responsible for the parent paths. 

This change is actually only for batching, the viper changes are done since without it there is no way to instantiate the exporters without a root reference.

exporter/exporterparser/zipkin_test.go
```diff
@@ -142,12 +143,14 @@ func TestZipkinExportersFromYAML_roundtripJSON(t *testing.T) {
...
-exporters:
-    zipkin:
-      upload_period: 1ms
-      endpoint: ` + cst.URL
-	tes, _, doneFns, err := ZipkinExportersFromYAML([]byte(config))
-	if err != nil {
+zipkin:
+  upload_period: 1ms
+  endpoint: ` + cst.URL
+	v := viper.New()
```
odeke-em: This pattern is quite common throughout the code and most likely might make writing tests warrant reading through viper but also bytes.NewBuffer.

How about let's make an internal helper to do all this parsing?
```go
func ViperFromYAML(yamlBlob []byte) (*viper.Viper, error) {
   v := viper.New()
   v.SetConfigType("yaml")
   if err := v.ReadConfig(bytes.NewBuffer(yamlBlob)); err != nil {
       return nil, err
   }
   return v, nil
}
```

cmd/ocagent/main.go
```diff
@@ -87,7 +89,10 @@ func runOCAgent() {
 		log.Fatalf("Could not instantiate logger: %v", err)
 	}
 
-	traceExporters, metricsExporters, closeFns, err := config.ExportersFromYAMLConfig(logger, yamlBlob)
+	// TODO(skaris): move the rest of the configs to use viper
+	v := viper.New()
```
odeke-em: We forgot to set below `v.SetConfigType("yaml")` which directly goes to my suggestion to extract such code into a single helper.

PR #382, opened 2019-02-13
Update README for intelligent sampling and batching

flands:


README.md
```diff
@@ -307,6 +292,56 @@ The collector also serves as a control plane for agents/clients by supplying
...
+
+### <a name="tail-sampling"></a>Intelligent Sampling
+
+```yaml
+sampling:
+  mode: tail
+  # amount of time from seeing the first span in a trace until making the sampling decision
+  decision-wait: 10s
+  # maximum number of traces kept in the memory
+  num-traces: 500
```
pjanotti: nit: put a higher number 10000 seems reasonable

README.md
```diff
@@ -307,6 +292,56 @@ The collector also serves as a control plane for agents/clients by supplying
...
+    # user-defined policy name
+    my-rate-limiting:
+      # exporters the policy applies to
+      exporters:
+        - jaeger
+        - omnition
+      policy: rate-limiting
+      configuration:
+        spans-per-second: 10
+#      policy: string-tag-filter
```
pjanotti: I think that these comments here will be confusing. If we want to show those here as example we should have other user defined policies.

README.md
```diff
@@ -307,6 +292,56 @@ The collector also serves as a control plane for agents/clients by supplying
...
+  policies:
+    # user-defined policy name
+    my-rate-limiting:
+      # exporters the policy applies to
+      exporters:
+        - jaeger
+        - omnition
+      policy: rate-limiting
+      configuration:
+        spans-per-second: 10
```
pjanotti: nit: 100 or 1000 :)

README.md
pjanotti: nit: use a higher number 10K seems reasonable here

README.md
pjanotti: I think these here as a comment are confusing. Either we added user defined policies and use them or do not include them.

README.md
pjanotti: nit: 100 seems a better number.

README.md
sjkaris: This can go in the receivers/ readme I think

pjanotti: Good idea @sjkaris and we can add all policies currently available.

flands:
@pjanotti @sjkaris addressed feedback -- please check again

README.md
```diff
@@ -307,6 +292,56 @@ The collector also serves as a control plane for agents/clients by supplying
...
+        values:
+          - value1
+          - value2
+    my-numeric-tag-filter:
+      exporters:
+        - jaeger
+        - omnition
+      policy: numeric-tag-filter
+      configuration:
+        tag: <tag-name>
```
pjanotti: nit: do not use `<` and `>` on the tag name, use tag2 or something like that.

exporter/exporterparser/README.md
```diff
@@ -2,12 +2,55 @@ A variety of exporters are available to the OpenCensus Service (both Agent and C
...
+    # user-defined policy name
+    my-rate-limiting:
+      # exporters the policy applies to
+      exporters:
+        - jaeger
+        - omnition
+      policy: rate-limiting
+      configuration:
+        spans-per-second: 1000
+#      policy: string-tag-filter
```
pjanotti: I favor removing the comments, I can add some examples with all other policies and their configuration to other location.

flands:
@pjanotti addressed -- check now

Issue #375, opened 2019-02-12
Travis ci sometimes hangs when running tests


sjkaris:
```
go test -v -race ./...
?   	github.com/census-instrumentation/opencensus-service/cmd/ocagent	[no test files]
?   	github.com/census-instrumentation/opencensus-service/cmd/occollector	[no test files]
=== RUN   TestReceiversEnabledByPresenceWithDefaultSettings
--- PASS: TestReceiversEnabledByPresenceWithDefaultSettings (0.00s)
=== RUN   TestReceiversDisabledByPresenceWithDefaultSettings
--- PASS: TestReceiversDisabledByPresenceWithDefaultSettings (0.00s)
=== RUN   TestMultiAndQueuedSpanProcessorConfig
--- PASS: TestMultiAndQueuedSpanProcessorConfig (0.00s)
PASS
ok  	github.com/census-instrumentation/opencensus-service/cmd/occollector/app/builder	1.022s
?   	github.com/census-instrumentation/opencensus-service/cmd/occollector/app/collector	[no test files]
?   	github.com/census-instrumentation/opencensus-service/cmd/occollector/app/sender	[no test files]
?   	github.com/census-instrumentation/opencensus-service/data	[no test files]
?   	github.com/census-instrumentation/opencensus-service/example	[no test files]
?   	github.com/census-instrumentation/opencensus-service/exporter	[no test files]
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

I believe this is somwhere in running the tests in `github.com/census-instrumentation/opencensus-service/exporter/exporterparser`

pjanotti:
I was about to open this issue but I got this on my box...

odeke-em:
That just seems like Travis CI hanging, this happens spuriously for even other projects and even at times before even running a single test.

songy23:
> That just seems like Travis CI hanging, this happens spuriously for even other projects and even at times before even running a single test.

+1. I've seen similar issues in other projects from time to time. This seems to be a Travis issue and is not actionable on our end.

sjkaris:
That is unfortunate. Closing this out then.

pjanotti:
We need to remember that by default golang cache test results, as result, devs have to do extra steps to run the full set of tests. It is likely that Travis is running all the tests much more frequently than we so if there is some test deadlocking/flaky it is much more likely for us to see it on Travis than on dev boxes.

pjanotti:
[...]

PR #383, opened 2019-02-13
Makefile: add a timeout to tests

pjanotti:
Adding a timeout to tests to help cache any test that maybe deadlocking with certain frequency.

This is related to #375


sjkaris:
Thanks!

PR #391, opened 2019-02-14
Make TestZipkinExportersFromViper_roundtripJSON reliable

### LINKED ISSUE #181: PR Travis run failure: data race on TestZipkinExportersFromYAML_roundtripJSON

pjanotti:
The test depends on the 3 spans from the test data to arrive as a single batch. However, the zipkin exporter does not exposes this fine control of the zipkin reporter used to send the spans. In principle we could expose an option to adjust zipkin reporter batch size but that still requires the test to coordinate the send/receive of the batch. A simpler solution is to implement a mock of the zipkin reporter that behaves synchronously, avoiding any synchronization issue. This also avoids more API surface on the product code.

Fix #181 
Related to #375

### DIFF
--- a/exporter/exporterparser/zipkin_test.go
+++ b/exporter/exporterparser/zipkin_test.go
@@ -17,6 +17,8 @@ package exporterparser
 import (
 	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


pjanotti:
Thanks for reviewing @songy23 and @sjkaris

PR #394, opened 2019-02-14
Use Zipkin "lc" annotation as last resort for service name

### LINKED ISSUE #393: zipkinv1: prefer Host.ServiceName over LOCAL_COMPONENT
Per https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift#L215-L239 LOCAL_COMPONENT is a complement to service name available on the Host field of a binary annotation. 

The code currently gives priority to LOCAL_COMPONENT over Host.ServiceName and it should be the reverse. Notice that we did find some binary annotations without the Host field, in this case we should keep falling back to LOCAL_COMPONENT if available.

pjanotti:
Per zipkinCore.Thrift comments for "lc" (LOCAL_COMPONENT) it should be used to provide the "component" of a local span, however, the service context should come from BinnaryAnnotation.host. The recent changes were overwriting binary annotation host information if the "lc" was available, however, "lc" should be the last resort. This change only uses "lc" if no endpoint or binary annotation host is available.

There is logic duplication and very similar code for both zipkin v1 json and Thrift, the issue is that to have a single piece of code we would need to convert to a single format or create some interface to work over both formats, neither looks worth to me at this point.

Fix #393


odeke-em:
Please update your commit message accordingly too to make it easy to examine logs :)

pjanotti:
Thanks for reviewing @odeke-em and @songy23!

PR #395, opened 2019-02-14
Improve oc-receiver and oc-exporter span throughput

sjkaris:
OC collector reception and export of spans did not scale up. This change
removes the `bundler` from the oc span receiver in favor of go-routines,
and additionally adds a worker count to the oc-exporter.

Note: there still is the limitation on gRPC message size that prevents batches of > 4M (uncompressed). I found this to be an acceptable limit, so did nothing here. With compression, under very high-throughput usecases, 4M will likely not be enough, so there is potential to followup with another change to modify the max message size.


exporter/exporterparser/opencensus.go
```diff
@@ -72,28 +78,38 @@ func OpenCensusTraceExportersFromViper(v *viper.Viper) (tes []exporter.TraceExpo
...
-		sde.Flush()
-		return nil
-	})
 	return tes, mes, doneFns, nil
 }
 
-func (sde *ocagentExporter) ExportSpans(ctx context.Context, td data.TraceData) error {
-	err := sde.exporter.ExportTraceServiceRequest(
+func (oce *ocagentExporter) ExportSpans(ctx context.Context, td data.TraceData) error {
+	err := oce.exporters[rand.Intn(len(oce.exporters))].ExportTraceServiceRequest(
```
pjanotti: Can you compare this versus a simple round-robin instead? Just thinking that under load you may get a sequence of requests going to a single exporter...

exporter/exporterparser/opencensus.go
```diff
@@ -72,28 +78,38 @@ func OpenCensusTraceExportersFromViper(v *viper.Viper) (tes []exporter.TraceExpo
...
+	numWorkers := defaultNumWorkers
+	if ocac.NumWorkers > 0 {
+		numWorkers = ocac.NumWorkers
+	}
+
+	exporters := make([]*ocagent.Exporter, 0, numWorkers)
+	for exporterIndex := 0; exporterIndex < numWorkers; exporterIndex++ {
+		exporter, serr := ocagent.NewExporter(opts...)
+		if serr != nil {
+			return nil, nil, nil, fmt.Errorf("Cannot configure OpenCensus Trace exporter: %v", serr)
```
pjanotti: nit: usually do not start error messages with capital letters.

sjkaris:
I don't have any formal benchmarking numbers, but with these changes (and a little config tuning) my laptop can do 60k+ spans per second when running an oc-collector to another oc-collector over gRPC with oc-proto. It probably could go faster, but the 4M gRPC message limit starts to hurt us some (limits batch sizes to 2-3k spans), as well as the overhead of running two collectors + the load generator locally. Prior to these changes I was unable to get anywhere near that throughput. I didn't see what exactly we capped out at before the first collector started dropping spans (since either its oc-exporter or the second's oc-receiver couldn't keep up). I think hitting this high throughput is very promising though.

odeke-em:
Thanks for working on this @sjkaris!

While this definitely increases throughput since we are sharing n workers in goroutines and then subsequent goroutines per received ExportTraceServiceRequest, I am curious/concerned about the growth and number of goroutines that this creates. Could be a potential DOS vector because now high traffic will by a factor increase the number of goroutines running at anytime which will kill scalability if all the cores are saturated but also the amount of CPU since even sending one span per batch will cause a goroutine

sjkaris:
The bundler, although it hides it somewhat from you, actually also launches goroutines (https://github.com/googleapis/google-api-go-client/blob/master/support/bundler/bundler.go#L266), so the overall impact of the change in terms of # of goroutines is smaller than it would initially seem. Additionally, I believe spawning a goroutine per request is pretty standard across go server designs, for example, the go http server spawns a goroutine per request (https://golang.org/src/net/http/server.go#L2851)

odeke-em:
>The bundler, although it hides it somewhat from you, actually also launches goroutines

@sjkaris of course it does launch goroutine! However those are goroutines launched on:
* Threshold counts exceeded
* Flush

and that's bounded concurrency.

>so the overall impact of the change in terms of # of goroutines is smaller than it would initially seem.

Empirically at least from this PR, that doesn't seem true. This PR creates a goroutine for every single message created. Basically it doesn't do any batching just for the sake of improving ingestion throughput.

>Additionally, I believe spawning a goroutine per request is pretty standard across go server designs, for example, the go http server spawns a goroutine per request

A server is a different beast as it is general purpose for running programs via protocol RPCs, and thus it is upto the user to provide boundedness or scale the service. Even then I've been plagued with machines running out of resources because of such unboundedness as well as many projects that are plagued with concurrency overloads.

Personally I have dealt with plagues of unbounded goroutines and I am not comfortable with merging this change as it is. Completely removing batching IMHO is a nuclear option.

### Suggestion/middle-ground
However, perhaps we can meet in the middle with what I believe is an idea that melds both of our directions -- I really like your idea of n-workers but add making each worker have a bundler for batching. Basically the problem that you are trying to solve is ingestion bandwidth but scale that concurrently and also improve output throughput. I'd rather have n-workers that still perform batching will theoretically scale your concern by almost n if n == number of cores.

Great change and ideas though!

sjkaris:
I disagree. I personally think worrying about the # of go routines is a very premature worry, since we would need to be at 5k-10k+ _requests_ (not spans) per second for it to matter. We currently can't get anywhere close to that throughput, so this change is only an improvement.  

Also this "bundling" is not really batching -- its simply calling process in a synchronous for loop. All this does is block faster smaller messages on slower bigger ones. Additionally, this "bundling" adds artificial latency (its timer) without any real benefit. It also creates a bursty pattern, in which we try to start a lot of work all at once instead of spreading it out.

IMO we can think about adding a worker pool -- but we aren't in a place where we need that yet, and certainly this is strictly an efficiency enhancement over the current setup.

odeke-em:
>Also this "bundling" is not really batching -- its simply calling process in a synchronous for loop. All this does is block faster smaller messages on slower bigger ones. Additionally, this "bundling" adds artificial latency (its timer) without any real benefit. It also creates a bursty pattern, in which we try to start a lot of work all at once instead of spreading it out.

Am now confused at what you mean by "batching" -- regardless of how unperformant it is, it is doing batching i.e. exporting a bunch of items together instead of individually, no?

>I disagree. I personally think worrying about the # of go routines is a very premature worry, since we would need to be at 5k-10k+ _requests_ (not spans) per second for it to matter. We currently can't get anywhere close to that throughput, so this change is only an improvement.

I think that we can agree to disagree. Unbounded concurrency is a big concern to worry about from day 1 IMHO. I personally am not too keen on this change as is so I won't provide an approval, but others seem to agree with it so I'll let consensus rule here to avoid wasting too much time on something that  I perhaps might be wrong on or prematurely worrying about. In the future, let's also do more experiments and provide some benchmarks and numbers and also keep our eyes open for how to fix any reported problems.

Thanks for the discourse.

sjkaris:
@odeke-em  thanks -- I will followup with a PR adding workers

PR #396, opened 2019-02-15
Add sampled / not sampled metrics per policy

sjkaris:


internal/collector/processor/tail_sampling/processor.go
```diff
@@ -249,9 +251,11 @@ func (tsp *tailSamplingSpanProcessor) ProcessSpans(batch *agenttracepb.ExportTra
 						zap.Uint64("failCount", failCount),
 						zap.Error(err))
 				}
+				stats.Record(policyAndDests.ctx, statCountTracesSampled.M(int64(1)))
```
pjanotti: This and the one below are not needed since the trace was already counted: here it is just deciding what to do with spans that are arriving after the decision.

internal/collector/processor/tail_sampling/metrics.go
```diff
@@ -34,6 +34,9 @@ var (
 
 	statPolicyEvaluationErrorCount = stats.Int64("sampling_policy_evaluation_error", "Count of sampling policy evaluation errors", stats.UnitDimensionless)
 
+	statCountTracesSampled    = stats.Int64("count_traces_sampled", "Count of traces that were sampled", stats.UnitDimensionless)
+	statCountTracesNotSampled = stats.Int64("count_traces_not_sampled", "Count of traces that were sampled", stats.UnitDimensionless)
```
pjanotti: "NOT" sampled :)

odeke-em:
@sjkaris you might want to consider using the same measure but tag values e.g.
tag_key: `sample_status` or `sampled`
tag_values: ["sampled", "not_sampled"] or ["true", "false"]

for which many backends can allow you to group and filter by tag_key equivalents.

/cc @bogdandrutu whom I learnt about this from.

sjkaris:
Ah, I like that -- changing

PR #400, opened 2019-02-15
Fix flaky test for Zipkin LOCAL_COMPONENT handling

pjanotti:
Internally the function being tested uses a map to generate the slice with the translation results, however, the order is not deterministic and some runs fail. There are just two elements on the translated results but to make clear the expectation of the test this change uses sort.Slice function to achieve the desired order.


pjanotti:
Thanks for looking @songy23 and @sjkaris.

PR #403, opened 2019-02-15
exporter/exporterparser/all: update SpanData translator

odeke-em:
Update the OpenCensus Proto TraceData
to OpenCensus-Go SpanData translator by:
* Name from:
    "exportSpans"
to
    "ocProtoSpansToOCSpanDataInstrumented"
to make it more intuitive but also declare
its intentions and add a reminder that we
should get vendors to start using OpenCensus Proto

* Added tracing for each export with a span that'll
record the errors and number of spans

### DIFF
--- a/exporter/exporterparser/exparser.go
+++ b/exporter/exporterparser/exparser.go
@@ -32,7 +32,32 @@ import (
 	tracetranslator "github.com/census-instrumentation/opencensus-service/translator/trace"
 )
 
-func exportSpans(ctx context.Context, exporterName string, te trace.Exporter, td data.TraceData) error {
+// ocProtoSpansToOCSpanDataInstrumented converts
+// OpenCensus Proto TraceData to OpenCensus-Go SpanData.
+// The "Instrumented" suffix serves to document that this
+// function is traced but also has stats for self-observability.
+//
+// This is a bootstrapping mechanism for us to re-use as many of
+// the OpenCensus-Go trace.SpanData exporters which were written
+// by various vendors and contributors. Eventually the goal is to
+// get those exporters converted to directly receive
+// OpenCensus Proto TraceData.
+func ocProtoSpansToOCSpanDataInstrumented(ctx context.Context, exporterName string, te trace.Exporter, td data.TraceData) (aerr error) {
+	ctx, span := trace.StartSpan(ctx,
+		"opencensus.service.exporter."+exporterName+".ExportTrace",
+		trace.WithSampler(trace.NeverSample()))
+
+	span.Annotate([]trace.Attribute{
+		trace.Int64Attribute("n_spans", int64(len(td.Spans))),
+	}, "")
+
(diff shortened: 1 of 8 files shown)


exporter/exporterparser/exparser.go
```diff
@@ -32,7 +32,32 @@ import (
...
+	ctx, span := trace.StartSpan(ctx,
+		"opencensus.service.exporter."+exporterName+".ExportTrace",
+		trace.WithSampler(trace.NeverSample()))
+
+	span.Annotate([]trace.Attribute{
+		trace.Int64Attribute("n_spans", int64(len(td.Spans))),
+	}, "")
+
+	defer func() {
+		if aerr != nil {
```
pjanotti: fancy golang thingy :) for a bit I was thinking where is this being set ...

exporter/exporterparser/exparser.go
```diff
@@ -32,7 +32,32 @@ import (
...
+// The "Instrumented" suffix serves to document that this
+// function is traced but also has stats for self-observability.
+//
+// This is a bootstrapping mechanism for us to re-use as many of
+// the OpenCensus-Go trace.SpanData exporters which were written
+// by various vendors and contributors. Eventually the goal is to
+// get those exporters converted to directly receive
+// OpenCensus Proto TraceData.
+func ocProtoSpansToOCSpanDataInstrumented(ctx context.Context, exporterName string, te trace.Exporter, td data.TraceData) (aerr error) {
+	ctx, span := trace.StartSpan(ctx,
```
pjanotti: q: what are the plans for these spans? iirc no trace exporter is setup and for now they are never sampled.

exporter/exporterparser/zipkin.go
```diff
@@ -192,7 +192,18 @@ func (ze *zipkinExporter) stop() error {
 	return ze.reporter.Close()
 }
 
-func (ze *zipkinExporter) ExportSpans(ctx context.Context, td data.TraceData) error {
+func (ze *zipkinExporter) ExportSpans(ctx context.Context, td data.TraceData) (zerr error) {
+	ctx, span := trace.StartSpan(ctx,
+		"opencensus.service.exporter.Zipkin.ExportTrace",
```
pjanotti: nit: all others are lower case.

exporter/exporterparser/exparser.go
odeke-em: zPages use them

exporter/exporterparser/zipkin.go
odeke-em: Roger that, thanks!

exporter/exporterparser/exparser.go
odeke-em: Yahh digg!

odeke-em:
Thank you for the reviews @songy23 @sjkaris @pjanotti!

Issue #369, opened 2019-02-07
cmd/*: Reduce the number of dependencies loaded at boot up time.


songy23:
Kind of a revisit of https://github.com/census-instrumentation/opencensus-service/pull/70 but @bogdandrutu has some new ideas on it.

bogdandrutu:
Here, it will be good if we have all the dependencies tied together in the main function, something like:

main {
  list_of_available_receivers = ...
  list_of_available_blocks = ...
  list_of_available_exporters = ...
  ValidateConfig(list_of_available_receivers, list_of_available_blocks, list_of_available_exporters, config);
  StartAgent(list_of_available_receivers, list_of_available_blocks, list_of_available_exporters, config);
}

This way only the main.go will link all the dependencies, so people can simply re-write the main.go file and build an agent with fewer dependencies.

songy23:
@odeke-em mentioned he has a probably simpler solution with dependency management.

songy23:
Closing, this will be fixed with the new config.

PR #406, opened 2019-02-15
Move every exporter in it's own directory.

bogdandrutu:
With this change we split a large package exporterparser (in terms of dependencies) and make sure that we don't download the entire internet if depend on one of the exporters. 

Updates https://github.com/census-instrumentation/opencensus-service/issues/369

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -26,7 +26,15 @@ import (
 	yaml "gopkg.in/yaml.v2"
 
 	"github.com/census-instrumentation/opencensus-service/exporter"
-	"github.com/census-instrumentation/opencensus-service/exporter/exporterparser"
+	"github.com/census-instrumentation/opencensus-service/exporter/awsexporter"
+	"github.com/census-instrumentation/opencensus-service/exporter/datadogexporter"
+	"github.com/census-instrumentation/opencensus-service/exporter/honeycombexporter"
+	"github.com/census-instrumentation/opencensus-service/exporter/jaegerexporter"
+	"github.com/census-instrumentation/opencensus-service/exporter/kafkaexporter"
+	"github.com/census-instrumentation/opencensus-service/exporter/opencensusexporter"
+	"github.com/census-instrumentation/opencensus-service/exporter/prometheusexporter"
+	"github.com/census-instrumentation/opencensus-service/exporter/stackdriverexporter"
+	"github.com/census-instrumentation/opencensus-service/exporter/zipkinexporter"
 	"github.com/census-instrumentation/opencensus-service/receiver/prometheusreceiver"
 )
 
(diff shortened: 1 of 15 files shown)


internal/config/config.go
```diff
@@ -397,15 +405,15 @@ func ExportersFromViperConfig(logger *zap.Logger, v *viper.Viper) ([]exporter.Tr
...
-		{name: "datadog", fn: exporterparser.DatadogTraceExportersFromViper},
-		{name: "stackdriver", fn: exporterparser.StackdriverTraceExportersFromViper},
-		{name: "zipkin", fn: exporterparser.ZipkinExportersFromViper},
-		{name: "jaeger", fn: exporterparser.JaegerExportersFromViper},
-		{name: "kafka", fn: exporterparser.KafkaExportersFromViper},
-		{name: "opencensus", fn: exporterparser.OpenCensusTraceExportersFromViper},
-		{name: "prometheus", fn: exporterparser.PrometheusExportersFromViper},
-		{name: "aws-xray", fn: exporterparser.AWSXRayTraceExportersFromViper},
-		{name: "honeycomb", fn: exporterparser.HoneycombTraceExportersFromViper},
+		{name: "datadog", fn: datadogexporter.DatadogTraceExportersFromViper},
```
sjkaris: Now that these are separate packages, we should rename the functions to just `FromViper(...)` perhaps, so we don't have the "stutter" that people sometime talk about.

odeke-em:
@bogdandrutu thanks for this change!

>With this change we split a large package exporterparser (in terms of dependencies) and make sure that we don't download the entire internet if depend on one of the exporters.

That isn't true specifically for this repository, unless you have a secondary code base for which you are doing imports.

Overall though, great change and it provides better separation, I don't know why we hadn't yet done this separation. Also please rebase, there are a bunch of conflicts from my last PR that was up before this one :)

internal/config/config.go
odeke-em: I second @sjkaris' suggestion!

pjanotti: I don't know how far you want to go but there are a series of similar good renames after they go to their own packages, e.g.: zipkinexporter.ZipkinConfig -> zipkinexporter.Config ...

bogdandrutu: @pjanotti happy to review some PRs :+1: 

I would play with the code more, but I think this is a good start.

pjanotti: It is a great start! I will try to help with some of those too.

bogdandrutu:
@odeke-em next step is to add a Builder/Factory interface that every exporter should implement. Then the config will accept a list of linked exporters, so the only place where you get a dependency on the exporter is in the main function which everyone can overwrite. This way I can build my own agent with less dependencies (sure in a different repo)

PR #408, opened 2019-02-16
remove unused vendor directory

odeke-em:
Remove this vestige of back in the day before we used Go modules
and had to manually vendor a copy of packages in here.


odeke-em:

PR #410, opened 2019-02-16
Switch oc receiver to use a worker pool instead of pure go-routines

sjkaris:
As discussed on #395 , use a worker pool to limit the # of go-routines.


receiver/opencensusreceiver/octrace/opencensus.go
```diff
@@ -73,6 +80,20 @@ func (oci *Receiver) Export(tes agenttracepb.TraceService_ExportServer) error {
 		return err
 	}
 
+	messageChan := make(chan *data.TraceData, 64)
```
odeke-em: How did you come to choosing this capacity? A value that you gave you some optimal throughput?

odeke-em: Perhaps let's add a comment or make this a const for easier discoverability and document it too.

receiver/opencensusreceiver/octrace/opencensus.go
```diff
@@ -93,7 +114,13 @@ func (oci *Receiver) Export(tes agenttracepb.TraceService_ExportServer) error {
 			resource = recv.Resource
 		}
 
-		go oci.export(ctxWithReceiverName, tes, lastNonNilNode, resource, recv.Spans)
```
odeke-em: Nice, thank you!

receiver/opencensusreceiver/octrace/opencensus.go
pjanotti: It feels like the setup of the workers should happen before the call to Export. Or is the intent really to have a channel (and respective workers) per call to Export?

sjkaris: I went back and forth on this -- the previous code had does the grouping per connection (per export).. Thinking more on it, I think it is probably better to have the worker pool per receiver as you are suggesting.

bogdandrutu:
Please ignore my approvals, I was testing codeowners.

receiver/opencensusreceiver/octrace/opencensus.go
```diff
@@ -24,26 +24,53 @@ import (
...
-		opt.WithReceiver(oci)
+		opt(oci)
+	}
+
+	// Setup and startup worker pool
+	workers := make([]*receiverWorker, 0, oci.numWorkers)
+	for index := 0; index < oci.numWorkers; index++ {
+		worker := newReceiverWorker(oci)
+		go worker.listenOn(messageChan)
+		workers = append(workers)
```
pjanotti: this doesn't look correct

receiver/opencensusreceiver/octrace/opencensus.go
```diff
@@ -107,17 +143,45 @@ func (oci *Receiver) Export(tes agenttracepb.TraceService_ExportServer) error {
...
-	tes agenttracepb.TraceService_ExportServer,
-	node *commonpb.Node,
-	resource *resourcepb.Resource,
-	spans []*tracepb.Span,
-) {
-	tracedata := data.TraceData{Node: node, Resource: resource, Spans: spans}
-	// We MUST unconditionally record metrics from this reception.
-	spansMetricsFn := internal.NewReceivedSpansRecorderStreaming(tes.Context(), receiverName)
-	spansMetricsFn(tracedata.Node, tracedata.Spans)
+// Stop stops this receivers workers
```
pjanotti: nit: suggestion: // Stop the receiver and its workers.

receiver/opencensusreceiver/octrace/opencensus.go
```diff
@@ -24,26 +24,53 @@ import (
...
 
 // New creates a new opencensus.Receiver reference.
 func New(sr receiver.TraceReceiverSink, opts ...Option) (*Receiver, error) {
 	if sr == nil {
 		return nil, errors.New("needs a non-nil receiver.TraceReceiverSink")
 	}
-	oci := &Receiver{spanSink: sr}
+
+	messageChan := make(chan *traceDataWithCtx, 64)
+	oci := &Receiver{
```
pjanotti: nit: not from your change but `oci` was when "interceptor" was used, but we changed it to "receiver", seems a good opportunity to change it.

receiver/opencensusreceiver/octrace/opencensus.go
```diff
@@ -107,17 +143,45 @@ func (oci *Receiver) Export(tes agenttracepb.TraceService_ExportServer) error {
...
+}
+
+func newReceiverWorker(receiver *Receiver) *receiverWorker {
+	return &receiverWorker{
+		receiver: receiver,
+		cancel:   make(chan struct{}),
+	}
+}
+
+func (rw *receiverWorker) listenOn(cn chan *traceDataWithCtx) {
```
pjanotti: nit: declare the direction of the channel on the function

PR #411, opened 2019-02-16
Update go mod and sum after running make.

bogdandrutu:
I don't really understand mod/sum but on my macbook these files changed after I ran `make`


bogdandrutu:
@odeke-em just done that. more changes, seriously not sure what I am doing here but I trust you :)

odeke-em:
Yeah that was pruning out old changes, for example since that genproto code was updated, the old one is no longer needed in the manifest but also it added some that it saw missing in other dependencies.

go.sum
```diff
@@ -6,8 +6,6 @@ cloud.google.com/go v0.32.0 h1:DSt59WoyNcfAInilEpfvm2ugq8zvNyaHAm9MkzOwRQ4=
 cloud.google.com/go v0.32.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 contrib.go.opencensus.io/exporter/aws v0.0.0-20181029163544-2befc13012d0 h1:YsbWYxDZkC7x2OxlsDEYvvEXZ3cBI3qBgUK5BqkZvRw=
 contrib.go.opencensus.io/exporter/aws v0.0.0-20181029163544-2befc13012d0/go.mod h1:uu1P0UCM/6RbsMrgPa98ll8ZcHM858i/AD06a9aLRCA=
-contrib.go.opencensus.io/exporter/ocagent v0.4.3 h1:QjNm697iO7CZ09IxxSiCUzOhALENIsLsixdPwjV1yGs=
```
odeke-em: This was pruned out, since now it is stale :)

go.sum
```diff
@@ -459,12 +465,13 @@ google.golang.org/genproto v0.0.0-20180601223552-81158efcc9f2/go.mod h1:JiN7NxoA
 google.golang.org/genproto v0.0.0-20180817151627-c66870c02cf8/go.mod h1:JiN7NxoALGmiZfu7CAH4rXhgtRTLTxftemlI0sWmxmc=
 google.golang.org/genproto v0.0.0-20180831171423-11092d34479b h1:lohp5blsw53GBXtLyLNaTXPXS9pJ1tiTw61ZHUoE9Qw=
 google.golang.org/genproto v0.0.0-20180831171423-11092d34479b/go.mod h1:JiN7NxoALGmiZfu7CAH4rXhgtRTLTxftemlI0sWmxmc=
-google.golang.org/genproto v0.0.0-20181101192439-c830210a61df h1:Ri2mROsxIxitlzRQ0pYoP8/dsqeLEolHrhh29dltSI4=
```
odeke-em: Ditto about not being used anymore though there is code that uses the older versions.

go.sum
bogdandrutu: Thanks for the explanation.

go.sum
bogdandrutu: Thanks for the explanation.

PR #414, opened 2019-02-16
all: ensure traces and stats for all receivers+exporters

odeke-em:
Ensure that every exporter and receiver has traces and stats.
Stats weren't available for any span recordings.

This change introduces a measurer that has an underlying span
and also tracks latency but also tags errors and statuses.

### DIFF
--- a/exporter/exporterparser/exparser.go
+++ b/exporter/exporterparser/exparser.go
@@ -26,7 +26,6 @@ import (
 
 	"go.opencensus.io/trace"
 
-	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
 	"github.com/census-instrumentation/opencensus-service/data"
 	"github.com/census-instrumentation/opencensus-service/internal"
 	tracetranslator "github.com/census-instrumentation/opencensus-service/translator/trace"
@@ -43,28 +42,19 @@ import (
 // get those exporters converted to directly receive
 // OpenCensus Proto TraceData.
 func OcProtoSpansToOCSpanDataInstrumented(ctx context.Context, exporterName string, te trace.Exporter, td data.TraceData) (aerr error) {
-	ctx, span := trace.StartSpan(ctx,
-		"opencensus.service.exporter."+exporterName+".ExportTrace",
-		trace.WithSampler(trace.NeverSample()))
+	var nGoodSpans int64
 
-	span.Annotate([]trace.Attribute{
-		trace.Int64Attribute("n_spans", int64(len(td.Spans))),
-	}, "")
+	observabilityRecorder := internal.NewExporterEventRecorder(exporterName)
 
-	defer func() {
-		if aerr != nil {
-			span.SetStatus(trace.Status{Code: trace.StatusCodeInternal, Message: aerr.Error()})
-		}
-		span.End()
-	}()
+	ctx = observabilityRecorder.Start(ctx, td.Node)
+	defer observabilityRecorder.End(ctx, nGoodSpans, aerr)
 
 	var errs []error
-	var goodSpans []*tracepb.Span
 	for _, span := range td.Spans {
 		sd, err := tracetranslator.ProtoSpanToOCSpanData(span)
 		if err == nil {
 			te.ExportSpan(sd)
-			goodSpans = append(goodSpans, span)
+			nGoodSpans++
 		} else {
 			errs = append(errs, err)
 		}
--- a/exporter/opencensusexporter/opencensus.go
+++ b/exporter/opencensusexporter/opencensus.go
@@ -105,20 +105,19 @@ func OpenCensusTraceExportersFromViper(v *viper.Viper) (tes []exporter.TraceExpo
 	return tes, mes, doneFns, nil
 }
 
-func (oce *ocagentExporter) ExportSpans(ctx context.Context, td data.TraceData) error {
+func (oce *ocagentExporter) ExportSpans(ctx context.Context, td data.TraceData) (err error) {
+	observabilityRecorder := internal.NewExporterEventRecorder("ocagent")
+
+	ctx = observabilityRecorder.Start(ctx, td.Node)
+	defer observabilityRecorder.End(ctx, int64(len(td.Spans)), err)
+
 	// Get an exporter worker round-robin
 	exporter := oce.exporters[atomic.AddUint32(&oce.counter, 1)%uint32(len(oce.exporters))]
-	err := exporter.ExportTraceServiceRequest(
+	return exporter.ExportTraceServiceRequest(
 		&agenttracepb.ExportTraceServiceRequest{
 			Spans:    td.Spans,
 			Resource: td.Resource,
 			Node:     td.Node,
 		},
 	)
-	if err != nil {
-		return err
-	}
-	nSpansCounter := internal.NewExportedSpansRecorder("ocagent")
-	nSpansCounter(ctx, td.Node, td.Spans)
-	return nil
 }
--- a/exporter/zipkinexporter/zipkin.go
+++ b/exporter/zipkinexporter/zipkin.go
@@ -30,7 +30,6 @@ import (
 	"go.opencensus.io/trace"
 
 	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
-	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
 	"github.com/census-instrumentation/opencensus-service/data"
 	"github.com/census-instrumentation/opencensus-service/exporter"
 	"github.com/census-instrumentation/opencensus-service/internal"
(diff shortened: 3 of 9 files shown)


internal/observability.go
```diff
@@ -65,42 +67,111 @@ var ViewExportedSpans = &view.View{
...
+// of activeness of the caller.
+func NewReceiverEventRecorder(receiverName string) SpansObservabilityRecorder {
+	return &spansRecorder{
+		predefinedSpanName: "opencensus.service.receiver." + receiverName + ".Receive",
+		startTime:          time.Now(),
+		receiverName:       receiverName,
+	}
+}
+
+// SpansObservabilityRecorder is a helper that is started when
```
bogdandrutu: incomplete comment.

internal/observability.go
```diff
@@ -65,42 +67,111 @@ var ViewExportedSpans = &view.View{
...
+type SpansObservabilityRecorder interface {
```
bogdandrutu: why is this an interface instead of a struct?

internal/observability.go
```diff
@@ -65,42 +67,111 @@ var ViewExportedSpans = &view.View{
...
-func ContextWithReceiverName(ctx context.Context, receiverName string) context.Context {
-	ctx, _ = tag.New(ctx, tag.Upsert(tagKeyReceiverName, receiverName))
+var _ SpansObservabilityRecorder = (*spansRecorder)(nil)
+
+func (sr *spansRecorder) Start(ctx context.Context, ni *commonpb.Node) context.Context {
+	sr.startTime = time.Now()
+	ctx, sr.span = trace.StartSpan(ctx, sr.predefinedSpanName, trace.WithSampler(trace.NeverSample()))
+
+	if sr.receiverName != "" {
+		ctx, _ = tag.New(ctx, tag.Upsert(tagKeyReceiverName, sr.receiverName))
```
bogdandrutu: Unnecessary overhead on the critical path the result of the`tag.Upsert(tagKeyReceiverName, sr.receiverName)` can be cached and reused.

Same everywhere.

internal/observability.go
```diff
@@ -65,42 +67,111 @@ var ViewExportedSpans = &view.View{
...
 }
 
-// ContextWithReceiverName adds the tag "opencensus_receiver" and the name of the
-// receiver as the value, and returns the newly created context.
-func ContextWithReceiverName(ctx context.Context, receiverName string) context.Context {
-	ctx, _ = tag.New(ctx, tag.Upsert(tagKeyReceiverName, receiverName))
+var _ SpansObservabilityRecorder = (*spansRecorder)(nil)
+
+func (sr *spansRecorder) Start(ctx context.Context, ni *commonpb.Node) context.Context {
+	sr.startTime = time.Now()
```
bogdandrutu: Isn't this duplicated with line 173?

internal/observability.go
```diff
@@ -65,42 +67,111 @@ var ViewExportedSpans = &view.View{
...
+		measurements = append(measurements, mReceivedSpans.M(nItems))
+	} else {
+		tagMutators = append(tagMutators, tag.Upsert(tagKeyExporterName, sr.exporterName))
+		measurements = append(measurements, mExportedSpans.M(nItems))
+	}
+
+	if err == nil {
+		tagMutators = append(tagMutators, tag.Upsert(tagKeyStatus, "OK"))
+	} else {
+		tagMutators = append(tagMutators, tag.Upsert(tagKeyStatus, "ERROR"), tag.Upsert(tagKeyError, err.Error()))
```
bogdandrutu: error is an int without any bounds, I don't think it is fine to put it as a tag in this case, that is the main reason we defined a type of errors (see the status in trace).

I would suggest to use if possible the same mapping that we have in trace (see https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto)

exporter/exporterparser/exparser.go
```diff
@@ -43,28 +42,19 @@ import (
...
+	observabilityRecorder := internal.NewExporterEventRecorder(exporterName)
 
-	defer func() {
-		if aerr != nil {
-			span.SetStatus(trace.Status{Code: trace.StatusCodeInternal, Message: aerr.Error()})
-		}
-		span.End()
-	}()
+	ctx = observabilityRecorder.Start(ctx, td.Node)
+	defer observabilityRecorder.End(ctx, nGoodSpans, aerr)
```
pjanotti: nGoodSpans is going to be always zero when deferred function is called, either it becomes a capture or is passed as a reference.

exporter/zipkinexporter/zipkin.go
```diff
@@ -193,18 +192,13 @@ func (ze *zipkinExporter) stop() error {
...
-	defer func() {
-		if zerr != nil {
-			span.SetStatus(trace.Status{Code: trace.StatusCodeInternal, Message: zerr.Error()})
-		}
-		span.End()
-	}()
+	observabilityRecorder := internal.NewExporterEventRecorder("zipkin")
+
+	ctx = observabilityRecorder.Start(ctx, td.Node)
+	defer observabilityRecorder.End(ctx, nGoodSpans, zerr)
```
pjanotti: same as above

internal/observability.go
```diff
@@ -65,42 +67,111 @@ var ViewExportedSpans = &view.View{
 	TagKeys:     []tag.Key{tagKeyExporterName},
 }
 
+var mLatencyMs = stats.Float64("oc.io/latency", "The latency of the various actions in milliseconds", "ms")
+
+var latencyDistribution = view.Distribution(
+	0, 0.01, 0.05, 0.1, 0.3, 0.6, 0.8, 1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80,
```
pjanotti: less than 0 is never going to be used, right? at least that is the way that this ends up in Prometheus.

internal/observability.go
```diff
@@ -30,12 +31,13 @@ import (
 	"go.opencensus.io/trace"
 
 	commonpb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/common/v1"
-	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
 )
 
 var (
 	tagKeyReceiverName, _ = tag.NewKey("opencensus_receiver")
 	tagKeyExporterName, _ = tag.NewKey("opencensus_exporter")
+	tagKeyError, _        = tag.NewKey("opencensus_error")
```
pjanotti: The prefix "opencensus_" seems a bit to verbose to me: the namespace already should give that this is OpenCensus and then the paris will look strange: opencensus_receiver="zipkin". The case that makes sense is if for instance status is being mapped to the OC status defined in proto.

exporter/exporterparser/exparser.go
pjanotti: deferred function is always going to be invoked with nGoodSpans set to zero, it either needs to be a capture or passed as reference.

exporter/opencensusexporter/opencensus.go
```diff
@@ -105,20 +105,19 @@ func OpenCensusTraceExportersFromViper(v *viper.Viper) (tes []exporter.TraceExpo
 	return tes, mes, doneFns, nil
 }
 
-func (oce *ocagentExporter) ExportSpans(ctx context.Context, td data.TraceData) error {
+func (oce *ocagentExporter) ExportSpans(ctx context.Context, td data.TraceData) (err error) {
+	observabilityRecorder := internal.NewExporterEventRecorder("ocagent")
+
+	ctx = observabilityRecorder.Start(ctx, td.Node)
+	defer observabilityRecorder.End(ctx, int64(len(td.Spans)), err)
```
sjkaris: err here will always be nil I believe. https://play.golang.org/p/kFOauu3Y7YF Similar to above, probably need a closure here

exporter/exporterparser/exparser.go
sjkaris: `aerr` is never set

exporter/zipkinexporter/zipkin.go
sjkaris: `zerr` is never set. Also, as a nit, would prefer just `err` as a name here

internal/observability.go
```diff
@@ -65,42 +67,111 @@ var ViewExportedSpans = &view.View{
...
+	Start(ctx context.Context, ni *commonpb.Node) context.Context
+	End(ctx context.Context, nItems int64, err error) context.Context
+	UnderlyingSpan() *trace.Span
+}
+
+type spansRecorder struct {
+	predefinedSpanName string
+	span               *trace.Span
+	receiverName       string
+	exporterName       string
```
sjkaris: I think you can form the wanted tags on construction (so they don't need to be reconstructed frequently), and then get rid of any receiver/exporter specificity within this struct itself

receiver/jaegerreceiver/trace_receiver.go
```diff
@@ -226,19 +226,26 @@ func (jr *jReceiver) stopTraceReceptionLocked(ctx context.Context) error {
...
 func (jr *jReceiver) SubmitBatches(ctx thrift.Context, batches []*jaeger.Batch) ([]*jaeger.BatchSubmitResponse, error) {
-	jbsr := make([]*jaeger.BatchSubmitResponse, 0, len(batches))
-	spansMetricsFn := internal.NewReceivedSpansRecorderStreaming(ctx, "jaeger-collector")
+	var err error
+	var nSpans int64
+
+	observabilityRecorder := internal.NewExporterEventRecorder("jaeger-collector")
+
+	octx := observabilityRecorder.Start(ctx, nil)
+	defer observabilityRecorder.End(octx, nSpans, err)
```
sjkaris: same problem with `err` and `nSpans` -- I think this exists everywhere its used, so not going to comment on each place

songy23:
Any updates on this PR?

bogdandrutu:
I did something different to achieve the same result https://github.com/census-instrumentation/opencensus-service/pull/484

pjanotti:
> I did something different to achieve the same result #484

@odeke-em I am closing this one in favor of #484 - let's know if you think it should be otherwise.

PR #418, opened 2019-02-18
Re-package tracetranslators by translation type

sjkaris:
repackages trace translators into subpackages for each translation type. This should also help with modularity (using the zipkin receiver only shouldn't pull in jaeger dependencies and vice versa).


translator/trace/spandata/protospan_to_spandata_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tracetranslator_test
+package spandata_test
```
pjanotti: could you please rename it to spandata

translator/trace/big_endian_converter.go
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+package tracetranslator
+
+import (
+	"encoding/binary"
+)
+
+// Int64TraceIDToByteTraceID takes a long representaition of a trace id
+// and converts it to a []byte representation
```
pjanotti: nit: punctuation for the comments (it becomes godoc), same below.

PR #419, opened 2019-02-19
nodebatcher: Allow batcher pre-allocations to reduce on new batch

sjkaris:
Currently, we re-use the previous capacity size when pre-allocating
space for the new batch in nodebatcher. This doesn't work well when a
node has had high throughput and then stops sending traffic, or maybe
the process that was sending a lot of traffic shifts to another node.
This PR reduces allocation space based not just on current cap, but also
current size to fix this.


internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -237,13 +237,18 @@ func (nb *nodeBatcher) add(spans []*tracepb.Span) {
 
 // cutBatch cuts the provided batch, and sets a new batch on this nodeBatcher
 func (nb *nodeBatcher) cutBatch(b *batch) {
+	initialCap := b.getCurrentCap()
+	currSize := b.getCurrentItemCount()
+	for currSize < initialCap>>1 && initialCap>>1 > initialBatchCapacity {
```
pjanotti: nit: adding a for loop local (`newInitialCap` or `reducedCap` perhaps?) makes easier to see what the loop is doing, and avoids repeating the `>>`  operation

PR #420, opened 2019-02-20
Avoid go build in AppVeyor so bzr is not used

pjanotti:
This tries to work around the issue with AppVeyor by not having a explicit go build.

Opening a PR to test this idea.


pjanotti:
AppVeyor was removed.

PR #423, opened 2019-02-20
Add unified interface for receiver/exporter/spanprocessor.

bogdandrutu:
After this we will replace:
* TraceReceiverSink -> TraceDataProcessor
* MetricsReceiverSink -> MetricsDataProcessor
* MetricsExporter -> MetricsDataProcessor
* TraceExporter -> TraceDataProcessor
* SpanProcessor -> TraceDataProcessor

### DIFF
--- a/processor/processor.go
+++ b/processor/processor.go
@@ -1,0 +1,37 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package processor
+
+import (
+	"context"
+
+	"github.com/census-instrumentation/opencensus-service/data"
+)
+
+// MetricsDataProcessor is an interface that receives data.MetricsData, process it as needed, and
+// sends it to the next processing node if any or to the destination.
+//
+// ProcessMetricsData receives data.MetricsData for processing by the MetricsDataProcessor.
+type MetricsDataProcessor interface {
+	ProcessMetricsData(ctx context.Context, md data.MetricsData) error
+}
+
+// TraceDataProcessor is an interface that receives data.TraceData, process it as needed, and
+// sends it to the next processing node if any or to the destination.
+//
+// ProcessTraceData receives data.TraceData for processing by the TraceDataProcessor.
+type TraceDataProcessor interface {
+	ProcessTraceData(ctx context.Context, td data.TraceData) error
+}
--- a/processor/processortest/noop_processor.go
+++ b/processor/processortest/noop_processor.go
@@ -1,0 +1,45 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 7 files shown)


processor/processor.go
```diff
@@ -0,0 +1,37 @@
...
+type MetricsDataProcessor interface {
+	ProcessMetricsData(ctx context.Context, md data.MetricsData) error
+}
+
+// TraceDataProcessor is an interface that receives data.TraceData, process it as needed, and
+// sends it to the next processing node if any or to the destination.
+//
+// ProcessTraceData receives data.TraceData for processing by the TraceDataProcessor.
+type TraceDataProcessor interface {
+	ProcessTraceData(ctx context.Context, td data.TraceData) error
```
sjkaris: How do you feel about making the proto model the "core" data model for the unified interface? Currently TraceData is an identical structure, and forces us to do an additional mapping.

bogdandrutu: I think it is better to use [Trace|Metrics]Data because in the future we can add extra informations to these like "source" or "num_filtered".

processor/processor.go
```diff
@@ -0,0 +1,37 @@
...
+// MetricsDataProcessor is an interface that receives data.MetricsData, process it as needed, and
+// sends it to the next processing node if any or to the destination.
+//
+// ProcessMetricsData receives data.MetricsData for processing by the MetricsDataProcessor.
+type MetricsDataProcessor interface {
+	ProcessMetricsData(ctx context.Context, md data.MetricsData) error
+}
+
+// TraceDataProcessor is an interface that receives data.TraceData, process it as needed, and
+// sends it to the next processing node if any or to the destination.
```
sjkaris: If it sends it to the next destination, we probably need a `RegisterNext(...)` or something

bogdandrutu: Not really, I was thinking that we will add that to the factory method that creates the Processor. Reason being that you cannot update the next stages at runtime.

sjkaris: I was also thinking we could potentially have this interface that packages need to implement, and then a struct that wraps this interface that deals with the common "core" functionality, such as reconfiguring (reloading) the TraceDataProcessor from its factory, and storing the `name` of the instance.

processor/processor.go
sjkaris: What do you think about adding a `GetTelemetry()` or something so that we can get stats from each processor?

Similarly, a `Flush()` for shutdown.

bogdandrutu: I think it will be great. Actually I was thinking to have utils to do the telemetry. For flash I think it is great. I will add it, do you mind if I do that in the followup PR?

sjkaris: follow up sgtm

processor/processortest/noop_processor.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019, OpenCensus Authors
```
songy23: Why is `noopProcessor ` under `processortest`?

bogdandrutu: Because this is mostly used in tests. It is a test util in my opinion and should never be used in production.

processor/processor.go
pjanotti: Technically it doesn't need to send to anything after it. It makes sense to have this interface as it is to be a consumer in the pipeline and a separate interface to indicate that it sends its output (if any) to others (it could be part of creation but perhaps a separate interface). The example being the final blocks in the pipeline (exporters): they should have these Process(Metrics|Trace)Data but they don't have a next one.

processor/processor.go
pjanotti: yeah, flush is great. Regarding telemetry: perhaps we should add logging, telemetry level at initialization, perhaps something like (comment compiled):
```golang
func InitFromSettings(v *viper.Viper, logger zap.Logger, telemetryLevel telemetry.Level) error
```

I say that because while I think that some utils can  capture some generic cases there will be some metrics/logs specific to some blocks.

bogdandrutu: I would say that this function belongs to the factory interface that we will define.

PR #424, opened 2019-02-21
 Remove Receiver Trace and Metrics sinks.

bogdandrutu:


bogdandrutu:
This PR is a continuation of https://github.com/census-instrumentation/opencensus-service/pull/423

cmd/ocagent/main.go
```diff
@@ -217,19 +208,19 @@ func runOCReceiver(logger *zap.Logger, acfg *config.Config, sr receiver.TraceRec
 
 	switch {
 	case acfg.CanRunOpenCensusTraceReceiver() && acfg.CanRunOpenCensusMetricsReceiver():
-		if err := ocr.Start(ctx, sr, mr); err != nil {
+		if err := ocr.Start(ctx, tdp, mdp); err != nil {
```
pjanotti: not for this PR but the Start methods in general should pass a way to communicate asynchronous errors. These methods most of the time involve launching some go routine but the method signature doesn't allow for these to report errors, so there is code simply ignoring errors inside the launched goroutines, while we probably should be shutting down if the error is not recoverable.

bogdandrutu: I agree, this will be fixed as part of the unification because this code should not live here.

PR #428, opened 2019-02-21
Add Unified factory interfaces

sjkaris:
This changes adds three factory interfaces, one for each type of generic component that can be added. 

`*DataExporterFactory ` creates a new `*DataExporter` which is a thin extension of `*DataProcessor`. By inheriting this interface, we can use an Exporter as a *DataProcessor, while still getting strong-typing guarantees when evaluating whether or not a pipeline is valid (it should end in an exporter). 

`*DataProcessorFactory` creates a new `*DataProcessor` which can be used as an intermediate step in the processing workflow. 

`*DataProcessorFactory` creates a new `*DataProcessor` which can be used as an intermediate step in the trace processing workflow.


exporter/factory.go
```diff
@@ -0,0 +1,67 @@
...
+	// GetExportFormat gets the name of the format in which this exporter sends its data.
+	GetExportFormat() string
+}
+
+// TraceDataExporterFactory is an interface that builds a new TraceDataExporter based on
+// some viper.Viper configuration.
+type TraceDataExporterFactory interface {
+	// GetType gets the type of the TraceDataExporter created by this factory.
+	GetType() string
+	// NewFromViper takes a viper.Viper config and creates a new TraceDataExporter.
```
pjanotti: What about having the two viper related methods that are config related in a separate interface? My reasoning is that we will need to create things from the config that are not exporters, processors, etc. That interface becomes what anything that is created from the config needs to implement.

exporter/factory.go
```diff
@@ -0,0 +1,67 @@
...
+type TraceDataExporter interface {
+	processor.TraceDataProcessor
+
+	// GetExportFormat gets the name of the format in which this exporter sends its data.
+	GetExportFormat() string
+}
+
+// TraceDataExporterFactory is an interface that builds a new TraceDataExporter based on
+// some viper.Viper configuration.
+type TraceDataExporterFactory interface {
```
bogdandrutu: Based on how golang works any factory that implements one of these interface will be able to be used as any other. So probably we need to change the name "NewFromViper" with something link "NewExporterFromViper" and equivalent for all others.

exporter/factory.go
pjanotti: the downside is the interface to be returned... hum...

exporter/factory.go
```diff
@@ -0,0 +1,67 @@
...
+
+// TraceDataExporter inherits functions from TraceDataProcessor, and additionally
+// adds some exporter-specific functions. This helps the collector core to
+// identify which TraceDataProcessors are Exporters and which are internal
+// processing components, so that better validation of pipelines can be done.
+type TraceDataExporter interface {
+	processor.TraceDataProcessor
+
+	// GetExportFormat gets the name of the format in which this exporter sends its data.
+	GetExportFormat() string
```
pjanotti: nit: it is not idiomatic go to put `Get` on getters :)

bogdandrutu: Why do you need this? Is this not the type?

exporter/factory.go
sjkaris: I think since the return type is different, that is not the case here.

processor/factory.go
```diff
@@ -0,0 +1,41 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package processor
+
+import "github.com/spf13/viper"
+
+// TraceDataProcessorFactory is an interface that builds a new TraceDataProcessor based on
+// some viper.Viper configuration.
+type TraceDataProcessorFactory interface {
```
bogdandrutu: I would prefer to have this with a different name. Names that I have in my mind "component", "middleware", "stage", "processingstage"

exporter/factory.go
sjkaris: Yea i prefer the strong typing, I think

exporter/factory.go
sjkaris: It could be -- not sure what is best here exactly. To get the typing, we need an additional function that `TraceDataProcessor` doesnt implement I think.

processor/factory.go
sjkaris: You mean renaming `TraceDataProcessor` -> `TraceProcessingStage`?

exporter/factory.go
pjanotti: at minimum the interface could cover every type thata we can put and connect on the pipeline.

sjkaris: `TraceDataExporter` is an extension of `TraceDataProcessor`, so `TraceDataProcessor` is basically what you are talking about I think. I think keeping the typing like this will allow us to use `TraceDataExporter`s as `TraceDataProcessor`s when its convenient, while still allowing us to know the functional difference that `TraceDataExporter`s are required as the end of pipeline.

processor/factory.go
bogdandrutu: I am fine with the current name, I realized that I was not clear in my thoughts.

processor/factory.go
```diff
@@ -0,0 +1,41 @@
...
+	// GetType gets the type of the TraceDataProcessor created by this factory.
+	GetType() string
+	// NewFromViper takes a viper.Viper config and creates a new TraceDataProcessor.
+	NewFromViper(cfg *viper.Viper) (TraceDataProcessor, error)
```
bogdandrutu: Question:
Why pointer? The config should not be changed, also it is not on the critical path, an extra copy will not have performance implications.

Probably we should have:
```go
NewProcessorFromViper(cfg viper.Viper, next TraceDataProcessor) (TraceDataProcessor, error)
```

Reason to not have a list for next is because this can be called by our config module:
```go
factory.NewProcessorFromViper(cfg, processor.NewMultiTraceDataProcessor(tdps))
```

Also we may provide some helper structs:
```go
type RawTraceDataProcessor {
  ProcessTraceData(ctx contex.Context, td data.TraceData) (context.Context, data.TraceData, error)
}

type ChainedTraceDataProcessor {
  next TraceDataProcessor
  current RawTraceDataProcessor
}

var _ TraceDataProcessor = &ChainedTraceDataProcessor(nil)

func NewChainedTraceDataProcessor(current RawTraceDataProcessor, next TraceDataProcessor) {
  return &ChainedTraceDataProcessor{next: next, current: current}
}

func (ctdp *ChainedTraceDataProcessor) ProcessTraceData(ctx contex.Context, td data.TraceData) error {
  ctx, td, err := current.ProcessTraceData(ctx, td)
  if (err != nil) {
    return err;
  }
  return next.ProcessTraceData(ctx, td)
}
```

Similar to ChainedTraceDataProcessor we can have a ChainedWithObservabilityTraceDataProcessor that also records stats/traces.

receiver/factory.go
```diff
@@ -0,0 +1,43 @@
...
+	GetDefaultConfig() *viper.Viper
+}
+
+// MetricsReceiverFactory is an interface that builds a new MetricsReceiver based on
+// some viper.Viper configuration.
+type MetricsReceiverFactory interface {
+	// GetType gets the type of the MetricsReceiver created by this factory.
+	GetType() string
+	// NewFromViper takes a viper.Viper config and creates a new MetricsReceiver.
+	NewFromViper(cfg *viper.Viper) (MetricsReceiver, error)
```
bogdandrutu: Same, here we need a `next` TraceDataProcessor, and remove it from the Start call in MetricsReceiver.

exporter/factory.go
```diff
@@ -0,0 +1,67 @@
...
+	// GetDefaultConfig returns the default configuration for TraceDataExporter
+	// created by this factory.
+	GetDefaultConfig() *viper.Viper
+}
+
+// MetricsDataExporter inherits functions from MetricsDataProcessor, and additionally
+// adds some exporter-specific functions. This helps the collector core to
+// identify which MetricsDataProcessors are Exporters and which are internal
+// processing components, so that better validation of pipelines can be done.
+type MetricsDataExporter interface {
```
bogdandrutu: Why do we need this at all?

processor/factory.go
sjkaris: Its a pointer since thats how the viper library returns viper instances and generally how it handles itself (e.g. https://godoc.org/github.com/spf13/viper#New). I don't see the benefit of requiring us to dereference the viper (although I agree the performance hit is negligible in this case). 
Agree on the not needing a list b/c of the helpers we will have.

exporter/factory.go
sjkaris: Added it as a dual to the TraceDataExporter. I think that we should have the Metrics and Trace pipelines function in the same way.

sjkaris: Happy to remove this though

bogdandrutu: The same question applies to TraceDataExporter as well.

sjkaris: Are you suggesting that the `Exporters`, since they implement `TraceDataProcessor`, just use that factory? If so, I think that is do-able, but I do kinda like having a separate factory for the separate type. Additionally, I think it will help when people are implementing exporters, since everything that they need to implement is then exporter focused.

sjkaris: With the addition of `next`, this factory makes even more sense, since I think exporters should not have a `next` in their constructor

exporter/factory.go
pjanotti: Let's go your way, it seems better at this stage.

exporter/factory.go
bogdandrutu: I am suggesting that the FromViper returns directly a TraceDataProcessor.

sjkaris: `TraceDataExporter` is a `TraceDataProcessor`, and I also think that we should make sure that `TraceDataExporterFactory` can only create `TraceDataExporter`s, incase someone tries to return just a `TraceDataProcessor`. We could get rid of `TraceDataExporter` altogether, although I kinda like having the distinction for the reasons in the comment

exporter/factory.go
```diff
@@ -0,0 +1,67 @@
...
+package exporter
+
+import (
+	"github.com/spf13/viper"
+
+	"github.com/census-instrumentation/opencensus-service/processor"
+)
+
+// TraceDataExporter inherits functions from TraceDataProcessor, and additionally
+// adds some exporter-specific functions. This helps the collector core to
```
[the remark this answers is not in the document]
sjkaris: Ah agreed -- will update everywhere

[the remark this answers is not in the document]
sjkaris: done.

odeke-em:
Thanks for the PR @sjkaris, I haven't examined this PR much but I'd advise reducing on the "Factory" terminology as that's very Java-esque than Go style. The recommended Go style for interface naming is suffixing `er` or perhaps `MetricsDataExporterCreator` instead of `MetricsDataExporterFactory` https://golang.org/doc/effective_go.html#interface-names

Also we don't have `Inheritance` in Go, but rather `Composition`, so please accordingly adjust comments etc.

sjkaris:
> By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc.

I think the -er suffix is for one-method interfaces that only implement one method. 

Here factory is more referring to the factory design pattern (https://en.wikipedia.org/wiki/Factory_method_pattern) which is being used to abstract away creation of the *DataProcessors. I think the Factory name here helps to indicate how the factory structs will be used. If we think that we don't want to reference this pattern, we can change the name. 

Updated the comments to reflect that we are using composition.

processor/factory.go
pjanotti: We agree that performance is not the criteria here. While in principle the configuration shouldn't be changed that can't be enforced just by passing by value since the type contains pointers. If we want to enforce that we need to define a proper interface in which we select the allowed methods. Passing by value can be confusing because things like maps used by Viper will be shared while side effects of some functions (e.g.: `SetConfigType`) won't. Passing it as a pointer makes clear that any change to it will be shared. My preference is to keep it as a pointer (at least for now).

pjanotti: Clarification regarding performance: I don't expect value vs pointer discussed here to be noticeable but notice that startup time is an important consideration for oc-service in general.

receiver/factory.go
```diff
@@ -0,0 +1,46 @@
...
+package receiver
+
+import (
+	"github.com/census-instrumentation/opencensus-service/processor"
+	"github.com/spf13/viper"
+)
+
+// TraceReceiverFactory is an interface that builds a new TraceReceiver based on
+// some viper.Viper configuration.
+type TraceReceiverFactory interface {
```
pjanotti: Given that these interfaces are the same as the factories for processors do we really need them? It can be useful to test if the pipeline is starting with an actual receiver but that seems to come easily by checking that they support some interface specific of receivers.

pjanotti: Per offline conversation, you convinced me otherwise :)

PR #432, opened 2019-02-22
Add optional net/http/pprof support

pjanotti:
Add a command-line flag to enable net/http/pprof. This is the standard golang profiler and having it available is an easy way to collect cpu/mem profile from users that may face such issues.

Opportunistic fix: add SIGTERM for proper k8s shutdown.

Update demo to expose pprof endpoints (makes an easy setup to investigate performance issues).


cmd/ocagent/cli.go
```diff
@@ -15,17 +15,22 @@
 package main
```
pjanotti: The existence of this file is making `go run` awkward for ocagent, one needs to have both main.go and cli.go on the command... I think I will merge them...

cmd/occollector/app/collector/collector.go
```diff
@@ -93,14 +96,17 @@ func (app *Application) execute() {
 		os.Exit(1)
```
sjkaris: nit: maybe not for this PR, but we should be consistent in our use of `log.Fatalf` vs log followed by os.Exit

internal/pprofserver/pprofserver.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.uber.org/zap"
+)
+
+const (
+	httpPprofPortCfg = "http-pprof-port"
+)
+
+// pport is used to ensure that if AddFlags was called the setup works even if viper
+// was not set.
+var pport *uint
```
sjkaris: I think we should use a `viper` over a global var, even if we have to do some extra work in the ocagent/main.go and ocagent/cli.go

cmd/ocagent/cli.go
sjkaris: Do you have a sense on how much more work it would be to just unify these files with the occollector startup files?

internal/pprofserver/pprofserver.go
pjanotti: I started with that but as you saw it requires more changes to the agent, so I simplified. Let me see the cost of merging cli.go and main.go and fix both things at once.

cmd/ocagent/cli.go
pjanotti: It will be better to unify config first, then it becomes easier.

internal/pprofserver/pprofserver.go
pjanotti: Let's do it separate it starts to pull a lot of other changes, it will be better to separate in a change to move agent to viper.

sjkaris: Can we implement this with viper though and then have the agent do the extra work to pass the viper? I believe it is similarly done for the FromViper that the receivers are built with

sjkaris: that way we wont need to change as well when we go to viper

pjanotti: ok, it won't be pretty but seems reasonable for now

cmd/ocagent/cli.go
pjanotti: deleted cli.go

internal/pprofserver/pprofserver.go
pjanotti: done

Issue #276, opened 2018-12-19
cmd/occollector: baseline performance


flands:
As we get to beta1 we need to begin baselining the occollector. This will be especially important as we introduce tail-based sampling (i.e. need to understand performance impact of tail-based sampling).

PR #436, opened 2019-02-27
Publish initial performance numbers

flands:
Addresses #276


PERFORMANCE.MD
```diff
@@ -0,0 +1,65 @@
...
+has a few configurable parameters which can impact the results of the tests. The
+parameters used are defined below.
+
+* FlushInterval(ms) [default: 1000]
+* MaxQueueSize [default: 100]
+* SubmissionRate(spans/sec): 100,000
+
+## Results without tail-based sampling
+
+| Span<br>Format    | CPU<br>(2+ GHz) | RAM<br>(GB) | Sustained<br>Rate | Recommended<br>Maximum |
```
bogdandrutu: 1. Can we please add numbers with our own format? :)
1. What is the exporter in all these case?

pjanotti: 1. We still need to change the synthetic-load-generator to emit oc (although we could measure oc-agent -> oc-collector)
2. In this case we are interested on ingress rate so we measure with a "null exporter" (i.e.: --debug-processor)

bogdandrutu: We can do the following:

(1)load generator (import_format) -> (2)"test_instance" (export_fromat) -> (3)oc_agent with debug

where:
(1) is the load generator
(2) is the instance we monitor
(3) acts just as a backend (in reality they may have issues with scaling the backend, but this is beyond this project scope).

PERFORMANCE.MD
```diff
@@ -0,0 +1,65 @@
...
+OpenCensus Service, are applicable primarily to the OpenCensus Collector and
+are measured only for traces. In the future, more configurations will be tested.
+
+Note with the OpenCensus Agent you can expect as good if not better performance
+with lower resource utilization. This is because the OpenCensus Agent does not
+today support features such as batching or retries and will not support
+tail-sampling.
+
+It is important to note that the performance of the OpenCensus Collector depends
+on a variety of factors including:
```
songy23: Also mention OS platform here?

PERFORMANCE.MD
```diff
@@ -0,0 +1,65 @@
...
+
+* The receiving format: Jaeger thrift (14268) or Zipkin v2 JSON (9411)
+* The size of the spans (tests are based on number of tags): 20
+* Whether tail-sampling is enabled or not
+* CPU / Memory allocation
+
+## Testing
```
songy23: Maybe also specify which OS you're using when running the tests?

PERFORMANCE.MD
flands: In order to get these numbers published today we will skip, but the goal is to include these going forward (ideally by using the Java ocagent exporter)

flands:
Filed https://github.com/census-instrumentation/opencensus-service/issues/438

Makefile
```diff
@@ -48,7 +48,7 @@ lint:
 
 .PHONY: install-tools
 install-tools:
-	go get -u golang.org/x/lint/golint
```
bogdandrutu: The -u flag instructs get to use the network to update the named packages and their dependencies. By default, get uses the network to check out missing packages but does not use it to look for updates to existing packages.

I don't think removing this is necessary good. Maybe we should file an issue to restore this after lint fixes?

flands: SGTM

pjanotti: Yes, the idea is that this is a temporary workaround.

pjanotti: #439

PERFORMANCE.MD
yurishkuro: Please define what sustained rate means

flands: @yurishkuro done: https://github.com/census-instrumentation/opencensus-service/pull/441

yurishkuro: sorry, still not clear, that change just says "running for a minimum of one hour (i.e. sustained rate)". Rate of what? Number of spans received by the collector? What does it do with the spans besides receiving them?

bogdandrutu:
Hi all,

Can we get a [pprof](https://golang.org/doc/diagnostics.html#profiling) sample while we run the loadtests. I am very interested to see where we have the performance problems - if possible get cpu/block/mutex (all would be great).

flands:
We did not have pprof initially, but when we run the next set of tests we will.

PR #437, opened 2019-02-28
Fix some of the stats exported.

bogdandrutu:
Count -> Sum changes are because count is the aggregation that counts the number of recorded measurements not the sum of the recorded values. In the current case this should be the same.

For process telemetry the TotalAlloc and CPUSeconds are exported as cumulative data so they need to use LastValue, unfortunately they will be exported as gauges even though they are cumulative values (this is unfortunately, but when metrics package will be done in Go we should use that).


bogdandrutu:
@songy23 this is a fix for the current code. I am planning to unify observability.

pjanotti:
@flands is going to piggyback a fix for the build break on Travis, after that is merged yours should pass without issue.

bogdandrutu:
total_memory_alloc and cpu_seconds were broken.

PR #442, opened 2019-02-28
For receivers and exporters record number of received and dropped spans.

bogdandrutu:


internal/observability.go
```diff
@@ -28,80 +28,86 @@ import (
...
-	Description: "The number of spans received by the receiver",
-	Measure:     mReceivedSpans,
-	Aggregation: itemsDistribution,
-	TagKeys:     []tag.Key{tagKeyReceiverName},
+// ViewReceiverReceivedSpans defines the view for the receiver received spans metric.
+var ViewReceiverReceivedSpans = &view.View{
+	Name:        mReceiverReceivedSpans.Name(),
+	Description: mReceiverReceivedSpans.Description(),
+	Measure:     mReceiverReceivedSpans,
+	Aggregation: view.Sum(),
```
songy23: I don't have a strong opinion but maybe histogram is more useful here, since you can also see the rate (mean) of exported spans over time.

bogdandrutu: If we fix the exporters in Go to export cumulative metrics for Sum (by not allowing negative values in Record) rates are available in the backend. The capability to calculate rates is actually the main difference between cumulative vs gauge.

internal/observability.go
```diff
@@ -28,80 +28,86 @@ import (
...
-	tracepb "github.com/census-instrumentation/opencensus-proto/gen-go/trace/v1"
 )
 
 var (
-	tagKeyReceiverName, _ = tag.NewKey("opencensus_receiver")
-	tagKeyExporterName, _ = tag.NewKey("opencensus_exporter")
-)
-
-var mReceivedSpans = stats.Int64("oc.io/receiver/received_spans", "Counts the number of spans received by the receiver", "1")
+	tagKeyTraceReceiverName, _ = tag.NewKey("opencensus_trace_receiver")
```
pjanotti: nit: we already have "oc" on the metric name, and probably on namespace too, perhaps just "receiver", since we counting spans it can't apply to metrics.

internal/observability.go
```diff
@@ -28,80 +28,86 @@ import (
...
-// the number of spans exported.
-func NewExportedSpansRecorder(exporterName string) func(context.Context, *commonpb.Node, []*tracepb.Span) {
-	return func(ctx context.Context, ni *commonpb.Node, spans []*tracepb.Span) {
-		ctx, _ = tag.New(ctx, tag.Upsert(tagKeyExporterName, exporterName))
-		stats.Record(ctx, mExportedSpans.M(int64(len(spans))))
-	}
+// ContextWithTraceExporterName adds the tag "opencensus_trace_exporter" and the name of the
+// receiver as the value, and returns the newly created context.
+func ContextWithTraceExporterName(ctx context.Context, traceReceiverName string) context.Context {
+	ctx, _ = tag.New(ctx, tag.Upsert(tagKeyTraceReceiverName, traceReceiverName))
```
pjanotti: fix tag name and parameter name.

exporter/zipkinexporter/zipkin.go
```diff
@@ -222,8 +222,8 @@ func (ze *zipkinExporter) ProcessTraceData(ctx context.Context, td data.TraceDat
 	}
 
 	// And finally record metrics on the number of exported spans.
-	nSpansCounter := internal.NewExportedSpansRecorder("zipkin")
-	nSpansCounter(ctx, td.Node, goodSpans)
+	ctxWithExporterName := internal.ContextWithTraceExporterName(ctx, "zipkin")
+	internal.RecordTraceExporterMetrics(ctxWithExporterName, len(td.Spans), len(td.Spans)-len(goodSpans))
```
pjanotti: not from your change but having  a slice to count "good spans" is unnecessary, just a counter suffice.

internal/observability.go
```diff
@@ -28,80 +28,86 @@ import (
...
+	TagKeys:     []tag.Key{tagKeyTraceReceiverName},
+}
+
+// ViewExporterReceivedSpans defines the view for the exporter received spans metric.
+var ViewExporterReceivedSpans = &view.View{
+	Name:        mExporterReceivedSpans.Name(),
+	Description: mExporterReceivedSpans.Description(),
+	Measure:     mExporterReceivedSpans,
+	Aggregation: view.Sum(),
+	TagKeys:     []tag.Key{tagKeyTraceReceiverName, tagKeyTraceExporterName},
```
pjanotti: why tagKeyTraceReceiverName is here?

internal/observability.go
```diff
@@ -28,80 +28,86 @@ import (
...
-// ViewExportedSpans defines the view for exported spans metric.
-var ViewExportedSpans = &view.View{
-	Name:        "oc.io/receiver/exported_spans",
-	Description: "Tracks the number of exported spans",
-	Measure:     mExportedSpans,
-	Aggregation: itemsDistribution,
-	TagKeys:     []tag.Key{tagKeyExporterName},
+// ViewExporterDroppedSpans defines the view for the exporter dropped spans metric.
+var ViewExporterDroppedSpans = &view.View{
+	Name:        mExporterReceivedSpans.Name(),
```
pjanotti: cut & paste error: dropped spans for view fields

internal/observability.go
```diff
@@ -28,80 +28,86 @@ import (
...
+	Description: mExporterReceivedSpans.Description(),
+	Measure:     mExporterReceivedSpans,
+	Aggregation: view.Sum(),
+	TagKeys:     []tag.Key{tagKeyTraceReceiverName, tagKeyTraceExporterName},
```
pjanotti: same q as above: why tagKeyTraceReceiverName?

receiver/jaegerreceiver/trace_receiver.go
```diff
@@ -245,7 +249,7 @@ func (jr *jReceiver) SubmitBatches(ctx thrift.Context, batches []*jaeger.Batch)
 			ok = true
 			jr.nextProcessor.ProcessTraceData(ctx, td)
 			// We MUST unconditionally record metrics from this reception.
-			spansMetricsFn(td.Node, td.Spans)
+			internal.RecordTraceReceiverMetrics(ctxWithReceiverName, len(batch.Spans), len(batch.Spans)-len(td.Spans))
```
pjanotti: not from your change, so let's do later but "internal" is not a good name for a package, it is part of its location on the repo but then it should be observability or something else.

exporter/opencensusexporter/opencensus.go
```diff
@@ -116,9 +116,12 @@ func (oce *ocagentExporter) ProcessTraceData(ctx context.Context, td data.TraceD
...
 	)
 	if err != nil {
+		// TODO: If failed to send all maybe record a different metric. Failed to "Sent", but
+		// this may not be accurate if we have retry outside of this exporter. Maybe the retry
+		// processor should record these metrics.
 		return err
 	}
-	nSpansCounter := internal.NewExportedSpansRecorder("ocagent")
-	nSpansCounter(ctx, td.Node, td.Spans)
+	ctxWithExporterName := internal.ContextWithTraceExporterName(ctx, "ocagent")
```
pjanotti: the receiver is "opencensus" I think that it makes more sense to say the same here, it is not about ocagent it is about transport/serialization format.

receiver/jaegerreceiver/trace_receiver.go
pjanotti: also not from your change directly but my thougts is that we should record all spans received and separately the dropped ones. I mean if the translator report an error we don't record those spans as received, and we just take into account the ones that were translated successfully. In practice there should be very translation issues but the metric of received and dropped will be incorrect on that case

exporter/zipkinexporter/zipkin.go
bogdandrutu: done.

bogdandrutu: done.

internal/observability.go
bogdandrutu: Rename it to oc_receiver. We will encode the signal (trace/metrics/etc) in the value, for example "oc_trace", reason is that if we extract metrics from Spans we want to know what was the source of the exported metrics, and it was not "oc" because that may mean the metrics came from "oc_metrics".

internal/observability.go
bogdandrutu: That is the source so you want all these metrics to be broken down by [receiver, exporter].

internal/observability.go
bogdandrutu: done.

internal/observability.go
bogdandrutu: To find more details you can read about "tags". Short story we add the "oc_receiver" tag to the context so all the stats we record against that context will have this tag available, it is exactly what you did by manually propagating the "source" in the internal/collector/processor.

internal/observability.go
bogdandrutu: see previous answer.

internal/observability.go
bogdandrutu: thanks. I will add e2e tests in a separate PR.

receiver/jaegerreceiver/trace_receiver.go
bogdandrutu: Not sure I understand what you mean. Dropped is intended to measure only the spans dropped by this receiver and the only reason is via translation.

If for every step (block, processor, however we call it) we record the number of received and dropped we will have a very good image about what is happening.

exporter/opencensusexporter/opencensus.go
bogdandrutu: Rename it to oc_trace. See comments why we should include the signal in the name (e.g. exporting metrics from traces).

receiver/jaegerreceiver/trace_receiver.go
pjanotti: I mean that when SubmitBatches is called we already received the spans, they should count as received. Right now if `ThriftBatchToOCProto` returns error those spans are not going to be counted at all (nor as received or dropped).

internal/observability.go
pjanotti: I see so if the context flowing correctly we will know the source, nice.

receiver/jaegerreceiver/trace_receiver.go
bogdandrutu: I think I got what you want, but this was an error before.

bogdandrutu: Added TODOs in places where we should also record metrics, also I recorded metrics where data were available.

sjkaris:
LGTM

PR #457, opened 2019-03-02
Add TLS option to opencensus exporter

sjkaris:
Adds the ability to specify the TLS cert in the opencensus exporter configuration to enable TLS communication on gRPC.


sjkaris:
good point, will add

PR #462, opened 2019-03-03
Add more helper observability test functions, enable oc-trace observability tests.

bogdandrutu:
Moved all tests from octrace/observability_test to octrace/opencensus_test


bogdandrutu:
Spent time to understand how the hack coverage went down when I added extra tests. The answer is simple codecov does not track packages without a test file inside, in this case the observability package was not tracked at all and after I added some tests is now tracked, and obviously we don't test for all details there.

I looked at other projects and saw (especially in Jaeger) they enforce every package to have an .nocover file or empty_test.go

bogdandrutu:
@odeke-em I think I restored that file.

observability/obeservability_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/census-instrumentation/opencensus-service/observability"
+	"github.com/census-instrumentation/opencensus-service/observability/observabilitytest"
+)
+
+const (
+	receiverName = "fake_receiver"
+	exporterName = "fake_exporter"
+)
+
+func TestTracePieplineRecordedMetrics(t *testing.T) {
```
songy23: s/Piepline/Pipeline

PR #463, opened 2019-03-03
Change FormatName to [Trace|Metrics]FormatName to be able to use as tags.

bogdandrutu:
I want to hear opinions about this. If we do have different "FormatName" functions then exporters which implement multiple signals can have different name for each signal like "oc_trace" and "oc_metrics".

What do you think?


pjanotti:
It seems reasonable to me. The [Trace|Metrics]Receiver already have distinct interfaces, having the same method name/interface will make hard to deal with exporters that implement trace and metrics.

bogdandrutu:
@pjanotti that was exactly my point, if the same exporter implements both I cannot have different names.

PR #470, opened 2019-03-04
Implement receivers factory

pjanotti:
Implement the receivers factory by adding a template type that can be used by receivers in general. Small changes were made to the factory interface to better cover the usage in the actual code.

Removed the change that was initially done for zipkin-scribe, instead, I will switch all receivers to factory after this PR is merged.

[EDIT]: Removing example with zipking-scribe.

### DIFF
--- a/receiver/factory.go
+++ b/receiver/factory.go
@@ -15,8 +15,9 @@
 package receiver
 
 import (
-	"github.com/census-instrumentation/opencensus-service/consumer"
 	"github.com/spf13/viper"
+
+	"github.com/census-instrumentation/opencensus-service/consumer"
 )
 
 // TraceReceiverFactory is an interface that builds a new TraceReceiver based on
@@ -25,11 +26,11 @@ type TraceReceiverFactory interface {
 	// Type gets the type of the TraceReceiver created by this factory.
 	Type() string
 	// NewFromViper takes a viper.Viper config and creates a new TraceReceiver which uses next as the
-	// next TraceConsumer in the pipeline.
-	NewFromViper(cfg *viper.Viper, next consumer.TraceConsumer) (TraceReceiver, error)
-	// DefaultConfig returns the default configuration for TraceReceivers
+	// next TraceConsumer in the pipeline. Returning also the configuration used to create it.
+	NewFromViper(v *viper.Viper, next consumer.TraceConsumer) (receiver TraceReceiver, config interface{}, err error)
+	// AddDefaultConfig adds the default configuration for the TraceReceivers
 	// created by this factory.
-	DefaultConfig() *viper.Viper
+	AddDefaultConfig(v *viper.Viper)
 }
 
 // MetricsReceiverFactory is an interface that builds a new MetricsReceiver based on
@@ -38,9 +39,9 @@ type MetricsReceiverFactory interface {
 	// Type gets the type of the MetricsReceiver created by this factory.
 	Type() string
 	// NewFromViper takes a viper.Viper config and creates a new MetricsReceiver which uses next as the
-	// next MetricsConsumer in the pipeline.
-	NewFromViper(cfg *viper.Viper, next consumer.MetricsConsumer) (MetricsReceiver, error)
-	// DefaultConfig returns the default configuration for MetricsReceivers
+	// next MetricsConsumer in the pipeline. Returning also the configuration used to create it.
+	NewFromViper(v *viper.Viper, next consumer.MetricsConsumer) (receiver MetricsReceiver, config interface{}, err error)
+	// AddDefaultConfig adds the default configuration for the MetricReceivers
 	// created by this factory.
-	DefaultConfig() *viper.Viper
+	AddDefaultConfig(v *viper.Viper)
 }
--- a/receiver/factory.go
+++ b/receiver/factory.go
@@ -28,9 +28,9 @@ type TraceReceiverFactory interface {
 	// NewFromViper takes a viper.Viper config and creates a new TraceReceiver which uses next as the
 	// next TraceConsumer in the pipeline. Returning also the configuration used to create it.
 	NewFromViper(v *viper.Viper, next consumer.TraceConsumer) (receiver TraceReceiver, config interface{}, err error)
-	// AddDefaultConfig adds the default configuration for the TraceReceivers
+	// DefaultConfig gets the default configuration for the TraceReceiver
 	// created by this factory.
-	AddDefaultConfig(v *viper.Viper)
+	DefaultConfig() interface{}
 }
 
 // MetricsReceiverFactory is an interface that builds a new MetricsReceiver based on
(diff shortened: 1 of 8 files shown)


receiver/factorytemplate/factory_template.go
```diff
@@ -0,0 +1,186 @@
...
+
+// configFromViper takes a viper.Viper, generates a default config and returns the
+// resulting configuration.
+func (f *factory) configFromViper(v *viper.Viper) (cfg interface{}, err error) {
+	if v == nil {
+		return nil, ErrNilViper
+	}
+
+	cfg = f.newDefaultCfg()
+	err = v.UnmarshalKey("receivers."+f.Type(), cfg)
```
bogdandrutu: Not sure how Viper works but was expecting that we pass only the subtree of the config to the factory, and factories do not know about the the whole structure of the tree, I saw `Sub` function that does that, similar for retrieving the default config was expecting that we return the subtree of the config that is specific to this factory.

The core config module will build the entire tree, and parse the entire tree in subtrees. This way we isolate where the structure of the tree lives, currently every receiver/exporter/processor needs to know about the whole tree, which will make impossible to change in the future.

pjanotti: Yes, this is not correct and not needed (assuming caller passing the proper viper sub).

sjkaris: Agreed. We should only pass the subtrees, I believe `ExportersFromViperConfig` does this sort of thing in `internal/config/config.go` as a potential example of the use of the `Sub` that @bogdandrutu mentioned.

bogdandrutu: Also I think the `DefaultConfig() *viper.Viper` is better because we just return the subtree specific for this factory.

pjanotti: I didn't see a direct way to merge the various viper.Viper with different configurations so if we return a new viper for each DefaultConfig the onus is on us to properly merge that information.

pjanotti: A related issue is that our discussed usage of array of items doesn't go straight with v.Sub as we used in other places (or at least with the APIs that I've seen from Viper)

pjanotti:
* [Example of using `AddDefaultConfig`](https://github.com/census-instrumentation/opencensus-service/pull/470/files#diff-e2ebc79ad46047e0f728001e32a81d52R451)
* [Test file](https://github.com/census-instrumentation/opencensus-service/pull/470/files#diff-c27d163d5c51ba35ea81c67910d9bf94R1) showing how config file sections dealing with name and type should look like, [code handling](https://github.com/census-instrumentation/opencensus-service/pull/470/files#diff-e2ebc79ad46047e0f728001e32a81d52R183) that is on the respective test.

receiver/factorytemplate/factory_template.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Type gets the type of the receiver created by this factory.
+func (f *factory) Type() string {
+	return f.receiverType
+}
+
+// NewFromViper takes a viper.Viper configuration and creates a new TraceReceiver.
+// Returning also the configuration used to create it.
+func (trf *traceReceiverFactory) NewFromViper(v *viper.Viper, next consumer.TraceConsumer) (receiver.TraceReceiver, interface{}, error) {
```
bogdandrutu: The name newfromviper may cause a problem for OC which tries to implement both trace/metrics in the same struct. The requirement to be in the same struct comes from the fact that we need to share the grpc server (same addr:port).

How do we plan to handle that?

receiver/factorytemplate/factory_template.go
bogdandrutu: Maybe Viper is not the right tool then. I think from a design point of view the receiver/exporter should not know about the whole structure of our config tree, otherwise will be in a huge problem later because we encode this structure in all the components.

receiver/factorytemplate/factory_template.go
pjanotti: We have 2 different interfaces: TraceReceiverFactory and MetricsReceiverFactory, the NewFromViper on each interface returns a different type, respectively TraceReceiver and MetricsReciver so OC receiver has both factories. 

Receivers that handle both trace and metrics on a single endpoint need some code to handle if the same "server/endpoint" is handling both, e.g.: they need to be a singleton for a given endpoint and internally it should use the same for metrics and traces.

receiver/factorytemplate/factory_template.go
pjanotti: Only the factories needs to deal with viper, receivers/exporters don't need to have anything to deal with viper. That said it is true that the factories are having a side effect on the viper instance. For reading the config I like viper but it is really not made to output a config file like we want to do for default, one alternative is for us to marshal the configs directly.

receiver/factory.go
```diff
@@ -25,11 +26,11 @@ type TraceReceiverFactory interface {
 	// Type gets the type of the TraceReceiver created by this factory.
 	Type() string
 	// NewFromViper takes a viper.Viper config and creates a new TraceReceiver which uses next as the
-	// next TraceConsumer in the pipeline.
-	NewFromViper(cfg *viper.Viper, next consumer.TraceConsumer) (TraceReceiver, error)
-	// DefaultConfig returns the default configuration for TraceReceivers
+	// next TraceConsumer in the pipeline. Returning also the configuration used to create it.
+	NewFromViper(v *viper.Viper, next consumer.TraceConsumer) (receiver TraceReceiver, config interface{}, err error)
```
bogdandrutu: Why do you need to return the configuration used?

receiver/factory.go
```diff
@@ -25,11 +26,11 @@ type TraceReceiverFactory interface {
...
+	// DefaultConfig gets the default configuration for the TraceReceiver
 	// created by this factory.
-	DefaultConfig() *viper.Viper
+	DefaultConfig() interface{}
```
bogdandrutu: Do we need a ValidateConfig(v *viper.Viper) error? That we can use in the config module to validate a config?

receiver/factory.go
pjanotti: for logging

receiver/factory.go
pjanotti: If we want to run "validate" config by itself then yes, but, the ultimate validation is to create the receivers. We don't necessarily need to start them if we just want to validate the config: build the pipelines and terminate without starting any reception. Depending on how receivers are built we may miss some cases that require starting reception to error out, but, that is also not solved by a ValidateConfig method.

BTW, I should change from `Unmarshal` to `UnmarshalExact` so viper errors on exta-fields (typically these are typos).

receiver/factory.go
bogdandrutu: Logging is another topic of unification, we use different logging systems (at least I saw that in some new PRs).

Can we instead get a logger as argument?

bogdandrutu:
@pjanotti also please make sure the coverage is 100% "The diff coverage is 96.66%", you did not fix with a wrong config.

receiver/factory.go
pjanotti: will do

receiver/factorytemplate/factory_template.go
```diff
@@ -0,0 +1,197 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package factorytemplate allows easy construction of factories for receivers.
+// It requires the consumer to provide only functions to create the default
+// configuration and how to create the receiver from that configuration.
+package factorytemplate
```
bogdandrutu: nit: Let's may a convention that the main file in a package should have the same name, in this case factorytemplate.go

bogdandrutu: *make a convention

pjanotti: ok, renaming it.

PR #471, opened 2019-03-04
Move agent main.go to use viper instead of yaml.v2

sjkaris:
We need to consolidate the agent and collector configurations. This unifies the two components to both use viper, which is the first step towards further unification. 

This should also fix some issues, such as TLS configs not working for the collector service due to mixed tagging (both yaml and mapstructure).


cmd/ocagent/main.go
```diff
@@ -80,19 +79,22 @@ func main() {
 }
 
 func runOCAgent() {
-	yamlBlob, err := ioutil.ReadFile(configYAMLFile)
+	viperCfg.SetConfigFile(configYAMLFile)
```
odeke-em: Where did `viperCfg` come from? Or where is it declared?

cmd/ocagent/main.go
```diff
@@ -138,6 +134,7 @@ func runOCAgent() {
 		closeFns = append(closeFns, zCloseFn)
 	}
 
+	// TODO: generify the startup of these receivers when unifying them w/ collector
```
odeke-em: s/generify/Generalize/g

receiver/prometheusreceiver/metrics_receiver.go
```diff
@@ -45,11 +49,29 @@ type Preceiver struct {
...
+	var cfg Configuration
+
+	// Unmarshal our config values (using viper's mapstructure)
+	err := v.Unmarshal(&cfg)
+	if err != nil {
+		return nil, fmt.Errorf("prometheus receiver failed to parse config: %s", err)
+	}
+
+	// Unmarshal prometheus's config values. Since prometheus uses `yaml` tags, so use `yaml`.
+	promCfgMap := v.Sub("config").AllSettings()
```
odeke-em: I believe in the /internal helpers we have something to parse YAML from *viper.Viper

cmd/ocagent/main.go
sjkaris: line 58

odeke-em: Cool, thanks!

receiver/prometheusreceiver/metrics_receiver.go
sjkaris: This is a little different. The utility is to read in a yaml file (represented as bytes) into a viper struct. Here, we already have a viper struct, but instead are marshalling the config values here into a prometheus library defined struct that is tagged with `yaml` tags. `viper` won't marshal into `yaml` tags (as it understands `mapstructure` tags instead), so here we use yaml.v2 to do the final marshalling.

PR #476, opened 2019-03-06
List required tools to build from sources

pjanotti:
Adds the list of required tools to build from sources and so other minor changes to CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,6 +55,8 @@ $ make install-tools # Only first time.
 $ make
 ```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -23,7 +23,17 @@ information on using pull requests.
 
 [GitHub Help]: https://help.github.com/articles/about-pull-requests/
 
-## Instructions
+## Required Tools
+
+Working with the project sources requires the following tools:
+
+1. [git](https://git-scm.com/)
```
bogdandrutu: One think that you may want to add is that some of the tools we use during the build like golint are installed into the `$GOPATH/bin` directory, so one required thing may be to do is `PATH=$GOPATH/bin:$PATH`

pjanotti: I will add a note to that effect

PR #477, opened 2019-03-06
Make all receivers take "next" on creation

pjanotti:
This is required in order to move the code to use factories for the receivers. The "next" parameter will be passed to the factory and then to the receiver constructor, this way Start[Trace|Metrics]Reception doesn't need the next parameter anymore. Since typically the Start* methods launch goroutines took the opportunity to add a channel to report asynchronous errors.

Opportunistic change: renamed WrapWithSpanSink and related variables.


pjanotti:
After this is merged I will do a follow-up PR to ensure that all receivers are using the asynErrorChan in the goroutines that they launch.

pjanotti:
Interesting I will need to pay a code coverage tax :)

receiver/receiver.go
```diff
@@ -52,7 +54,10 @@ type MetricsReceiver interface {
 	MetricsSource() string
 
 	// StartMetricsReception tells the receiver to start its processing.
-	StartMetricsReception(ctx context.Context, nextConsumer consumer.MetricsConsumer) error
+	// Any fatal errors happening in goroutines started by the receiver should be
+	// reported on asyncErrorChannel.
+	// By convention the consumer of the data received is set at creation time.
+	StartMetricsReception(ctx context.Context, asyncErrorChannel chan<- error) error
```
bogdandrutu: Not sure how this async error channel works. Can you please explain? I am pretty new in go, but why do we need a channel to listen to errors? also why is that not passed in the ctor?

pjanotti: (full disclaimer: I am also q relative new comer to golang almost 5 months :))

These Start*Reception methods most of the time are going to launch some goroutine that has a blocking call that only returns when the receiver is stopped. Right now errors returned by these functions can call something like `log.Fatal` and terminate the app 
(rude shutdown), just log something, or as some receivers are doing right now just ignore the return values. The 2 last options keep the app running while the receiver is broken. This `asynErrorChan` allows these goroutines to notify if they encounter an error, allowing us to log it and attempt a graceful shutdown.

The pattern goes like this:
https://github.com/census-instrumentation/opencensus-service/blob/bad985a1a534657c421252352ef7154477d5d740/cmd/occollector/app/collector/collector.go#L106-L114

It seems more natural to pass it via the interface since we want the receivers to only get data after we call Start*Reception. Unlike the `next` for the factories the receivers already should error when they are started more than once, so there is no way to change the channel after start.

With all of that said this is a simple pattern that came easily with goroutines and channels but perhaps there are better patterns for this type of scenario I just didn't find them.

bogdandrutu: Question about receiver:
Do we allow users to call Stop then Start again? If yes then we need to protect this channel variable or use atomic reference load/store.

Question about processor/exporter:
Do we want to have this channel for them as well?

Also some random thoughts:
When you have a fatal error usually it is very hard to recover, and probably almost impossible to graceful shutdown. I would like to understand what is a fatal error that you think I should send via this channel?

pjanotti: > Do we allow users to call Stop then Start again? If yes then we need to protect this channel variable or use atomic reference load/store.

The `once` protection around start/stop should be enough for that, but, that said I am not convinced that we should support Start/Stop multiple times. We do want some kind of shutdown to flush ensure stop receiving while flushing the pipes.

> Do we want to have this channel for them as well?

At this point no, the working of a processor doesn't depend (at least so far) of some blocking call running in the background. So it doesn't seem that it needs this kind of interruption.

> fatal error

A fatal error in this context is like a call to `ServeHTTP` returning before you close the server. Guaranteed it seems an unlikely thing but on the other hand if it does happens in a goroutine that is just ignoring such errors the process is going to continuer without that endpoint working. Some examples on the code:

https://github.com/census-instrumentation/opencensus-service/blob/bad985a1a534657c421252352ef7154477d5d740/receiver/zipkinreceiver/trace_receiver.go#L102-L105

or to avoid waits like this one:
https://github.com/census-instrumentation/opencensus-service/blob/62b0ae5be7ca9ebc3f372c09686124be786e6dc9/receiver/opencensusreceiver/opencensus.go#L272-L282

pjanotti: @bogdandrutu do you want to pursue this further? in lack of an immediate alternative I think we should keep the channel for now, if we find a better pattern then we refactor it.

pjanotti: another thing that we may want to consider is to remove the Stop method from the interface and actually return a stop function on the start call: it makes clear the pairing and requires one to have called start to have a stop available while also reducing the interface.

bogdandrutu:
Happy to make progress here. let's ship it.

bogdandrutu:
I will approve after rebase :)

pjanotti:
*ei immergitur* according to google translator :)

bogdandrutu:
Any reason not to merge this?

pjanotti:
Just had been busy with other stuff... merging after another rebase.

PR #481, opened 2019-03-07
receiver/prometheus: migrate code in here

### LINKED ISSUE #480: receiver/prometheus: bring in source code from Orijtech, Inc.
Following suite with #478, this issue is meant to bring in code for the Prometheus receiver that I implemented and parked at Orijtech, Inc's https://github.com/orijtech/promreceiver because of issues and churn creating repositories under census-ecosystem. I'll bring it in here and then later it's final home can be decided.

odeke-em:
Hand over prometheus exporter code that I had
parked in https://github.com/orijtech/promreceiver

That code was written for OpenCensus Authors, and was just
waiting for a final destination. This change finishes that.

Fixes #480


pjanotti:
Hi @odeke-em, I'm closing this PR since it is stale for some time and iirc was superseded by other (later closed), let's know if that is not correct.

PR #484, opened 2019-03-09
Add exporterhelper to wrap all exporters with observability.

bogdandrutu:
Only Logging and OpenCensus exporters changed in this PR, I will update all the others in a followup PR.

### DIFF
--- a/exporter/exporterhelper/common.go
+++ b/exporter/exporterhelper/common.go
@@ -1,0 +1,66 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exporterhelper
+
+import (
+	"go.opencensus.io/trace"
+)
+
+var (
+	okStatus = trace.Status{Code: trace.StatusCodeOK}
+)
+
+// ExporterOptions contains options concerning how an Exporter is configured.
+type ExporterOptions struct {
+	// TOOD: Retry logic must be in the same place as metrics recording because
+	// if a request is retried we should not record metrics otherwise number of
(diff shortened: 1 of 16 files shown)


exporter/exporterhelper/common.go
```diff
@@ -0,0 +1,66 @@
...
+	var opts ExporterOptions
+	for _, op := range options {
+		op(&opts)
+	}
+	return opts
+}
+
+func errToStatus(err error) trace.Status {
+	if err != nil {
+		return trace.Status{Code: trace.StatusCodeUnknown, Message: err.Error()}
```
songy23: Should we check if `err` is an RPC error and set precise status here?

exporter/exporterhelper/metricshelper.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/census-instrumentation/opencensus-service/observability"
+	"go.opencensus.io/trace"
+
+	"github.com/census-instrumentation/opencensus-service/data"
+	"github.com/census-instrumentation/opencensus-service/exporter"
+)
+
+const (
+	numDroppedMetricsAttribute  = "num_dropped_metrics"
```
songy23: Consider putting these in `constants.go`?

exporter/exporterhelper/tracehelper.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"go.opencensus.io/trace"
+
+	"github.com/census-instrumentation/opencensus-service/data"
+	"github.com/census-instrumentation/opencensus-service/exporter"
+	"github.com/census-instrumentation/opencensus-service/observability"
+)
+
+const (
+	numDroppedSpansAttribute  = "num_dropped_spans"
```
songy23: Same here.

exporter/exporterhelper/common.go
bogdandrutu: Added a TODO for the moment.

exporter/exporterhelper/common.go
```diff
@@ -0,0 +1,66 @@
...
+// Construct the ExporterOptions from multiple ExporterOption.
+func newExporterOptions(options ...ExporterOption) ExporterOptions {
+	var opts ExporterOptions
+	for _, op := range options {
+		op(&opts)
+	}
+	return opts
+}
+
+func errToStatus(err error) trace.Status {
```
pjanotti: except for this function and the related var it seems that a better name will be options.go, then the errToStatus and the var perhaps can be moved to other file

bogdandrutu: Will do that in a different PR. Thanks for the suggestion.

PR #485, opened 2019-03-11
Change SpanProcessor to use Context.

bogdandrutu:
Caveats:
- Lose some source format in logs;
- Some recorded metrics may not include the source, but I left TODOs to fix that.

### DIFF
--- a/internal/collector/processor/nodebatcher/node_batcher.go
+++ b/internal/collector/processor/nodebatcher/node_batcher.go
@@ -107,14 +107,14 @@ func NewBatcher(name string, logger *zap.Logger, sender processor.SpanProcessor,
 
 // ProcessSpans implements batcher as a SpanProcessor and takes the provided spans and adds them to
 // batches
-func (b *batcher) ProcessSpans(td data.TraceData, spanFormat string) error {
-	bucketID := b.genBucketID(td.Node, td.Resource, spanFormat)
-	bucket := b.getOrAddBucket(bucketID, td.Node, td.Resource, spanFormat)
-	bucket.add(td.Spans)
+func (b *batcher) ProcessSpans(ctx context.Context, td data.TraceData) error {
+	bucketID := b.genBucketID(td.Node, td.Resource)
+	bucket := b.getOrAddBucket(bucketID, td.Node, td.Resource)
+	bucket.add(ctx, td.Spans)
 	return nil
 }
 
-func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource, spanFormat string) string {
+func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource) string {
 	h := md5.New()
 	if node != nil {
 		nodeKey, err := proto.Marshal(node)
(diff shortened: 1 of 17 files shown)


internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -172,7 +171,6 @@ type nodeBatcher struct {
 	currBatch     *batch
 	node          *commonpb.Node
 	resource      *resourcepb.Resource // TODO(skaris) remove when resource is added to span
-	spanFormat    string
```
songy23: Why remove this?

internal/collector/processor/tailsampling/processor.go
```diff
@@ -189,7 +192,7 @@ func (tsp *tailSamplingSpanProcessor) ProcessSpans(td data.TraceData, spanFormat
 	idToSpans := make(map[traceKey][]*tracepb.Span)
 	for _, span := range td.Spans {
 		if len(span.TraceId) != 16 {
-			tsp.logger.Warn("Span without valid TraceId", zap.String("spanFormat", spanFormat))
+			tsp.logger.Warn("Span without valid TraceId")
```
pjanotti: That information should be on the ctx, it seems good information to be on the warning message.

internal/collector/processor/nodebatcher/node_batcher.go
pjanotti: Why not a reference to context in the place of the field?

bogdandrutu:
@songy23 @pjanotti I think I will re-do this PR with a different change because:
1. No way to access the source name from ctx to log (we actually work on the ability to add tags to the log messages but not available yet).
1. Avoid changing the collector stats for the moment.

Proposed alternative:
1. Add source format to the TraceData for the moment, with a comment that this will be removed in the near feature after the unification on metrics/stats.
2. Change the SpanProcessor to use (Context, TraceData) and don't change any logic around format name because this will still be available.

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -421,6 +415,7 @@ func (b *batch) add(spans []*tracepb.Span) (cut bool, closed bool) {
 		b.grow(openTill)
 	}
 
+	// Why not: copy(b.items.Load().([]*tracepb.Span)[openFrom:openTill], spans)?
```
sjkaris: @bogdandrutu we should just supplant this all this logic with something simpler as you said. https://github.com/census-instrumentation/opencensus-service/pull/487 replaces this, addressing the PR comments there now

bogdandrutu: As I mentioned in my previous comment I feel like I am changing too much in this PR and I want to go with that proposal of still passing the source name for the moment in the TraceData then remove it later from there.

bogdandrutu:
I would like to get an agreement on what you think about the latest proposal of not removing the source name for the moment and do that later, as I mentioned in the comment https://github.com/census-instrumentation/opencensus-service/pull/485#issuecomment-471740800

songy23:
> Add source format to the TraceData for the moment, with a comment that this will be removed in the near feature after the unification on metrics/stats.
Change the SpanProcessor to use (Context, TraceData) and don't change any logic around format name because this will still be available.

+1 to this. For now let's leave the span format name.

pjanotti:
I agree with the proposal, adding to td seems a reasonable compromise.

bogdandrutu:
Closing this in favor of https://github.com/census-instrumentation/opencensus-service/pull/489

PR #486, opened 2019-03-11
Fix test checking idToTrace map used in tail sampling

pjanotti:
The test seems to have been running in sequence on most (all?) of Travis runs: it makes assumptions about which traces will be present on idToTrace map and that is not possible when traces are added concurrently since there are no guarantees about the insertion order.

Fixed the concurrent test and added a sequential test in which the ids that are expected to have been dropped can be actually validated.


pjanotti:
node_batcher tests have some concurrent condition that affects its coverage numbers, this PR doesn't affect any of that code.

pjanotti:
Thanks for pointing this out and for the review @bogdandrutu

PR #487, opened 2019-03-11
Fix nodebatcher to be simpler and more efficient

sjkaris:
As @bogdandrutu has brought up, the current node batcher implementation is both overly complex and not very efficient. This simplifies the code by just using a simple mutex.


internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -135,154 +130,83 @@ func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource
...
 }
 
 func (b *batcher) getOrAddBucket(
 	bucketID string, node *commonpb.Node, resource *resourcepb.Resource, spanFormat string,
-) *nodeBatcher {
+) *nodeBatch {
 	bucket, alreadyStored := b.buckets.LoadOrStore(
 		bucketID,
-		newNodeBucket(b, node, resource, spanFormat, b.sendBatchSize, b.timeout, b.logger),
+		newNodeBatch(b, spanFormat, node, resource),
```
bogdandrutu: I would optimize this for the "common" case, which should be the fact that we find the batch:

```go
bucket, loaded := b.buckets.Load(bucketID)
if !loaded {
  // slow path
  bucket, loaded = b.buckets.LoadOrStore(bucketID, newNodeBatch(b, spanFormat, node, resource))
  if !loaded {
     ...
  }
}
return bucket.(*nodeBatch)
```

Up to you, but currently we always call newNodeBatch which seems inefficient.

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -135,154 +130,83 @@ func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource
...
-		} else {
-			stats.RecordWithTags(context.Background(), statsTags, statBatchOnDeadNode.M(1))
-		}
-		// Shoot off event so that we don't block this add on other adds
-		go nb.cutBatch(b)
-	}
-}
+func (nb *nodeBatch) add(spans []*tracepb.Span) {
+	nb.mu.Lock()
+	nb.items = append(nb.items, spans...)
```
bogdandrutu: spans []*tracepb.Span should be immutable so why not make `items [][]*tracepb.Span` then this will become `nb.items = append(nb.items, spans[:len(spans)])` or just `nb.items = append(nb.items, spans)`

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -135,154 +130,83 @@ func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource
...
-}
+	nb.mu.Unlock()
 
-func (nb *nodeBatcher) sendBatch(batch *batch) {
-	spans := batch.getSpans()
-	statsTags := processor.StatsTagsForBatch(
-		nb.parent.name, processor.ServiceNameForNode(nb.node), nb.spanFormat,
-	)
-	stats.RecordWithTags(context.Background(), statsTags, statBatchSize.M(int64(len(spans))))
+	if len(itemsToProcess) > 0 {
```
bogdandrutu: In order to not penalize on goroutine for this probably we should put this into a channel and another goroutine should pick this up. Maybe for the moment a TODO and a followup PR is fine.

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -332,37 +256,39 @@ func (bt *bucketTicker) start() {
...
-							// and try to cut a final batch (just in case something was added in between
-							// the above if and marking this nodeBatcher as dead)
-							atomic.StoreUint32(&nb.dead, nodeStatusDead)
-							// Reload batch in case `add` loaded a new batch that was then added to.
-							currBatchPtr := (*unsafe.Pointer)(unsafe.Pointer(&nb.currBatch))
-							b = (*batch)(atomic.LoadPointer(currBatchPtr))
-							if b.getCurrentItemCount() > 0 {
-								nb.cutBatch(b)
+						nb.mu.Lock()
+						if len(nb.items) > 0 {
```
bogdandrutu: Can we have some different functions for this part since the indentation is going crazy here and is very hard to read?

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: agreed that what you mentioned is better

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: I think we would end up having to loop through to make a single list at the end to pass along since that is what `data.TraceData` requires. Idk if you know a good/more efficient way to collapse these lists

bogdandrutu: Indeed you need to do that but you don't do that while holding the lock. So you avoid contention.

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: added a todo.

internal/collector/processor/nodebatcher/node_batcher.go
bogdandrutu: I saw the same code in tailbase probably fix that code in a different PR.

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: done

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: ah, makes sense, will do

bogdandrutu: Are you planning to do this?

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -320,159 +261,67 @@ func (bt *bucketTicker) add(bucketID string) {
...
+			nb.lastSent = time.Now().UnixNano()
+			nb.totalItemCount = 0
+		}
+		nb.mu.Unlock()
 
-func (b *batch) getCurrentCap() uint32 {
-	return atomic.LoadUint32(&b.currCap)
-}
+		if len(itemsToProcess) > 0 {
+			statsTags := processor.StatsTagsForBatch(
```
bogdandrutu: Stats recording should be part of the sendItems probably because you want also to record this when sendItmes is called from `(nb *nodeBatch) add`

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: Yes, it should be done in the most recent code

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -320,159 +261,67 @@ func (bt *bucketTicker) add(bucketID string) {
...
-	atomic.StoreUint32(&b.currCap, newCap)
-}
+func (bt *bucketTicker) processNodeBatch(nbKey string, nb *nodeBatch) {
+	nb.mu.Lock()
+	if nb.totalItemCount > 0 {
+		// If the batch is non-empty, go ahead and send it
+		itemCount := nb.totalItemCount
+		var itemsToProcess [][]*tracepb.Span
+		if nb.lastSent+bt.parent.timeout.Nanoseconds() < time.Now().UnixNano() {
+			itemsToProcess = nb.items
```
bogdandrutu: This seems like a good place to avoid duplicate code between this function and `func (nb *nodeBatch) add`

```go
func (nb *nodeBatch) getAndReset() ([][]*tracepb.Span, int){
  itemsToProcess := nb.items
  itemsCount := nb.totalItemCount
  nb.items = make([][]*tracepb.Span, 0, initialBatchCapacity)
  nb.lastSent = time.Now().UnixNano()
  nb.totalItemCount = 0
  return itemsToProcess, itemsCount
}
```

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -320,159 +277,59 @@ func (bt *bucketTicker) add(bucketID string) {
...
+		}
+		nb.mu.Unlock()
+	}
 }
 
-func (b *batch) getSpans() []*tracepb.Span {
-	// This shouldn't copy the backing array, just create a new slice with the correct
-	// length
-	return b.items.Load().([]*tracepb.Span)[:b.getCurrentItemCount()]
+func (bt *bucketTicker) stop() {
```
bogdandrutu: Test this function?

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -135,155 +130,117 @@ func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource
...
-		)
+		Spans:    tdItems,
 	}
+
+	statsTags := processor.StatsTagsForBatch(
+		nb.parent.name, processor.ServiceNameForNode(nb.node), nb.format,
+	)
+	_ = stats.RecordWithTags(context.Background(), statsTags, measure.M(1))
+
+	// TODO: This process should be done in an async way, perhaps with a channel + goroutine worker(s)
```
pjanotti: I think the batcher doesn't need to worry about this: the next element on the pipeline takes care of that if desired.

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -135,155 +130,117 @@ func (b *batcher) genBucketID(node *commonpb.Node, resource *resourcepb.Resource
...
+	_ = nb.parent.sender.ProcessSpans(td, nb.format)
+}
+
+func (nb *nodeBatch) getAndReset() ([][]*tracepb.Span, uint32) {
+	itemsToProcess := nb.items
+	itemsCount := nb.totalItemCount
+	nb.items = make([][]*tracepb.Span, 0, initialBatchCapacity)
```
pjanotti: why not `len(itemsToProcess)`? if some entries goes unused then next cycle the initial capacity is reduced

internal/collector/processor/nodebatcher/node_batcher.go
pjanotti: I'm not so sure: perhaps we delegate that to the next element in the pipeline: likely a queue that is already putting stuff via channel.

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -320,159 +277,59 @@ func (bt *bucketTicker) add(bucketID string) {
...
+		select {
+		case <-bt.ticker.C:
+			for nbKey := range bt.nodes {
+				nb := bt.parent.getBucket(nbKey)
+				// Need to check nil here incase the node was deleted from the parent batcher, but
+				// not deleted from bt.nodes
+				if nb == nil {
+					// re-delete just in case
+					delete(bt.nodes, nbKey)
+					continue
```
pjanotti: nit: since you have continue you don't need an `else`

internal/collector/processor/nodebatcher/node_batcher.go
pjanotti: yes, in the beginning of tail sampling there was a type with a similar LoadOrStore that took a factory function to be called when it was being stored, when that changed to sync.Map the behavior was missed (changed to sync.Map since the perf was better).

internal/collector/processor/nodebatcher/node_batcher.go
```diff
@@ -37,9 +34,8 @@ import (
 )
 
 const (
-	initialBatchCapacity     = uint32(512)
+	initialBatchCapacity     = uint32(16)
```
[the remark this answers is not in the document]
sjkaris: Switching the way that the batches are stored from a flat list to a list of lists. Still mostly just a guess here, its equivalent if there are 32 spans a message on ingress.

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: This measures whether the batcher is triggering a send due to a timeout or a sendSize, which can be useful when tuning how big the sendSize and timeout configurations should be

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: done

internal/collector/processor/nodebatcher/node_batcher.go
sjkaris: removed continue, since i think having the else is more clear

internal/collector/processor/nodebatcher/node_batcher.go
pjanotti: I agree, with the metrics my point is about the TODO. We can discuss that when we either remove the TODO or implement it.

sjkaris: replied to wrong line.... yea, we can see about what makes the most sense with this, it might be that we just have the other stages return back quickly and do the queuing on that side

PR #433, opened 2019-02-24
Jaeger grpc exporter

annanay25:
Resolves #279 

Added oc-proto to jaeger-proto translation, and jaeger-grpc sender.

PR #496, opened 2019-03-13
Final CLA test

flands:


songy23:
Hi @flands, unfortunately CLA will always fail if one PR contains commits from multiple authors (I believe it's the case for https://github.com/census-instrumentation/opencensus-service/pull/433). Even if all the authors have signed CLA, it will still fail. This is a known issue to us and we can override it so don't worry about this.

Issue #492, opened 2019-03-13
receiver/opencensus/tests: Fix TestGrpcGateway_endToEnd.


songy23:
TestGrpcGateway_endToEnd failed after switching to opencensus-proto v0.2.0. I suspect this is due to changes in method descriptor (similar to checksum). I disabled this test in https://github.com/census-instrumentation/opencensus-service/pull/490/commits/b62856de8f75d2b9c429090e605f3483188442a3#diff-ef93d12fd0c92defff8fe7dbacf0f2e8R39. File this issue to do some investigation on the test failure and re-enable it.

PR #501, opened 2019-03-26
Make agent grpc-gateway work with grpc 1.18+

draffensperger:
The `google.golang.org/grpc` v1.18.0 release includes a [change that made handshakes required by default](https://github.com/grpc/grpc-go/pull/2565). That when combined with the `cmux` multiplexer caused an "all SubConns are in TransientFailure" error message for HTTP calls made via grpc-gateway (see this [issue comment](https://github.com/grpc/grpc-go/issues/2636#issuecomment-472199078) for more details). The fix is to use a [setting that cmux offers](https://github.com/soheilhy/cmux#limitations) that handles the handshake correctly.

Our grpc 1.18 upgrade was rolled back in https://github.com/census-instrumentation/opencensus-service/pull/503 so this is not immediately needed, but having it in place will make the grpc-gateway functionality continue to work once the grpc dependency is upgraded past 1.18.

See https://github.com/census-instrumentation/opencensus-service/issues/492 - this will enable the grpc-gateway test to keep passing (and the functionality to keep working) after the future grpc upgrade.


draffensperger:
cc/ @songy23

pjanotti:
@draffensperger we reverted grpc to 1.17.0 since there were other things broken. It looks like that we may skip v1.18.0 will this PR still be needed in this case?

songy23:
I'd suggest to put this PR on hold and merge it once we roll forward the proto (and gRPC) updates.

draffensperger:
The main thing I would like is for the agent grpc-gateway functionality to continue to work as future changes are made, since I'm using that for the [opencensus-web](https://github.com/census-instrumentation/opencensus-web) project to write spans via HTTP.

I'm open to waiting until the grpc dependency is updated but then I'd like this change to be done in the same PR as the upgrade to preserve the functionality.

Given that, I actually think it would be simpler to just merge this now so that there will be one less thing to think about when upgrading to grpc 1.18 or 1.19.

pjanotti:
Makes sense @draffensperger.

PR #490, opened 2019-03-13
Upgrade opencensus-proto version to v0.2.0.

songy23:

PR #503, opened 2019-03-26
Revert "Upgrade opencensus-proto version to v0.2.0. (#490)"

### LINKED ISSUE #502: receiver/opencensus: fail to receive spans/metrics over RPC.

pjanotti:
This reverts commit 392ee4a2fd1b1ee95e2c8aee70a39a5dc091b03c.

Fixes #502


odeke-em:
Oh oh, what happened?

pjanotti:
@odeke-em I'm not sure but the RPC between agent and collector in the demo/traces is broken with the upgrade, reverting it fixes the issue.

pjanotti:
Using https://github.com/grpc/grpc-go/releases/tag/v1.18.1 also fixes demo/trace perhaps will go with that instead of the revert.

songy23:
> Using https://github.com/grpc/grpc-go/releases/tag/v1.18.1 also fixes demo/trace perhaps will go with that instead of the revert.

Great to know, so the issue may be the RPC client and server generated with protoc v3.6.1 don't work with this gRPC version.

pjanotti:
Per conversation on issue #502 going with the revert.

PR #511, opened 2019-04-03
Upgrade OC versions in go.mod

songy23:
OC-Go 0.20.0 is released. Upgrade OC versions and run `GO111MODULE=on go mod tidy`.


songy23:
The build failure is because `github.com/uber/tchannel-go`, `github.com/jaegertracing/jaeger` and `github.com/omnition/scribe-go` don't work with the latest version of Thrift (IIUC they vendor a different version of Thrift).

songy23:
Tried to use a lower version of Thrift:
```bash
$ GO111MODULE=on go get github.com/apache/thrift@v0.0.0-20161221203622-b2a4d4ae21c7
```
However that forced us to use a lower version of OC as well (see https://github.com/census-instrumentation/opencensus-service/pull/511/commits/dd856148009a9be78ab82507a18461dbbee886a0). So it looks like the transitive dependency conflict between OC-Go and the other 3 libraries is incompatible. I'm going to close this PR for now and explore other options.

PR #512, opened 2019-04-03
Upgrade OC version to 0.19.3

songy23:
Cannot upgrade to 0.20.0 because of transitive dependency conflicts.


go.mod
```diff
@@ -41,18 +42,16 @@ require (
...
+	github.com/openzipkin/zipkin-go v0.1.6
 	github.com/orijtech/prometheus-go-metrics-exporter v0.0.3-0.20190313163149-b321c5297f60
 	github.com/orijtech/promreceiver v0.0.5
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/philhofer/fwd v1.0.0 // indirect
 	github.com/pkg/errors v0.8.0
 	github.com/prashantv/protectmem v0.0.0-20171002184600-e20412882b3a // indirect
-	github.com/prometheus/client_golang v0.9.2
-	github.com/prometheus/procfs v0.0.0-20181204211112-1dc9a6cbc91a
+	github.com/prometheus/client_golang v0.9.3-0.20190127221311-3c4408c8b829
```
odeke-em: I'd stay away from Prometheus version updates. In the past they once made all floats into ints, then reverted, now it seems like they've unreverted and you can see a failure here https://travis-ci.org/census-instrumentation/opencensus-service/builds/515417341#L1241

songy23: Unfortunately this is a consequence of running `GO111MODULE=on go mod tidy`. I think I have to fix the test failure again.

exporter/prometheusexporter/prometheus_test.go
```diff
@@ -167,7 +167,7 @@ prometheus:
 	_ = res.Body.Close()
 	want := `# HELP test_this_one_there_where_ Extra ones
 # TYPE test_this_one_there_where_ counter
-test_this_one_there_where_{arch="x86",code="one",foo="bar",os="windows"} 99.0
+test_this_one_there_where_{arch="x86",code="one",foo="bar",os="windows"} 99
```
odeke-em: Please revert this and keep the Prometheus version as before.

odeke-em: I posted some comments earlier on why I am not enthusiastic about the Prometheus import change.

songy23: Tried but that didn't work. Even if I reverted the Prometheus version in go.mod, after I ran `GO111MODULE=on go mod tidy` it will still be updated to `v0.9.3-0.20190127221311-3c4408c8b829`.

songy23:
FYI in https://github.com/census-instrumentation/opencensus-service/pull/512/commits/e66794d3e65a5bfa012bdd48eb78a43adea7f8ed I only manually modified the version for `go.opencensus.io` and then ran `GO111MODULE=on go mod tidy`. All the other changes are automatically generated by go module and I cannot edit.

odeke-em:
This change from opencensus-go https://github.com/census-instrumentation/opencensus-go/blob/43463a80402d8447b7fce0d2c58edf1687ff0b58/go.mod#L15 is what introduced it
/cc @rghetia

songy23:
We're working on moving Go exporters to their own repos. Will update the dependencies after that work is done.

PR #516, opened 2019-04-05
During exporter initialization, log error and continue to run.

rghetia:
In a case where there are multiple exporters and if one them doesn't work for whatever reason, it should still continue to run and export metrics/traces through other exporters.

### DIFF
--- a/cmd/ocagent/main.go
+++ b/cmd/ocagent/main.go
@@ -113,9 +113,11 @@ func runOCAgent() {
 		log.Fatalf("Failed to start net/http/pprof: %v", err)
 	}
 
-	traceExporters, metricsExporters, closeFns, err := config.ExportersFromViperConfig(logger, viperCfg)
-	if err != nil {
-		log.Fatalf("Config: failed to create exporters from YAML: %v", err)
+	traceExporters, metricsExporters, closeFns, errs := config.ExportersFromViperConfig(logger, viperCfg)
+	if len(errs) > 0 {
+		for _, err := range errs {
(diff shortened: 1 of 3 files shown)


rghetia:
Thx @songy23

rghetia:
@pjanotti can you please look at it?

pjanotti:
Hi @rghetia error at that stage should be a configuration error, in principle I think it is better to fail to start in such cases. Note that exporters failing to connect to destinations or other intermittent failure shouldn't prevent the start of agent or collector. Did you experience something different than that?

rghetia:
> Hi @rghetia error at that stage should be a configuration error, in principle I think it is better to fail to start in such cases. Note that exporters failing to connect to destinations or other intermittent failure shouldn't prevent the start of agent or collector. Did you experience something different than that?

I also experience another scenario where configuration is right but SD fails due to lack of credentials. I have demo setup that runs on GKE with SD exporter configured with right project. Same setup when I run locally on mac, SD exporter returns an error and agent stops.

pjanotti:
I am also inconvenienced by this behavior when testing/trying stuff, but, there is a purpose on it: avoid running in a degraded mode due to bad config, since it enhances chances of early detection, instead of proceeding with some exporters totally disabled (indicated by a single log message at startup). 

If SD credentials are something that can be changed during runtime then having SD exporter log a warning and not fail at the start seems a safer approach.

rghetia:
Makes sense. I'll close this PR.

PR #520, opened 2019-04-06
[WIP]fix aggregation type for per process stats and system stats.

rghetia:
Also recording delta for sum aggregation by preserving previous values.

### DIFF
--- a/receiver/vmmetricsreceiver/vm_metrics_collector.go
+++ b/receiver/vmmetricsreceiver/vm_metrics_collector.go
@@ -148,19 +148,27 @@ func (vmc *VMMetricsCollector) scrape(prevProcStat *procfs.ProcStat, prevStat *p
 		cpuStat := stat.CPUTotal
 		stats.Record(
 			ctx,
-			mProcessesCreated.M(int64(stat.ProcessCreated)),
 			mProcessesRunning.M(int64(stat.ProcessesRunning)),
-			mProcessesBlocked.M(int64(stat.ProcessesBlocked)),
-			mUserCPUSeconds.M(cpuStat.User),
-			mNiceCPUSeconds.M(cpuStat.Nice),
-			mSystemCPUSeconds.M(cpuStat.System),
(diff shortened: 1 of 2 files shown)


receiver/vmmetricsreceiver/vm_metrics_collector.go
```diff
@@ -127,7 +136,14 @@ func (vmc *VMMetricsCollector) scrape() {
 			mSystemCPUSeconds.M(cpuStat.System),
 			mIdleCPUSeconds.M(cpuStat.Idle),
 			mIowaitCPUSeconds.M(cpuStat.Iowait))
+		if prevStat != nil {
+			stats.Record(ctx, mProcessesCreated.M(int64(stat.ProcessCreated - prevStat.ProcessCreated)))
```
songy23: Should `ProcessRunning` and `ProcessBlocked` be updated too?

songy23:
Also please format the files.

receiver/vmmetricsreceiver/vm_metrics_collector.go
rghetia: No. I checked, they are instantaneous values.

rghetia:
@songy23 , I reverted back to aggregation sum for system (user, system, idle, nice, iowait). But added the delta recording for each.

VM stat is correctly reported now. However, when I run the agent as daemon set and mount /proc from host as /data/proc on agent container, vm stats are correct but agent process stat is not. If I don't mount /data/proc, process stat is correct but vm stat is not.

rghetia:
@songy23, @pjanotti thx for the review. There is one more update which contains following two changes.
1. 'prev*' was not correctly initialized in earlier commit. It should be set at every poll. 
2. The process level cpu_seconds has only fractional seconds between two polls. Converting fractions to int64 makes it zero. Hence changed it to float64.

PTAL.

rghetia:
@songy23 or @pjanotti can you please merge it? I don't have access.

PR #521, opened 2019-04-08
receiver/vmmetrics: Allow to have a different process mount point.

songy23:
In cases like when OC-Agent is running as a daemon set, the mount point of daemon set process may be different from the default one. Adding an option to allow users specify a different mount point for Agent process.

### DIFF
--- a/receiver/vmmetricsreceiver/metrics_receiver.go
+++ b/receiver/vmmetricsreceiver/metrics_receiver.go
@@ -60,7 +60,7 @@ func New(v *viper.Viper, consumer consumer.MetricsConsumer) (*Receiver, error) {
 	var cfg Configuration
 
 	// Unmarshal our config values (using viper's mapstructure)
-	err := v.Unmarshal(&cfg)
+	err := unmarshal(&cfg, v.AllSettings())
 	if err != nil {
 		return nil, fmt.Errorf("vmmetrics receiver failed to parse config: %s", err)
 	}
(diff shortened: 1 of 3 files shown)


receiver/vmmetricsreceiver/metrics_receiver.go
```diff
@@ -107,3 +108,21 @@ func (vmr *Receiver) StopMetricsReception(ctx context.Context) error {
 	})
 	return err
 }
+
+// TODO(songya): investigate why viper.Unmarshal didn't work, remove this method and use viper.Unmarshal instead.
```
pjanotti: At first this looks strange, perhaps is it because the suffix of two entries are matching?

receiver/vmmetricsreceiver/metrics_receiver.go
```diff
@@ -107,3 +108,21 @@ func (vmr *Receiver) StopMetricsReception(ctx context.Context) error {
...
+func unmarshal(cfg *Configuration, settings map[string]interface{}) error {
+	if interval, ok := settings["scrape_interval"]; ok {
+		intervalInSecs := interval.(int)
```
pjanotti: this can crash on bad config but assuming that we will figure out this soon and don't need this function anymore

receiver/vmmetricsreceiver/metrics_receiver.go
songy23: Actually even if I removed both *_mount_point configs Viper still failed to parse the config. I'll investigate and fix it in a separate PR.

receiver/vmmetricsreceiver/metrics_receiver.go
songy23: Yes, filed https://github.com/census-instrumentation/opencensus-service/issues/522 to keep track.

PR #536, opened 2019-04-29
Add span attribute key mapping support (#9)

pjanotti:
Many libraries and services didn't implement consistent naming for their span attribute keys, this feature provides the capability of making the attribute keys consistent for different libraries and services without the need to change their code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,9 +277,13 @@ agent/client health information/inventory metadata to downstream exporters.
 
 ### <a name="global-attributes"></a> Global Attributes
 
-The collector also takes some global configurations that modify its behavior for all receivers / exporters. One of the configurations
-available is to add Attributes or Tags to all spans passing through this collector. These additional attributes can be configured to either overwrite
-attributes if they already exists on the span, or respect the original values. An example of this is provided below.
+The collector also takes some global configurations that modify its behavior for all receivers / exporters. 
+
+1. Add Attributes to all spans passing through this collector. These additional attributes can be configured to either overwrite existing keys if they already exists on the span, or respect the original values.
+2. The key of each attribute can also be mapped to different stringss using the `key-mapping` configuration. The key matching is case sensitive.
+
+An example using these configurations of this is provided below.
+
 ```yaml
 global:
   attributes:
@@ -290,6 +294,14 @@ global:
       some_int: 1234
       some_float: 3.14159
       some_bool: false
+    key-mapping:
+      # key-mapping is used to replace the attribute key with different keys
+      - key: servertracer.http.responsecode
+        replacement: http.status_code
+      - key:  servertracer.http.responsephrase
+        replacement: http.message
+        overwrite: true # replace attribute key even if the replacement string is already a key on the span attributes
+        keep: true # keep the attribute with the original key
 ```
 
 ### <a name="tail-sampling"></a>Intelligent Sampling
--- a/README.md
+++ b/README.md
@@ -279,8 +279,8 @@ agent/client health information/inventory metadata to downstream exporters.
 
 The collector also takes some global configurations that modify its behavior for all receivers / exporters. 
 
-1. Add Attributes to all spans passing through this collector. These additional attributes can be configured to either overwrite existing keys if they already exists on the span, or respect the original values.
-2. The key of each attribute can also be mapped to different stringss using the `key-mapping` configuration. The key matching is case sensitive.
+1. Add Attributes to all spans passing through this collector. These additional attributes can be configured to either overwrite existing keys if they already exist on the span, or respect the original values.
+2. The key of each attribute can also be mapped to different strings using the `key-mapping` configuration. The key matching is case sensitive.
 
 An example using these configurations of this is provided below.
 
--- a/cmd/occollector/app/builder/processor_builder_test.go
+++ b/cmd/occollector/app/builder/processor_builder_test.go
@@ -1,0 +1,89 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package builder
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/census-instrumentation/opencensus-service/processor/attributekeyprocessor"
+)
+
+func TestGlobalProcessorCfg_InitFromViper(t *testing.T) {
+	tests := []struct {
+		name string
+		file string
+		want *AttributesCfg
+	}{
+		{
+			name: "key_mapping",
+			file: "./testdata/global_attributes_key_mapping.yaml",
+			want: &AttributesCfg{
+				KeyReplacements: []attributekeyprocessor.KeyReplacement{
+					{
+						Key:    "servertracer.http.responsecode",
+						NewKey: "http.status_code",
+					},
+					{
+						Key:          "servertracer.http.responsephrase",
+						NewKey:       "http.message",
+						Overwrite:    true,
+						KeepOriginal: true,
+					},
+				},
+			},
+		},
+		{
+			name: "all_settings",
+			file: "./testdata/global_attributes_all.yaml",
+			want: &AttributesCfg{
+				Overwrite: true,
+				Values:    map[string]interface{}{"some_string": "hello world"},
+				KeyReplacements: []attributekeyprocessor.KeyReplacement{
+					{
+						Key:    "servertracer.http.responsecode",
+						NewKey: "http.status_code",
+					},
+					{
+						Key:          "servertracer.http.responsephrase",
+						NewKey:       "http.message",
+						Overwrite:    true,
+						KeepOriginal: true,
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			v, err := loadViperFromFile(tt.file)
+			if err != nil {
+				t.Fatalf("Failed to load viper from test file: %v", err)
+			}
+
+			cfg := NewDefaultMultiSpanProcessorCfg().InitFromViper(v)
+
+			got := cfg.Global.Attributes
+			if got == nil {
+				t.Fatalf("got nil, wan non-nil")
+			}
+
+			if diff := cmp.Diff(got, tt.want); diff != "" {
+				t.Errorf("Mismatched global configuration\n-Got +Want:\n\t%s", diff)
+			}
+		})
+	}
+}
--- a/cmd/occollector/app/builder/processor_builder_test.go
+++ b/cmd/occollector/app/builder/processor_builder_test.go
@@ -78,7 +78,7 @@ func TestGlobalProcessorCfg_InitFromViper(t *testing.T) {
 
 			got := cfg.Global.Attributes
 			if got == nil {
-				t.Fatalf("got nil, wan non-nil")
+				t.Fatalf("got nil, want non-nil")
 			}
 
 			if diff := cmp.Diff(got, tt.want); diff != "" {
--- a/cmd/occollector/app/builder/testdata/global_attributes_key_mapping.yaml
+++ b/cmd/occollector/app/builder/testdata/global_attributes_key_mapping.yaml
@@ -1,0 +1,9 @@
+global:
+  attributes:
+    key-mapping:
+      - key: servertracer.http.responsecode
+        replacement: http.status_code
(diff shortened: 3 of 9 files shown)


pjanotti:
@songy23 @bogdandrutu ptal when you have a chance

README.md
```diff
@@ -277,9 +277,13 @@ agent/client health information/inventory metadata to downstream exporters.
 
 ### <a name="global-tags"></a> Global Tags
 
-The collector also takes some global configurations that modify its behavior for all receivers / exporters. One of the configurations
-available is to add Attributes or Tags to all spans passing through this collector. These additional tags can be configured to either overwrite
-attributes if they already exists on the span, or respect the original values. An example of this is provided below.
+The collector also takes some global configurations that modify its behavior for all receivers / exporters. 
+
+1. Add Attributes to all spans passing through this collector. These additional attributes can be configured to either overwrite existing keys if they already exists on the span, or respect the original values.
```
songy23: Nit: they already exists -> they already exist

README.md
```diff
@@ -277,9 +277,13 @@ agent/client health information/inventory metadata to downstream exporters.
...
+2. The key of each attribute can also be mapped to different stringss using the `key-mapping` configuration. The key matching is case sensitive.
```
songy23: Nit: s/stringss/strings

README.md
```diff
@@ -290,6 +294,14 @@ global:
...
       some_float: 3.14159
       some_bool: false
+    key-mapping:
+      # key-mapping is used to replace the attribute key with different keys
+      - key: servertracer.http.responsecode
+        replacement: http.status_code
+      - key:  servertracer.http.responsephrase
+        replacement: http.message
+        overwrite: true # replace attribute key even if the replacement string is already a key on the span attributes
+        keep: true # keep the attribute with the original key
```
songy23: Nit: `keep_original` may be clearer.

README.md
```diff
@@ -290,6 +294,14 @@ global:
       some_int: 1234
       some_float: 3.14159
       some_bool: false
+    key-mapping:
+      # key-mapping is used to replace the attribute key with different keys
+      - key: servertracer.http.responsecode
+        replacement: http.status_code
+      - key:  servertracer.http.responsephrase
+        replacement: http.message
+        overwrite: true # replace attribute key even if the replacement string is already a key on the span attributes
```
songy23: Nit: consider `overwrite_replacement`.

flands: I am not sure I follow overwrite and keep -- why have two options? Either you override or you do not, right?

cmd/occollector/app/builder/testdata/global_attributes_key_mapping.yaml
```diff
@@ -0,0 +1,9 @@
+global:
+  attributes:
+    key-mapping:
+      - key: servertracer.http.responsecode
+        replacement: http.status_code
+      - key: servertracer.http.responsephrase
+        replacement: http.message
+        overwrite: true
```
flands: Same question

processor/attributekeyprocessor/attributekeyprocessor_test.go
```diff
@@ -0,0 +1,317 @@
...
+						{
+							Name:       &tracepb.TruncatableString{Value: "test"},
+							Attributes: &tracepb.Span_Attributes{},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "span_replace_key",
```
songy23: Consider adding another test where both `Overwrite` and `KeepOriginal` for the same `KeyReplacement`.

processor/attributekeyprocessor/attributekeyprocessor.go
```diff
@@ -0,0 +1,100 @@
...
+func NewTraceProcessor(nextConsumer consumer.TraceConsumer, replacements ...KeyReplacement) (processor.TraceProcessor, error) {
+	if nextConsumer == nil {
+		return nil, errors.New("nextConsumer is nil")
+	}
+
+	lenReplacements := len(replacements)
+	if lenReplacements > 0 {
+		seenKeys := make(map[string]bool, lenReplacements)
+		seenNewKeys := make(map[string]bool, lenReplacements)
+		for _, replacement := range replacements {
```
songy23: What about cycles? e.g {foo:bar, bar:foo}

processor/attributekeyprocessor/attributekeyprocessor.go
```diff
@@ -0,0 +1,100 @@
...
+
+// NewTraceProcessor returns a processor.TraceProcessor
+func NewTraceProcessor(nextConsumer consumer.TraceConsumer, replacements ...KeyReplacement) (processor.TraceProcessor, error) {
+	if nextConsumer == nil {
+		return nil, errors.New("nextConsumer is nil")
+	}
+
+	lenReplacements := len(replacements)
+	if lenReplacements > 0 {
+		seenKeys := make(map[string]bool, lenReplacements)
```
songy23: IMO you can keep one map. All keys need to be unique, no matter it's an existing key or a new one.

README.md
pjanotti: Perhaps needs better naming/explanation, the idea is that since attributes are entries in a map/hashtable `keep` offers the option of preserving the original key-value pair independent of the choice to `overwrite` the original or not. I think you are seeing it as a single step operation and I am seeing as the two steps that happen with the operation in a map: add/update entry for new key -value pair according to `overwrite`, then decide what to do about the old key-value pair according to `keep`.

Makes sense?

processor/attributekeyprocessor/attributekeyprocessor.go
pjanotti: Good point, it also deals with the cycle issue that you mentioned above.

README.md
pjanotti: I'm going to keep the shorter variations, I think people will need to check the readme anyway for this.

README.md
pjanotti: Clarified via offline conversation.

pjanotti:
@songy23 thanks for the careful review

flands:
IMO overwrite and keep are clear -- also fan of brevity

pjanotti:
@tigrannajaryan not sure why I can't add you as a reviewer for this one, but, ptal.

cmd/occollector/app/collector/processors.go
```diff
@@ -315,12 +317,23 @@ func startProcessor(v *viper.Viper, logger *zap.Logger) (consumer.TraceConsumer,
...
+				addattributesprocessor.WithOverwrite(multiProcessorCfg.Global.Attributes.Overwrite),
+			)
+		}
+		if len(multiProcessorCfg.Global.Attributes.KeyReplacements) > 0 {
+			if tp == nil {
+				tp = multiconsumer.NewTraceProcessor(traceConsumers)
+			}
+			tp, _ = attributekeyprocessor.NewTraceProcessor(tp, multiProcessorCfg.Global.Attributes.KeyReplacements...)
+		}
 		return tp, closeFns
```
tigrannajaryan: This will return `tp=nil` if both `Values` and `KeyReplacements` are empty, is that OK?

processor/attributekeyprocessor/attributekeyprocessor.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/census-instrumentation/opencensus-service/data"
+	"github.com/census-instrumentation/opencensus-service/processor"
+)
+
+// KeyReplacement identifies a key and its respective replacement.
+type KeyReplacement struct {
+	// Key the attribute key to be replaced.
+	Key string `mapstructure:"key"`
+	// NewKey is the value that will be used as the new key for the attribute.
+	NewKey string `mapstructure:"replacement"`
```
tigrannajaryan: Perhaps for consistency name this field `Replacement` instead of `NewKey`.

processor/attributekeyprocessor/attributekeyprocessor.go
tigrannajaryan: Is it invalid configuration to attempt to replace 2 different source keys by the same replacement key? 
For example:
```yaml
    key-mapping:
      - key: servertracer.http.responsecode
        replacement: http.status_code
      - key:  anotherservice.http.responsecode
        replacement: http.status_code
```
This might be a valid use case, right?

processor/attributekeyprocessor/attributekeyprocessor.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, span := range td.Spans {
+		if span == nil || span.Attributes == nil || len(span.Attributes.AttributeMap) == 0 {
+			// Nothing to do
+			continue
+		}
+
+		attribMap := span.Attributes.AttributeMap
+		for _, replacement := range akp.replacements {
+			if keyValue, oldKeyPresent := attribMap[replacement.Key]; oldKeyPresent {
+				if _, newKeyPresent := attribMap[replacement.NewKey]; !newKeyPresent || replacement.Overwrite {
```
tigrannajaryan: Will this do unnecessary map lookup for the case when `replacement.Overwrite==true`? Is Go compiler smart enough to eliminate the lookup call in that case?

cmd/occollector/app/builder/processor_builder_test.go
```diff
@@ -0,0 +1,89 @@
...
+			v, err := loadViperFromFile(tt.file)
+			if err != nil {
+				t.Fatalf("Failed to load viper from test file: %v", err)
+			}
+
+			cfg := NewDefaultMultiSpanProcessorCfg().InitFromViper(v)
+
+			got := cfg.Global.Attributes
+			if got == nil {
+				t.Fatalf("got nil, wan non-nil")
```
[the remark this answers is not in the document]
pjanotti: done

processor/attributekeyprocessor/attributekeyprocessor.go
pjanotti: yeah, that initial lookup can be avoided.

processor/attributekeyprocessor/attributekeyprocessor.go
pjanotti: Internally I prefer `NewKey` to emphasize that it is a new entry the map, for the config itself `new-key` is a implementation detail.

cmd/occollector/app/collector/processors.go
pjanotti: good catch, I had manually tested with `values: {}`, `key-mapping: {}`, and `attributes: {}` but if there is `attributes.overwrite: true` it will cause to return nil and crash. Adding a test to cover this case.

processor/attributekeyprocessor/attributekeyprocessor.go
pjanotti: Correct, I missed that. It is an "advanced" case but I don't see a reason to block it. Putting together with the ordering and the overwrite and keep flags the behavior is actually fine controllable.

cmd/occollector/app/collector/processors.go
```diff
@@ -310,17 +312,32 @@ func startProcessor(v *viper.Viper, logger *zap.Logger) (consumer.TraceConsumer,
 	}
 
 	// Wraps processors in a single one to be connected to all enabled receivers.
-	if multiProcessorCfg.Global != nil && multiProcessorCfg.Global.Attributes != nil {
+	hasAttributesProcessing := multiProcessorCfg.Global != nil &&
+		multiProcessorCfg.Global.Attributes != nil &&
+		(len(multiProcessorCfg.Global.Attributes.Values) > 0 ||
+			len(multiProcessorCfg.Global.Attributes.KeyReplacements) > 0)
+	if hasAttributesProcessing {
```
sjkaris: Should we maybe leave this inital check just to check `multiProcessorCfg.Global != nil && multiProcessorCfg.Global.Attributes != nil` ?  We check the other two conditions again below anyway.

pjanotti: there is a possible way to make both Values and Replacements to be empty see https://github.com/census-instrumentation/opencensus-service/pull/536#discussion_r281872489 for details

processor/attributekeyprocessor/attributekeyprocessor.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, span := range td.Spans {
+		if span == nil || span.Attributes == nil || len(span.Attributes.AttributeMap) == 0 {
+			// Nothing to do
+			continue
+		}
+
+		attribMap := span.Attributes.AttributeMap
+		for _, replacement := range akp.replacements {
+			if keyValue, oldKeyPresent := attribMap[replacement.Key]; oldKeyPresent {
+				newKeyMapped := func() bool {
```
sjkaris: nit: should this be a separate function or just used inline? `if _, found := attribMap[replacement.NewKey]; replacement.Overwrite || !found) {`

cmd/occollector/app/collector/processors.go
sjkaris: We would chain the building:
Eg (in psuedo code ish)
```
tp := multiconsumer.NewTraceProcessor(traceConsumers)
if multiProcessorCfg.Global != nil && multiProcessorCfg.Global.Attributes != nil {
    if doAdd {
        tp = addattributesprocessor.NewTraceProcessor(tp, ....)
    }
    if doRemap {
        tp = attributekeyprocessor.NewTraceProcessor(tp, ...) 
    }
}
return tp
```

Issue #519, opened 2019-04-06
exporter/stackdriver: error posting too frequent when >1 pod reports same metrics, 


rghetia:
On GKE deployment if more than one pod reports same metrics then SD backend rejects data points with the reason
""rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: One or more points were written more frequently than the maximum sampling period configured for the metric. "

This is because there is no label that differentiates data points from two different nodes.
Here is a sample tracez page from an ocagent.

```
Span Name: contrib.go.opencensus.io/exporter/stackdriver.uploadMetrics

10 Requests

When                       Elapsed (sec)
----------------------------------------
2019/04/05-20:24:48.505738      .224350 trace_id: 4dca2a35177c0f21c30522b54722b337 span_id: ef227ffa88b5fb70
                                        Status{canonicalCode=UNKNOWN, description="rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/server/received_bytes_per_rpc, Timestamps: {Youngest Existing: '2019/04/05-13:24:44.182', New: '2019/04/05-13:24:44.504'}}: timeSeries[1]; One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/server/sent_bytes_per_rpc, Timestamps: {Youngest Existing: '2019/04/05-13:24:44.182', New: '2019/04/05-13:24:44.504'}}: timeSeries[2]"}
2019/04/05-20:25:48.507615      .259883 trace_id: de00c24549ebb805bf823ea48dbf707b span_id: e969bae9e354d838
                                        Status{canonicalCode=UNKNOWN, description="rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: One or more points were written more frequently than the maximum sampling period configured for the metric. {Metric: custom.googleapis.com/opencensus/grpc.io/server/completed_rpcs, Timestamps: {Youngest Existing: '2019/04/05-13:25:44.182', New: '2019/04/05-13:25:44.504'}}: timeSeries[1]"}
2019/04/05-20:14:48.539989      .292571 trace_id: 8036da2a6e7a5eb3f2dba4dd06a74949 span_id: 219fc31ed27e0a09
[...]

PR #537, opened 2019-05-01
Add metric labels and trace attributes to stackdriver exporter

bakins:
Add ability to set trace attributes and metric labels as described in https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stackdriver.go#L210-L228

We set these when using Stackdriver exporter directly in our applications to set pod, namespace, etc inside Kubernetes.

Should also allow one to add labels to address https://github.com/census-instrumentation/opencensus-service/issues/519

### DIFF
--- a/exporter/stackdriverexporter/stackdriver.go
+++ b/exporter/stackdriverexporter/stackdriver.go
@@ -30,10 +30,13 @@ import (
 )
 
 type stackdriverConfig struct {
-	ProjectID     string `mapstructure:"project,omitempty"`
-	EnableTracing bool   `mapstructure:"enable_tracing,omitempty"`
-	EnableMetrics bool   `mapstructure:"enable_metrics,omitempty"`
-	MetricPrefix  string `mapstructure:"metric_prefix,omitempty"`
+	ProjectID       string                 `mapstructure:"project,omitempty"`
+	EnableTracing   bool                   `mapstructure:"enable_tracing,omitempty"`
+	EnableMetrics   bool                   `mapstructure:"enable_metrics,omitempty"`
+	MetricPrefix    string                 `mapstructure:"metric_prefix,omitempty"`
+	TraceAttributes map[string]interface{} `mapstructure:"trace_attributes,omitempty"`
+	MetricLabels    map[string]string      `mapstructure:"metric_labels,omitempty"`
+	Location        string                 `mapstructure:"location,omitempty"`
 }
 
 // TODO: Add metrics support to the exporterwrapper.
@@ -63,6 +66,11 @@ func StackdriverTraceExportersFromViper(v *viper.Viper) (tps []consumer.TraceCon
 	// TODO:  For each ProjectID, create a different exporter
 	// or at least a unique Stackdriver client per ProjectID.
 
+	labels := &stackdriver.Labels{}
+	for k, v := range sc.MetricLabels {
+		labels.Set(k, v, k)
+	}
+
 	sde, serr := stackdriver.NewExporter(stackdriver.Options{
 		// If the project ID is an empty string, it will be set by default based on
 		// the project this is running on in GCP.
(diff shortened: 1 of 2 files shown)


exporter/stackdriverexporter/stackdriver.go
```diff
@@ -30,10 +30,13 @@ import (
...
 type stackdriverConfig struct {
-	ProjectID     string `mapstructure:"project,omitempty"`
-	EnableTracing bool   `mapstructure:"enable_tracing,omitempty"`
-	EnableMetrics bool   `mapstructure:"enable_metrics,omitempty"`
-	MetricPrefix  string `mapstructure:"metric_prefix,omitempty"`
+	ProjectID       string                 `mapstructure:"project,omitempty"`
+	EnableTracing   bool                   `mapstructure:"enable_tracing,omitempty"`
+	EnableMetrics   bool                   `mapstructure:"enable_metrics,omitempty"`
+	MetricPrefix    string                 `mapstructure:"metric_prefix,omitempty"`
+	TraceAttributes map[string]interface{} `mapstructure:"trace_attributes,omitempty"`
```
pjanotti: Any reason to not use [global tags](https://github.com/Omnition/internal-opencensus-service#-global-tags) instead? In general we've been approaching this type of issue making it happen in the oc proto format so all exporters can generate the same data.

We still don't have the same for metric labels and prefix but both are applicable too.

exporter/stackdriverexporter/stackdriverexporter_test.go
```diff
@@ -14,4 +14,120 @@
...
+		wantErr    string
+		traceLen   int
+		metricsLen int
+	}{
+		{
+			name: "empty config",
+		},
+		{
+			name: "no project",
+			config: map[string]interface{}{
```
pjanotti: this test is not returning error...

exporter/stackdriverexporter/stackdriver.go
bakins: That URL 404's

pjanotti: ops, just fixed the link

exporter/stackdriverexporter/stackdriverexporter_test.go
bakins: Travis runs on GCE, so the project is detected. I removed the test case.

exporter/stackdriverexporter/stackdriver.go
bakins: If you have links for other docs, that'd be great.  My searching came up empty.  If I can use an existing feature, I'd rather do so.  Thanks!

pjanotti: Right now we just have global tags for tracing https://github.com/census-instrumentation/opencensus-service#-global-tags but we should implement the equivalent for metrics /cc @songy23

pjanotti: btw, there is an issue regarding naming: on OC we should refer to these as attributes instead of tags but the docs were not updated yet.

pjanotti: another related feature for attributes is in the pipelines https://github.com/census-instrumentation/opencensus-service/pull/536

pjanotti: Hi @bakins, given that #536 was merged today, do you still need to add some functionality from this PR?

pjanotti:
Hi @bakins, given that #536 was merged 11 days ago, do you still need to add some functionality from this PR?

pjanotti:
@bakins I'm closing this PR, let's know if you still want to move ahead with it.

bakins:
I missed the notification for this. Sorry about that. #536 looks good for spans, I'd like something like that for metrics as well. I can take a look unless it is already in progress.

PR #538, opened 2019-05-01
 Upgrade versions of opencensus core and exporters.

songy23:
In OpenCensus-Go v0.21.0, all exporters have been moved to their own repos, and the vanity urls have been updated to "contrib.go.opencensus.io/exporter/" instead of "go.opencensus.io/exporter/".


exporter/opencensusexporter/opencensusexporter_test.go
```diff
@@ -22,11 +22,12 @@ import (
 
 func TestOpenCensusTraceExportersFromViper(t *testing.T) {
 	v := viper.New()
-	v.Set("opencensus", struct{}{})
+	v.Set("opencensus", map[interface{}]interface{}{})
```
pjanotti: q.: what went wrong with the old code after the change? this was the reason to add the `endpoint` too?

songy23: After this PR the change Viper expects "opencensus" to be a `map` rather than a `struct`. If we set it to a struct here we will get a different error saying "Decoding failed, expect map got struct".

songy23:
> vet failed on Travis you probably have the package on your local modcache

Yeah I was thinking about the same thing, that's why I updated the make rules in https://github.com/census-instrumentation/opencensus-service/pull/538/commits/5fd1548a97c7fac95321d22e11ae9f23560db9fe. Would appreciate it if you have a better suggestion here.

songy23:
Thanks - coverage decrease is because of https://github.com/census-instrumentation/opencensus-service/pull/538/files#diff-b62f24d39c37eacaa5d4545576874137R82. I'll send a follow-up PR to fix that test since there will be a lot of changes.

PR #539, opened 2019-05-02
receiver/prometheues: Fix test.

songy23:
Prometheus exporter has been migrated to consume `Metric` rather than `ViewData`. Fix this test by producing `Metric`s in the test.

### DIFF
--- a/receiver/prometheusreceiver/metrics_receiver_test.go
+++ b/receiver/prometheusreceiver/metrics_receiver_test.go
@@ -21,20 +21,19 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"reflect"
-	"sort"
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
 	"contrib.go.opencensus.io/exporter/prometheus"
 	"github.com/spf13/viper"
-	"go.opencensus.io/stats"
-	"go.opencensus.io/stats/view"
-	"go.opencensus.io/tag"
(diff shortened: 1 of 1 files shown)


receiver/prometheusreceiver/metrics_receiver_test.go
```diff
@@ -195,201 +198,88 @@ buffer_count: 2
...
-					Name: "demo",
-				},
-				Attributes: map[string]string{
-					"scheme": "http",
-					"port":   port,
-				},
+	wantNode := &commonpb.Node{
+		Identifier: &commonpb.ProcessIdentifier{
+			HostName:      host,
+			XXX_sizecache: 11,
```
odeke-em: How come these `XXX*` fields are getting populated?

songy23: I'm not sure either. My guess is these fields are added to the gen-go files when they're re-generated by a newer version of proto compiler. Anyway if I don't include these fields, comparison would fail.

odeke-em: Perhaps you'll need to examine how to fix these comparison options in https://github.com/census-instrumentation/opencensus-service/pull/539/files#diff-b62f24d39c37eacaa5d4545576874137R319 perhaps with `proto.Equal` otherwise anyone using this new comparison helper is going to have to add those `XXX` fields

songy23: Just tried `proto.Equal`, unfortunately by using it we need to know the exact `timestamp` of the given metrics, which is also tedious. I'll try another way around.

PR #541, opened 2019-05-02
Add zpages to collector

pjanotti:
Move zpages to its own type and share the usage of that type between agent and collector.


cmd/occollector/app/collector/collector_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func TestApplication_Start(t *testing.T) {
+	// Without exporters the collector will start and just shutdown, no error is expected.
+	App.v.Set("logging-exporter", true)
+	appDone := make(chan struct{})
+	go func() {
+		if err := App.Start(); err != nil {
+			t.Fatalf("App.Start() got %v, want nil", err)
+		}
+		close(appDone)
```
odeke-em: Please make this a defer at the very top of the goroutine `defer close(appDone)`

cmd/occollector/app/collector/collector_test.go
```diff
@@ -0,0 +1,50 @@
...
+	if !isAppReady(t) {
+		t.Fatalf("App didn't reach ready state")
+	}
+	close(App.stopTestChan)
+	<-appDone
+}
+
+func isAppReady(t *testing.T) bool {
+	client := &http.Client{}
+	resp, err := client.Get("http://localhost:13133")
```
odeke-em: This hardcoding of address is bound to start some errors and be inflexible. If we are starting
the application and have a start service running this might cause problems. Also test servers
can be spawned on any address.

pjanotti: Ensured no port conflicts for this test, I'm sure there are some others at the same risk on the code base but didn't want to go overboard finding and fixing them to avoid expanding too much this PR.

pjanotti:
Thanks for the review @odeke-em

PR #554, opened 2019-05-20
receiver/vmmetrics: Simplify metrics creation.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-service/issues/534.

Previously we create `measure`s and `view`s for VM metrics and then convert `view.Data` to proto metrics, which is redundant. This PR changes it to directly create proto metrics instead.


receiver/vmmetricsreceiver/vm_metrics_collector.go
```diff
@@ -113,104 +106,137 @@ func (vmc *VMMetricsCollector) StopCollection() {
...
-				mNiceCPUSeconds.M(cpuStat.Nice-prevStat.CPUTotal.Nice),
-				mSystemCPUSeconds.M(cpuStat.System-prevStat.CPUTotal.System),
-				mIdleCPUSeconds.M(cpuStat.Idle-prevStat.CPUTotal.Idle),
-				mIowaitCPUSeconds.M(cpuStat.Iowait-prevStat.CPUTotal.Iowait))
+			processCreatedDelta = stat.ProcessCreated - prevStat.ProcessCreated
+			cpuTimeUserDelta = cpuStat.User - prevStat.CPUTotal.User
+			cpuTimeSystemDelta = cpuStat.System - prevStat.CPUTotal.System
+			cpuTimeIdleDelta = cpuStat.Idle - prevStat.CPUTotal.Idle
+			cpuTimeNiceDelta = cpuStat.Nice - prevStat.CPUTotal.Nice
+			cpuTimeIOWaitDelta = cpuStat.Iowait - prevStat.CPUTotal.Iowait
```
rghetia: don't need delta if it is directly exported.

receiver/vmmetricsreceiver/vm_metrics_constants.go
```diff
@@ -15,131 +15,88 @@
...
-	Description: mUserCPUSeconds.Description(),
-	Measure:     mUserCPUSeconds,
-	Aggregation: view.Sum(),
-	TagKeys:     nil,
+var metricCPUSeconds = &metricspb.MetricDescriptor{
+	Name:        "system/cpu_seconds",
+	Description: "Total kernel/system CPU seconds broken down by different stages",
+	Unit:        "s",
+	Type:        metricspb.MetricDescriptor_CUMULATIVE_DOUBLE,
+	LabelKeys:   []*metricspb.LabelKey{{Key: "stage", Description: "Stage of CPU time, e.g user/system/idle"}},
```
rghetia: nit: would state be better than stage?
idle state
iowait state
running state (kernel vs user)

receiver/vmmetricsreceiver/vm_metrics_collector.go
songy23: You're right - good catch.

receiver/vmmetricsreceiver/vm_metrics_constants.go
songy23: SG, updated.

receiver/vmmetricsreceiver/vm_metrics_constants.go
```diff
@@ -15,131 +15,90 @@
...
-var mUserCPUSeconds = stats.Float64("system/cpu_seconds/user", "Total kernel/system user CPU seconds", "s")
-var viewUserCPUSeconds = &view.View{
-	Name:        mUserCPUSeconds.Name(),
-	Description: mUserCPUSeconds.Description(),
-	Measure:     mUserCPUSeconds,
-	Aggregation: view.Sum(),
-	TagKeys:     nil,
+var metricCPUSeconds = &metricspb.MetricDescriptor{
+	Name:        "system/cpu_seconds",
+	Description: "Total kernel/system CPU seconds broken down by different stages",
```
rghetia: nit: s/stages/states/

songy23: Yes, just fixed in https://github.com/census-instrumentation/opencensus-service/commit/d532ffd2c2a558b8d017ac0eeac60b34aa772d2d :)

receiver/vmmetricsreceiver/vm_metrics_collector.go
```diff
@@ -113,104 +103,107 @@ func (vmc *VMMetricsCollector) StopCollection() {
...
 
 	pid := os.Getpid()
 	var proc procfs.Proc
 	var err error
-	if vmc.processFs == "" {
-		proc, err = vmc.fs.NewProc(pid)
-	} else {
-		proc, err = vmc.processFs.NewProc(pid)
-	}
+	proc, err = vmc.processFs.NewProc(pid)
```
pjanotti: It seems that from `os.Getpid()` to here shouldn't happen on a timely basis, i.e. results could be saved and reused.

receiver/vmmetricsreceiver/vm_metrics_collector.go
```diff
@@ -113,104 +103,107 @@ func (vmc *VMMetricsCollector) StopCollection() {
...
+	return &metricspb.TimeSeries{
+		StartTimestamp: internal.TimeToTimestamp(vmc.startTime),
+		Points:         []*metricspb.Point{{Timestamp: internal.TimeToTimestamp(time.Now()), Value: &metricspb.Point_Int64Value{Int64Value: int64(val)}}},
 	}
-	if len(metrics) > 0 {
-		vmc.consumer.ConsumeMetricsData(ctx, data.MetricsData{Metrics: metrics})
+}
+
+func (vmc *VMMetricsCollector) getDoubleTimeSeries(val float64, labelVal *metricspb.LabelValue) *metricspb.TimeSeries {
+	labelVals := []*metricspb.LabelValue{}
```
pjanotti: nit: this slice escapes, but many times is used the empty one, could we have it as nil instead? Or does the `metricspb.TimeSeries` requires it to be non-nil?

receiver/vmmetricsreceiver/vm_metrics_collector.go
```diff
@@ -113,104 +103,107 @@ func (vmc *VMMetricsCollector) StopCollection() {
...
-			vd := view.Data{
-				View:  v,
-				Start: vmc.startTime,
-				End:   time.Now(),
-				Rows:  rows,
-			}
-			vds = append(vds, &vd)
-		}
+	if len(errs) > 0 {
+		span.SetStatus(trace.Status{Code: 15 /*DATA_LOSS*/, Message: fmt.Sprintf("Error(s) when scraping VM metrics: %v", errs)})
```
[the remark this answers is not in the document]
pjanotti: nit: `errs` with be show with a whitespace as separator depending on the messages it can be confuse. Suggestion: `internal.CombineErrors`

receiver/vmmetricsreceiver/vm_metrics_collector.go
songy23: Good suggestion, I moved it as part of `VMMetricsCollector` instead.

receiver/vmmetricsreceiver/vm_metrics_collector.go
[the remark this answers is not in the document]
songy23: SG, updated.

receiver/vmmetricsreceiver/vm_metrics_collector.go
songy23: You're right those don't need to be non-nil - updated.

PR #555, opened 2019-05-23
OC exporter: new configuration settings and use first available worker

pjanotti:
Adding the new configuration settings now exposed by the oc-agent exporter:

- keepalive
- reconnection delay

Plus an option, secure, to use system cert pool (useful for https with well-known CAs).

The workers for OC exporter were being used in a round-robin fashion that didn't take into account that if one was already available, this is beneficial in the case in which a send operation takes much longer than the others.

### DIFF
--- a/exporter/README.md
+++ b/exporter/README.md
@@ -85,7 +85,18 @@ queued-exporters:
     exporters:
       opencensus:
         endpoint: "127.0.0.1:55566"
+        cert-pem-file: "server_ca_public.pem" # PEM file used to enable TLS. For trusted CAs from
+                                              # system pool use "secure:" setting, see below.
         compression: "gzip"
+        reconnection-dealy: 2s # Delay (+70% jitter) before reconnection attempt in case of error.
+        secure: true # Used to export to destinations trusted by certificates from the system pool.
+                     # "cert-pem-file:" takes precedence over this setting.
+        keepalive:
+          # Keepalive settings for gRPC clients, see https://godoc.org/google.golang.org/grpc/keepalive#ClientParameters.
+          # Recommended to be set for cases that need to support bursts of data and periods of inactivity.
+          time: 30s
+          timeout: 5s
+          permit-without-stream: true
   my-org-jaeger: # A second processor with its own configuration options
     num-workers: 2
     queue-size: 100
--- a/exporter/README.md
+++ b/exporter/README.md
@@ -94,9 +94,9 @@ queued-exporters:
         keepalive:
           # Keepalive settings for gRPC clients, see https://godoc.org/google.golang.org/grpc/keepalive#ClientParameters.
           # Recommended to be set for cases that need to support bursts of data and periods of inactivity.
-          time: 30s
-          timeout: 5s
-          permit-without-stream: true
+          time: 30s # After inactive for this amount to time client will send a ping to the server.
+          timeout: 5s # Amount of time that client waits for a keepalive ping response before closing the connection.
+          permit-without-stream: true # Permits the keepalive ping even if no RPCs are in use, if false no ping is sent.
   my-org-jaeger: # A second processor with its own configuration options
     num-workers: 2
     queue-size: 100
--- a/exporter/opencensusexporter/opencensus.go
+++ b/exporter/opencensusexporter/opencensus.go
@@ -16,48 +16,73 @@ package opencensusexporter
 
 import (
 	"context"
-	"errors"
+	"crypto/x509"
 	"fmt"
-	"sync/atomic"
+	"sync"
+	"time"
 
 	"contrib.go.opencensus.io/exporter/ocagent"
 	agenttracepb "github.com/census-instrumentation/opencensus-proto/gen-go/agent/trace/v1"
 	"github.com/spf13/viper"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
 
 	"github.com/census-instrumentation/opencensus-service/consumer"
 	"github.com/census-instrumentation/opencensus-service/data"
 	"github.com/census-instrumentation/opencensus-service/exporter/exporterhelper"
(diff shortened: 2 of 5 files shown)


tigrannajaryan:
>The workers for OC exporter were being used in a round-robin fashion that didn't take into account that if one was already available, this is beneficial in the case in which a send operation takes much longer than the others

This unclear. Can you expand?

exporter/opencensusexporter/opencensus.go
```diff
@@ -129,22 +180,52 @@ func OpenCensusTraceExportersFromViper(v *viper.Viper) (tps []consumer.TraceCons
...
+	wg.Wait()
+
+	return internal.CombineErrors(errors)
+}
+
 func (oce *ocagentExporter) PushTraceData(ctx context.Context, td data.TraceData) (int, error) {
-	// Get an exporter worker round-robin
-	exporter := oce.exporters[atomic.AddUint32(&oce.counter, 1)%uint32(len(oce.exporters))]
+	// Get first available exporter.
+	exporter := <-oce.exporters
```
tigrannajaryan: I think we have a race condition here. If `stop()` is called after this line but before line 228, the `exporter` that we have just poped from `oce.exporters` will not be stopped.

pjanotti: that is not the case because `range` over channels just ends if the channel is closed but that never happens, so we leave the loop by counting the number of exporters (the capacity of the channel). Eventually this exporter is returned to the channel and the range picks it up to stop it.

exporter/README.md
```diff
@@ -85,7 +85,18 @@ queued-exporters:
...
+                                              # system pool use "secure:" setting, see below.
         compression: "gzip"
+        reconnection-dealy: 2s # Delay (+70% jitter) before reconnection attempt in case of error.
+        secure: true # Used to export to destinations trusted by certificates from the system pool.
+                     # "cert-pem-file:" takes precedence over this setting.
+        keepalive:
+          # Keepalive settings for gRPC clients, see https://godoc.org/google.golang.org/grpc/keepalive#ClientParameters.
+          # Recommended to be set for cases that need to support bursts of data and periods of inactivity.
+          time: 30s
+          timeout: 5s
```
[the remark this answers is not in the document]
pjanotti: I will expand a bit the comments in this section, I don't want to change the names themselves to avoid one having to map from gRPC docs to our names.

exporter/opencensusexporter/opencensus.go
tigrannajaryan: OK, makes sense, I didn't notice the counting logic.

How about the situation when stop() finishes and PushTraceData() tries to read from `oce.exporters`? It will get stuck because the channel is open but there is nothing to read from it.

exporter/README.md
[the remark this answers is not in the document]
tigrannajaryan: ok

exporter/opencensusexporter/opencensus.go
pjanotti: It will be stuck but this doesn't prevent shutdown of the app, but, I'm not considering the future scenario in which we may want to shutdown only the pipeline. We should define some behavior here (it seems more natural to require the workers to stop submitting data before asking for a stop of the exporter), that said, returning an error in this case is the correct approach, I will add that to the code.

pjanotti:
> The workers for OC exporter were being used in a round-robin fashion that didn't take into account that if one was already available, this is beneficial in the case in which a send operation takes much longer than the others

Previously an attempt to send data was going to wait on a specific worker to get released even if all others were idle, now, it picks the first one that is available.

exporter/opencensusexporter/opencensus.go
```diff
@@ -129,22 +182,61 @@ func OpenCensusTraceExportersFromViper(v *viper.Viper) (tps []consumer.TraceCons
...
+	}
+
 	err := exporter.ExportTraceServiceRequest(
 		&agenttracepb.ExportTraceServiceRequest{
 			Spans:    td.Spans,
 			Resource: td.Resource,
 			Node:     td.Node,
 		},
 	)
+	oce.exporters <- exporter
```
tigrannajaryan: If I remember correctly writing on closed channel panics.

If you can guarantee that send() and PushTraceData() are never called concurrently that would be the best way instead of trying to make them safe to call concurrently, just add comments explaining that it is unsafe.

pjanotti: that is correct, but the channel is only closed when all exporters were stopped at that moment there won't be any exporter in transition to be added back to the channel. If it was retrieved from the channel here it was not stopped yet so the channel was not closed yet.

tigrannajaryan: :+1:

exporter/README.md
```diff
@@ -85,7 +85,18 @@ queued-exporters:
...
+        cert-pem-file: "server_ca_public.pem" # PEM file used to enable TLS. For trusted CAs from
+                                              # system pool use "secure:" setting, see below.
         compression: "gzip"
+        reconnection-dealy: 2s # Delay (+70% jitter) before reconnection attempt in case of error.
+        secure: true # Used to export to destinations trusted by certificates from the system pool.
+                     # "cert-pem-file:" takes precedence over this setting.
+        keepalive:
+          # Keepalive settings for gRPC clients, see https://godoc.org/google.golang.org/grpc/keepalive#ClientParameters.
+          # Recommended to be set for cases that need to support bursts of data and periods of inactivity.
+          time: 30s # After inactive for this amount to time client will send a ping to the server.
```
songy23: Suggestion: consider `interval` or `period`.

pjanotti: I also don't like `time` very much (`interval` is better to me) but I'm following the names at https://godoc.org/google.golang.org/grpc/keepalive#ClientParameters so users can read that doc and directly map to the configuration items, if we start to diverge one needs to mentally map them.

songy23: Make sense to me.

PR #557, opened 2019-05-23
Implement V2 configuration loading

tigrannajaryan:
Implemented reading of generic parts of V2 configuration from Viper and
validation of the configuration.

Added stretchr/testify library to be able to use helpers like assert() in tests.

Testing done: automated tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,6 +44,7 @@ require (
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/viper v1.2.1
 	github.com/streadway/quantile v0.0.0-20150917103942-b0c588724d25 // indirect
+	github.com/stretchr/testify v1.3.0
 	github.com/tinylib/msgp v1.1.0 // indirect
 	github.com/uber-go/atomic v1.3.2 // indirect
 	github.com/uber/jaeger-client-go v2.16.0+incompatible // indirect
--- a/internal/configv2/configv2.go
+++ b/internal/configv2/configv2.go
@@ -63,21 +63,25 @@ func (e *configError) Error() string {
 	return e.msg
 }
 
-// receiversKeyName is the configuration key name for receivers section.
-const receiversKeyName = "receivers"
+// YAML top-level configuration keys
+const (
+	// receiversKeyName is the configuration key name for receivers section.
+	receiversKeyName = "receivers"
 
-// exportersKeyName is the configuration key name for exporters section.
-const exportersKeyName = "exporters"
+	// exportersKeyName is the configuration key name for exporters section.
+	exportersKeyName = "exporters"
 
-// processorsKeyName is the configuration key name for processors section.
-const processorsKeyName = "processors"
+	// processorsKeyName is the configuration key name for processors section.
+	processorsKeyName = "processors"
 
-// pipelinesKeyName is the configuration key name for pipelines section.
-const pipelinesKeyName = "pipelines"
+	// pipelinesKeyName is the configuration key name for pipelines section.
+	pipelinesKeyName = "pipelines"
+)
 
 // typeAndNameSeparator is the separator that is used between type and name in type/name composite keys.
 const typeAndNameSeparator = "/"
 
+// Data type strings
 const (
 	tracesDataTypeStr  = "traces"
 	metricsDataTypeStr = "metrics"
@@ -128,16 +132,13 @@ func Load(v *viper.Viper) (*configmodels.ConfigV2, error) {
 // The "type" part must be present, the forward slash and "name" are optional.
 func decodeTypeAndName(key string) (typeStr, fullName string, err error) {
 	items := strings.SplitN(key, typeAndNameSeparator, 2)
-	if len(items) < 1 {
-		// We need at least one part (the type).
-		err = errors.New("Invalid key")
-		return
+
+	if len(items) >= 1 {
+		typeStr = strings.TrimSpace(items[0])
 	}
 
-	typeStr = strings.TrimSpace(items[0])
-	if typeStr == "" {
-		// Type must be present.
-		err = errors.New("Invalid key")
+	if len(items) < 1 || typeStr == "" {
+		err = errors.New("type/name key must have the type part")
 		return
 	}
 
@@ -146,7 +147,7 @@ func decodeTypeAndName(key string) (typeStr, fullName string, err error) {
 		// "name" part is present.
 		nameSuffix = strings.TrimSpace(items[1])
 		if nameSuffix == "" {
-			err = errors.New("Invalid key")
+			err = errors.New("name part must be specified after " + typeAndNameSeparator + " in type/name key")
 			return
 		}
 	} else {
@@ -181,7 +182,7 @@ func loadReceivers(v *viper.Viper) (configmodels.Receivers, error) {
 		if err != nil || typeStr == "" {
 			return nil, &configError{
 				code: errInvalidTypeAndNameKey,
-				msg:  fmt.Sprintf("invalid key %q. Key must in the form type[/name]", key),
+				msg:  fmt.Sprintf("invalid key %q: %s", key, err.Error()),
 			}
 		}
 
@@ -236,7 +237,7 @@ func loadExporters(v *viper.Viper) (configmodels.Exporters, error) {
 		if err != nil || typeStr == "" {
 			return nil, &configError{
 				code: errInvalidTypeAndNameKey,
-				msg:  fmt.Sprintf("invalid key %q. Key must in the form type[/name]", key),
+				msg:  fmt.Sprintf("invalid key %q: %s", key, err.Error()),
 			}
 		}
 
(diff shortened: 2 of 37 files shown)


go.mod
```diff
@@ -44,6 +44,7 @@ require (
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/viper v1.2.1
 	github.com/streadway/quantile v0.0.0-20150917103942-b0c588724d25 // indirect
+	github.com/stretchr/testify v1.3.0
```
bogdandrutu: Why do we add a new dependency?

Makefile
```diff
@@ -73,14 +73,7 @@ lint:
 
 .PHONY: vet
 vet:
-	@VETOUT=`$(GOVET) ./... 2>&1`; \
```
bogdandrutu: Why this in the current PR?

go.mod
tigrannajaryan: It provides very handy helper functions for tests, like assert.Equal, e.g.: https://github.com/census-instrumentation/opencensus-service/pull/557/files#diff-ef311f1bc51e5169d51cfb73bc922052R181

Makefile
tigrannajaryan: From PR description:

>Fixed "vet" target in makefile. Previously it was coded to fail if govet produced
any output, which is not a correct assumption. Govet can produce output even in
case of success [1]. The correct way to know if govet is successful is to rely on exit
code, which is exactly what is done now after this change.
[1] - see for example https://travis-ci.org/census-instrumentation/opencensus-service/builds/536550187

bogdandrutu: Please separate in a different PR :)

go.mod
bogdandrutu: Please file an issue on the opentelemetry-go and service to use this then :) If it is that nice :)

Makefile
tigrannajaryan: Sure, let me do it.

go.mod
tigrannajaryan: Yes, I think it is pretty useful and makes tests cleaner and shorter :-)

Do you want me to open issues for these 2 repos? 
https://github.com/open-telemetry/opentelemetry-go
https://github.com/open-telemetry/opentelemetry-service

Makefile
tigrannajaryan: I created a separate PR: https://github.com/census-instrumentation/opencensus-service/pull/558

Please review.

go.mod
bogdandrutu: Yes, just to make sure we use this from the beginning in the new project

tigrannajaryan: Done:
https://github.com/open-telemetry/opentelemetry-go/issues/3
https://github.com/open-telemetry/opentelemetry-service/issues/7

tigrannajaryan:
@bogdandrutu do you have other comments on this PR?

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+}
+
+// decodeTypeAndName decodes a key in type[/name] format into type and fullName.
+// fullName is the key normalized such that type and name components have spaces trimmed.
+// The "type" part must be present, the forward slash and "name" are optional.
+func decodeTypeAndName(key string) (typeStr, fullName string, err error) {
+	items := strings.SplitN(key, typeAndNameSeparator, 2)
+	if len(items) < 1 {
+		// We need at least one part (the type).
+		err = errors.New("Invalid key")
```
pjanotti: nit: the comment has better information than the error message.

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+		return
+	}
+
+	typeStr = strings.TrimSpace(items[0])
+	if typeStr == "" {
+		// Type must be present.
+		err = errors.New("Invalid key")
```
pjanotti: ditto

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+		return
+	}
+
+	var nameSuffix string
+	if len(items) > 1 {
+		// "name" part is present.
+		nameSuffix = strings.TrimSpace(items[1])
+		if nameSuffix == "" {
+			err = errors.New("Invalid key")
```
pjanotti: what about "missing name part"?

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+	receivers := make(configmodels.Receivers, 0)
+
+	// Iterate over input map and create a config for each.
+	for key := range keyMap {
+		// Decode the key into type and fullName components.
+		typeStr, fullName, err := decodeTypeAndName(key)
+		if err != nil || typeStr == "" {
+			return nil, &configError{
+				code: errInvalidTypeAndNameKey,
+				msg:  fmt.Sprintf("invalid key %q. Key must in the form type[/name]", key),
```
pjanotti: if you enhance the errors as mentioned above, it seems worth to include err in the message. Consider also the case that someone later may add some other error message to decodeTypeAndName

Notice that there is no case in which `typeStr == ""` and `err == nil`

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+	exporters := make(configmodels.Exporters, 0)
+
+	// Iterate over exporters and create a config for each.
+	for key := range keyMap {
+		// Decode the key into type and fullName components.
+		typeStr, fullName, err := decodeTypeAndName(key)
+		if err != nil || typeStr == "" {
+			return nil, &configError{
+				code: errInvalidTypeAndNameKey,
+				msg:  fmt.Sprintf("invalid key %q. Key must in the form type[/name]", key),
```
pjanotti: ditto: nit: add err to msg.

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+	pipelines := make(configmodels.Pipelines, 0)
+
+	// Iterate over input map and create a config for each.
+	for key := range keyMap {
+		// Decode the key into type and name components.
+		typeStr, name, err := decodeTypeAndName(key)
+		if err != nil || typeStr == "" {
+			return nil, &configError{
+				code: errInvalidTypeAndNameKey,
+				msg:  fmt.Sprintf("invalid key %q. Key must in the form type[/name]", key),
```
pjanotti: ditto nit

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+// processorsKeyName is the configuration key name for processors section.
+const processorsKeyName = "processors"
+
+// pipelinesKeyName is the configuration key name for pipelines section.
+const pipelinesKeyName = "pipelines"
+
+// typeAndNameSeparator is the separator that is used between type and name in type/name composite keys.
+const typeAndNameSeparator = "/"
+
+const (
```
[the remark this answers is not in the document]
tigrannajaryan: I will group into logical groups.

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+
+	// Create normalized fullName.
+	if nameSuffix == "" {
+		fullName = typeStr
+	} else {
+		fullName = typeStr + typeAndNameSeparator + nameSuffix
+	}
+
+	err = nil
```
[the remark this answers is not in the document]
tigrannajaryan: Hmm, seems pretty idiomatic to me, e.g.: https://tour.golang.org/basics/7

internal/configv2/configv2.go
tigrannajaryan: Fixed.

internal/configv2/configv2.go
tigrannajaryan: Fixed.

internal/configv2/configv2.go
tigrannajaryan: Fixed.

internal/configv2/configv2.go
tigrannajaryan: Added to error message.

internal/configv2/configv2.go
tigrannajaryan: Done.

internal/configv2/configv2.go
```diff
@@ -0,0 +1,511 @@
...
+	processors := make(configmodels.Processors, 0)
+
+	// Iterate over processors and create a config for each.
+	for key := range keyMap {
+		// Decode the key into type and fullName components.
+		typeStr, fullName, err := decodeTypeAndName(key)
+		if err != nil || typeStr == "" {
+			return nil, &configError{
+				code: errInvalidTypeAndNameKey,
+				msg:  fmt.Sprintf("invalid key %q. Key must in the form type[/name]", key),
```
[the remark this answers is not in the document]
tigrannajaryan: Done.

internal/configv2/configv2.go
tigrannajaryan: Done.

tigrannajaryan:
@pjanotti I fixed almost everything (only one open comment remaining). Please have a look again.

internal/configv2/configv2.go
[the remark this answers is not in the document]
pjanotti: naked return yes, use of named return values too, the call `err = nil` that is kind of strange to me since it wasn't set to any other value above if it reaches this point of the code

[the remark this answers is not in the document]
tigrannajaryan: The alternate is to use unnamed return `return fullName, typeStr, err` however I find this error-prone because I can equally easy make a mistake and write `return typeStr, fullName, err`. The compiler will not complain but it is completely wrong. For this reason I prefer to use named returns in scenarios when I have more than one return parameter with the same type (like here). 

This may look slightly unconventional but it is safer.

tigrannajaryan:
@pjanotti please merge this.

PR #558, opened 2019-05-24
Fix "vet" target in makefile

tigrannajaryan:
Previously "vet" target was coded to fail if govet produced any output,
which is not a correct assumption. Govet can produce output even in
case of success [1]. The correct way to know if govet is successful is
to rely on exit code, which is exactly what is done now after this change.

[1] - see for example https://travis-ci.org/census-instrumentation/opencensus-service/builds/536550187

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,14 +73,8 @@ lint:
 
 .PHONY: vet
 vet:
-	@VETOUT=`$(GOVET) ./... 2>&1`; \
-	if [ "$$VETOUT" ]; then \
-		echo "$(GOVET) FAILED => clean the following vet errors:\n"; \
-		echo "$$VETOUT\n"; \
-		exit 1; \
-	else \
-	    echo "Vet finished successfully"; \
-	fi
+	@$(GOVET) ./...
+	@echo "Vet finished successfully"
 
 .PHONY: install-tools
 install-tools:


tigrannajaryan:
@pjanotti @bogdandrutu please review. The build failed because codecov flagged it as reduced coverage which I believe isn't true since I didn't touch Go code at all. I hope you can override codecov and merge regardless.

pjanotti:
q.: did you commit the two projects below to go.mod? 
```
go: downloading github.com/stretchr/testify v1.3.0
go: downloading github.com/pmezard/go-difflib v1.0.0
```

pjanotti:
btw, I can't override codecov/project (I've being paying coverage tax here and there to keep things going forward)

tigrannajaryan:
>q.: did you commit the two projects below to go.mod?

@pjanotti No, the only change in this PR is makefile.

tigrannajaryan:
>btw, I can't override codecov/project (I've being paying coverage tax here and there to keep things going forward)

@pjanotti  hmm, I am not sure what to do, my change has nothing to do with coverage. It would be strange to touch go code just to make codecov happy when my change is unrelated. Perhaps we can rerun codecov and it will behave properly this time?

tigrannajaryan:
OK, I added a cosmetic change to re-trigger the build, hope codecov passes this time. I looks like a codecov bug.

tigrannajaryan:
@pjanotti codecov passed this time, please have a look.

PR #560, opened 2019-05-29
Add grpc server configs to OC receiver

pjanotti:
This will be only applied to the collector since it is assumed that unlike the agent it is typically located behind some loadbalancer or proxy where these settings are more useful.


cmd/occollector/app/builder/builder.go
```diff
@@ -106,6 +107,37 @@ type OpenCensusReceiverCfg struct {
...
+}
+
+type serverParametersAndEnforcementPolicy struct {
+	ServerParameters  *keepaliveServerParameters  `mapstructure:"server-parameters,omitempty"`
+	EnforcementPolicy *keepaliveEnforcementPolicy `mapstructure:"enforcement-policy,omitempty"`
+}
+
+// keepaliveServerParameters allow configuration of the keepalive.ServerParameters.
+// See https://godoc.org/google.golang.org/grpc/keepalive#ServerParameters for details.
+type keepaliveServerParameters struct {
```
tigrannajaryan: Do we have a use case for allowing end user to configure all these options? I would be cautious and only expose what end users really need to be able to change. I am worried about proliferation of configuration options.

cmd/occollector/app/builder/builder.go
```diff
@@ -106,6 +107,37 @@ type OpenCensusReceiverCfg struct {
...
+	MaxConnectionIdle     time.Duration `mapstructure:"max-connection-idle,omitempty"`
+	MaxConnectionAge      time.Duration `mapstructure:"max-connection-age,omitempty"`
+	MaxConnectionAgeGrace time.Duration `mapstructure:"max-connection-age-grace,omitempty"`
+	Time                  time.Duration `mapstructure:"time,omitempty"`
+	Timeout               time.Duration `mapstructure:"timeout,omitempty"`
+}
+
+// keepaliveEnforcementPolicy allow configuration of the keepalive.EnforcementPolicy.
+// See https://godoc.org/google.golang.org/grpc/keepalive#EnforcementPolicy for details.
+type keepaliveEnforcementPolicy struct {
```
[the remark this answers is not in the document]
pjanotti: - Enforcement policy is for burst scenarios that you want to keep connections alive. 
- The server settings for keepalive are for the settings that can help with load balancing (recommended by gRPC).
- Max receive msg size is the only way to make your server receive something larger, notice that the client by default gas int32.Max as its maximum size (worst this loss can be silent)
- Max concurrent connections is a lever that can be used how many streams a server handles that also can help with load-balancing in certain cases (depends on the LB)

The problem of limiting these to a smaller sub-set is that then we require code changes to be able to use them, in one way or another I have used all of these (not at the same time).

cmd/occollector/app/builder/builder.go
pjanotti: I have all of those in use in different scenarios, granted that many users may do fine without them but we need to make these levers available.

PR #562, opened 2019-05-31
Add implementation of OpenCensus receiver config

tigrannajaryan:
- Added factory and config loading for OpenCensus receiver.

- Added support for custom unmarshaling of receiver configuration.

- Added ability for factory to create trace and metric receivers
  based on configuration. This will be later required when building
  the pipeline based on the config.

- Moved example factories used in tests to example_factories.go to
  make it easier to reuse them in other tests.

Testing done: automated tests

### DIFF
--- a/internal/configv2/test_helpers.go
+++ b/internal/configv2/test_helpers.go
@@ -1,3 +1,17 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package configv2
 
 import (
--- a/internal/configv2/test_helpers.go
+++ b/internal/configv2/test_helpers.go
@@ -1,0 +1,31 @@
+package configv2
+
+import (
+	"os"
+	"testing"
+
+	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
+	"github.com/spf13/viper"
+)
+
+// LoadConfigFile loads a config from file.
+func LoadConfigFile(t *testing.T, fileName string) (*configmodels.ConfigV2, error) {
+	// Open the file for reading.
+	file, err := os.Open(fileName)
+	if err != nil {
+		t.Error(err)
+		return nil, err
+	}
+
+	// Read yaml config from file
+	v := viper.New()
+	v.SetConfigType("yaml")
+	err = v.ReadConfig(file)
+	if err != nil {
+		t.Errorf("unable to read yaml, %v", err)
+		return nil, err
+	}
+
+	// Load the config from viper
+	return Load(v)
+}
--- a/internal/factories/factories.go
+++ b/internal/factories/factories.go
@@ -15,7 +15,13 @@
 package factories
 
 import (
+	"context"
 	"fmt"
+
+	"github.com/census-instrumentation/opencensus-service/consumer"
+	"github.com/census-instrumentation/opencensus-service/receiver"
+
+	"github.com/spf13/viper"
 
 	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
 )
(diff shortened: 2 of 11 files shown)


tigrannajaryan:
@pjanotti @bogdandrutu please review.

internal/configv2/test_helpers.go
```diff
@@ -0,0 +1,31 @@
+package configv2
```
pjanotti: missing header

receiver/opencensusreceiver/factory.go
```diff
@@ -0,0 +1,116 @@
...
+
+var _ = factories.RegisterReceiverFactory(&ReceiverFactory{})
+
+const (
+	// The value of "type" key in configuration.
+	typeStr = "opencensus"
+)
+
+// ReceiverFactory is the factory for receiver.
+type ReceiverFactory struct {
```
pjanotti: q.: it is already registered, why it needs to be public?

internal/factories/factories.go
```diff
@@ -31,8 +37,29 @@ type ReceiverFactory interface {
...
 	CreateDefaultConfig() configmodels.Receiver
+
+	// CustomUnmarshaler returns a custom unmarshaler for the configuration or nil if
+	// there is no need for custom unmarshaling. This is typically used if viper.Unmarshal()
+	// is not sufficient to unmarshal correctly.
+	CustomUnmarshaler() CustomUnmarshaler
+
+	// CreateTraceReceiver creates a trace receiver based on this config.
+	// If the receiver type does not support tracing or if the config is not valid
+	// error will be returned instead (ErrReceiverTypeIsNotSupported).
```
pjanotti: perhaps, I missed but I didn't see this error defined

receiver/opencensusreceiver/factory.go
```diff
@@ -0,0 +1,116 @@
...
+// CustomUnmarshaler returns nil because we don't need custom unmarshaling for this config.
+func (f *ReceiverFactory) CustomUnmarshaler() factories.CustomUnmarshaler {
+	return nil
+}
+
+// CreateDefaultConfig creates the default configuration for receiver.
+func (f *ReceiverFactory) CreateDefaultConfig() configmodels.Receiver {
+	return &ConfigV2{
+		ReceiverSettings: configmodels.ReceiverSettings{
+			Endpoint: "127.0.0.1:55678",
```
pjanotti: just a reminder for down the road: for the collector the default should be ":55678" (via the file defining the settings for the collector)

receiver/opencensusreceiver/factory.go
```diff
@@ -0,0 +1,116 @@
...
+		receivers[rCfg] = receiver
+	}
+	return receiver, nil
+}
+
+// This is the map of already created OpenCensus receivers for particular configurations.
+// We maintain this map because the factory is asked trace and metric receivers separately
+// when it gets CreateTraceReceiver() and CreateMetricsReceiver() but they must not
+// create separate objects, they must use one Receiver object per configuration.
+var receivers = map[*ConfigV2]*Receiver{}
```
pjanotti: q.: so the caller code of Create needs to ensure that the same config object is used, how that is going to be enforced?

receiver/opencensusreceiver/factory.go
tigrannajaryan: It doesn't. I will change.

internal/factories/factories.go
tigrannajaryan: Wrong comment, let me fix.

receiver/opencensusreceiver/factory.go
tigrannajaryan: I believe the proper way is to simply package the collector with its own default config file. So here we define some reasonable defaults which make sense both for Agent and Collector and any deviations should be reflected in the default config file that comes with the distribution. We will need to decide how exactly this packaging is done when we think about distribution and releases.

receiver/opencensusreceiver/factory.go
tigrannajaryan: It doesn't need to be enforced. The idea is that _if_ the caller (the pipeline builder) attempts to create a receiver using the same config instance the factory will end up returning the same receiver instance which is the desirable effect.

internal/configv2/test_helpers.go
tigrannajaryan: Fixed.

receiver/opencensusreceiver/factory.go
tigrannajaryan: Fixed.

internal/factories/factories.go
tigrannajaryan: Fixed.

tigrannajaryan:
@pjanotti please have a look one more time.

PR #564, opened 2019-06-04
Add implementation of OpenCensus exporter config

tigrannajaryan:
- Added factory and config loading for OpenCensus exporter.

Note: the factory is not compelte yet. We will need to add ability to
create the exporter based on its config. It will be done in a future PR.

Testing done: automated tests

### DIFF
--- a/exporter/opencensusexporter/config.go
+++ b/exporter/opencensusexporter/config.go
@@ -15,6 +15,8 @@
 package opencensusexporter
 
 import (
+	"time"
+
 	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
 )
 
@@ -26,4 +28,7 @@ type ConfigV2 struct {
 	Headers                       map[string]string        `mapstructure:"headers"`
 	NumWorkers                    int                      `mapstructure:"num-workers"`
 	CertPemFile                   string                   `mapstructure:"cert-pem-file"`
+	UseSecure                     bool                     `mapstructure:"secure,omitempty"`
+	ReconnectionDelay             time.Duration            `mapstructure:"reconnection-delay,omitempty"`
+	KeepaliveParameters           *keepaliveConfig         `mapstructure:"keepalive,omitempty"`
 }
--- a/exporter/opencensusexporter/config.go
+++ b/exporter/opencensusexporter/config.go
@@ -1,0 +1,29 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package opencensusexporter
+
+import (
+	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
+)
+
+// ConfigV2 defines configuration for OpenCensus exporter.
+type ConfigV2 struct {
+	configmodels.ExporterSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
+	Endpoint                      string                   `mapstructure:"endpoint"`
+	Compression                   string                   `mapstructure:"compression"`
+	Headers                       map[string]string        `mapstructure:"headers"`
+	NumWorkers                    int                      `mapstructure:"num-workers"`
+	CertPemFile                   string                   `mapstructure:"cert-pem-file"`
+}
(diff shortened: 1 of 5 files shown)


tigrannajaryan:
@pjanotti @bogdandrutu please review.

exporter/opencensusexporter/factory.go
```diff
@@ -0,0 +1,44 @@
...
+	typeStr = "opencensus"
+)
+
+// ExporterFactory is the factory for OpenCensus exporter.
+type ExporterFactory struct {
+}
+
+// Type gets the type of the Exporter config created by this factory.
+func (f *ExporterFactory) Type() string {
+	return typeStr
```
songy23: Nit: currently type is static, should we make it configurable?

exporter/opencensusexporter/factory.go
```diff
@@ -0,0 +1,44 @@
...
+
+var _ = factories.RegisterExporterFactory(&ExporterFactory{})
+
+const (
+	// The value of "type" key in configuration.
+	typeStr = "opencensus"
+)
+
+// ExporterFactory is the factory for OpenCensus exporter.
+type ExporterFactory struct {
```
pjanotti: why public?

exporter/opencensusexporter/testdata/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+exporters:
+  opencensus:
+  opencensus/2:
+    enabled: true
+    endpoint: "1.2.3.4:1234"
+    compression: "on"
+    num-workers: 123
+    cert-pem-file: /var/lib/mycert.pem
+    headers:
+      header1: 234
```
pjanotti: using `{` and `}` syntax can you double-check that you can have headers with `.`? it should work but let's double check it.

exporter/opencensusexporter/config.go
```diff
@@ -0,0 +1,29 @@
...
+package opencensusexporter
+
+import (
+	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
+)
+
+// ConfigV2 defines configuration for OpenCensus exporter.
+type ConfigV2 struct {
+	configmodels.ExporterSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
+	Endpoint                      string                   `mapstructure:"endpoint"`
```
pjanotti: We have more settings than these, please add:
```golang
	UseSecure           bool              `mapstructure:"secure,omitempty"`
	ReconnectionDelay   time.Duration     `mapstructure:"reconnection-delay,omitempty"`
	KeepaliveParameters *keepaliveConfig  `mapstructure:"keepalive,omitempty"`
```

pjanotti: While we can't freeze the current config hopefully the changes will be few (if any).

exporter/opencensusexporter/testdata/config.yaml
pjanotti: If it doesn't no need to fix it now, but then let's open an issue to track it.

exporter/opencensusexporter/factory.go
tigrannajaryan: Shouldn't be. Fixing.

exporter/opencensusexporter/testdata/config.yaml
tigrannajaryan: Sorry, I didn't get that. Can you expand?

exporter/opencensusexporter/factory.go
tigrannajaryan: The type is intended to be static (const), it is what defines the exporter and the format of the config file and the particular key that is used in the config file to refer to this exporter type.

exporter/opencensusexporter/config.go
tigrannajaryan: Looks like it evolved since I wrote this code :-) Adding.

exporter/opencensusexporter/factory.go
tigrannajaryan: Fixed.

exporter/opencensusexporter/config.go
tigrannajaryan: Added.

exporter/opencensusexporter/testdata/config.yaml
pjanotti: if you use ```yaml
   headers: { "can you have a . here?":"F0000000-0000-0000-0000-000000000000" }

tigrannajaryan: I see what you mean now. You don't need {}, a quoted string can be a key in YAML. Let me add an example to the test.

tigrannajaryan: Added, please have a look.

exporter/opencensusexporter/factory.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2018, OpenCensus Authors
```
pjanotti: nit: 2019

exporter/opencensusexporter/testdata/config.yaml
pjanotti: nice

PR #565, opened 2019-06-05
Add implementation of Jaeger receiver config

tigrannajaryan:
- Added factory and config loading for Jaeger receiver.
- Added ErrDataTypeIsNotSupported to factory definitions.

Testing done: automated tests

### DIFF
--- a/internal/factories/factories.go
+++ b/internal/factories/factories.go
@@ -16,14 +16,14 @@ package factories
 
 import (
 	"context"
+	"errors"
 	"fmt"
-
-	"github.com/census-instrumentation/opencensus-service/consumer"
-	"github.com/census-instrumentation/opencensus-service/receiver"
 
 	"github.com/spf13/viper"
 
+	"github.com/census-instrumentation/opencensus-service/consumer"
 	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
+	"github.com/census-instrumentation/opencensus-service/receiver"
 )
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -55,6 +55,11 @@ type ReceiverFactory interface {
 	CreateMetricsReceiver(cfg configmodels.Receiver,
 		consumer consumer.MetricsConsumer) (receiver.MetricsReceiver, error)
 }
+
+// ErrDataTypeIsNotSupported can be returned by CreateTraceReceiver or
+// CreateMetricsReceiver if the particular telemetry data type is not supported
+// by the receiver.
+var ErrDataTypeIsNotSupported = errors.New("telemetry type is not supported")
 
 // CustomUnmarshaler is a function that un-marshals a viper data into a config struct
 // in a custom way.
--- a/receiver/jaegerreceiver/config_test.go
+++ b/receiver/jaegerreceiver/config_test.go
@@ -47,11 +47,11 @@ func TestLoadConfig(t *testing.T) {
 			Protocols: map[string]*configmodels.ReceiverSettings{
 				"thrift-http": {
 					Enabled:  false,
-					Endpoint: "example.com:3456",
+					Endpoint: "127.0.0.1:3456",
 				},
 				"thrift-tchannel": {
 					Enabled:  true,
-					Endpoint: "omnition.com:123",
+					Endpoint: "0.0.0.0:123",
 				},
 			},
 		})
--- a/receiver/jaegerreceiver/config_test.go
+++ b/receiver/jaegerreceiver/config_test.go
@@ -1,0 +1,58 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package jaegerreceiver
+
+import (
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
+	"github.com/census-instrumentation/opencensus-service/internal/configv2"
+	"github.com/census-instrumentation/opencensus-service/internal/factories"
+)
+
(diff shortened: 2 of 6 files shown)


internal/factories/factories.go
```diff
@@ -56,6 +56,11 @@ type ReceiverFactory interface {
 		consumer consumer.MetricsConsumer) (receiver.MetricsReceiver, error)
 }
 
+// ErrDataTypeIsNotSupported can be returned by CreateTraceReceiver or
+// CreateMetricsReceiver if the particular data type is not supported by
+// the receiver.
+var ErrDataTypeIsNotSupported = errors.New("data type is not supported")
```
songy23: Nit: I think "telemetry type" may be clearer here.

receiver/jaegerreceiver/factory.go
```diff
@@ -0,0 +1,125 @@
...
+	config := Configuration{}
+
+	// Set ports
+	if protoHTTP != nil {
+		config.CollectorHTTPPort = exatractPortFromEndpoint(protoHTTP.Endpoint)
+	}
+	if protoTChannel != nil {
+		config.CollectorHTTPPort = exatractPortFromEndpoint(protoTChannel.Endpoint)
+	}
+
```
songy23: Should there be a check that at least one of HTTP or TChannel is enabled?

internal/factories/factories.go
tigrannajaryan: Good point, I will change.

receiver/jaegerreceiver/factory.go
tigrannajaryan: Thanks, your comment prompted me to add the config error checks and corresponding tests. Please check again.

internal/factories/factories.go
```diff
@@ -56,6 +56,11 @@ type ReceiverFactory interface {
 		consumer consumer.MetricsConsumer) (receiver.MetricsReceiver, error)
 }
 
+// ErrDataTypeIsNotSupported can be returned by CreateTraceReceiver or
```
pjanotti: nit: it will be good if the message states what is the telemetry type not supported ("metrics", "traces", "logs"). 

Related q: I assume it is public so each factory can use it but is it expected that anyone will handle this error?

receiver/jaegerreceiver/factory.go
```diff
@@ -0,0 +1,147 @@
...
+const (
+	// The value of "type" key in configuration.
+	typeStr = "jaeger"
+
+	// Protocol values.
+	protoThriftHTTP     = "thrift-http"
+	protoThriftTChannel = "thrift-tchannel"
+
+	// Default endpoints to bind to.
+	defaultHTTPBindEndpoint     = "127.0.0.1:14268"
```
pjanotti: hum, I'm not sure that for jaeger these defaults make sense: these are the endpoints for jaeger collector that isn't expected to be on the same network namespace as the clients. The agent in that case should have one of the UDP protocols instead for jaeger (but that is broken at this very moment). I don't think we need to resolve/change this now but c uld you please open an issue to update/change this later?

pjanotti: Or perhaps it makes even more sense to separate the jaeger receiver into jaeger-collector and jaeger-agent

receiver/jaegerreceiver/config_test.go
```diff
@@ -0,0 +1,58 @@
...
+	r0 := config.Receivers["jaeger"]
+	assert.Equal(t, r0, factory.CreateDefaultConfig())
+
+	r1 := config.Receivers["jaeger/customname"].(*ConfigV2)
+	assert.Equal(t, r1,
+		&ConfigV2{
+			Protocols: map[string]*configmodels.ReceiverSettings{
+				"thrift-http": {
+					Enabled:  false,
+					Endpoint: "example.com:3456",
```
pjanotti: nit: this is just a test but the endpoint doesn't make much sense, it should be a local address or empty to indicate all available network interfaces.

receiver/jaegerreceiver/factory.go
```diff
@@ -0,0 +1,147 @@
...
+
+	protoHTTP := rCfg.Protocols[protoThriftHTTP]
+	protoTChannel := rCfg.Protocols[protoThriftTChannel]
+
+	config := Configuration{}
+
+	// Set ports
+	if protoHTTP != nil {
+		var err error
+		config.CollectorHTTPPort, err = exatractPortFromEndpoint(protoHTTP.Endpoint)
```
pjanotti: do not implement your own function to extract port, use https://golang.org/pkg/net/#SplitHostPort instead (or at least call it to do the bulk of the work in a wrapper).

receiver/jaegerreceiver/testdata/config.yaml
```diff
@@ -0,0 +1,22 @@
+receivers:
+  jaeger:
+  jaeger/customname:
+    protocols:
+      thrift-http:
+        endpoint: "example.com:3456"
```
pjanotti: nit: given that the host part is ignored do not use it

internal/factories/factories.go
tigrannajaryan: I agree, I left it out because it was not yet clear to me how exactly the error message will be used. I will add it as soon as it is clear that the message is going to be visible to the end user (there is a chance the  the error message for end user will be formatted elsewhere and we will not need it here, so that's why I am delaying).

receiver/jaegerreceiver/factory.go
tigrannajaryan: What is the alternate? "0.0.0.0:14268"? Again, let's keep in mind these are only some sane defaults which will likely be overridden by a default config _file_ that will be included in the distribution. This stays as an open question that I would like to address in due time.

receiver/jaegerreceiver/config_test.go
tigrannajaryan: Yeah, for the purpose of the particular test it doesn't matter. I can change to avoid confusion.

receiver/jaegerreceiver/factory.go
tigrannajaryan: Ahh, thanks, that's useful!

receiver/jaegerreceiver/factory.go
pjanotti: "0.0.0.0:14268" will work, but in golang people usually simply write ":14268"

internal/factories/factories.go
pjanotti: sgtm

receiver/jaegerreceiver/testdata/config.yaml
tigrannajaryan: I think it is still worth using in the test to verify parsing (since the parsing code is present).

tigrannajaryan:
@pjanotti I made fixes. Please review again.

PR #566, opened 2019-06-06
Add implementation of Attributes processor config

tigrannajaryan:
- Added factory and config loading for Attributes processor.
- Added ability to create processors in ProcessorFactory interface.

Testing done: automated tests

### DIFF
--- a/processor/addattributesprocessor/testdata/config.yaml
+++ b/processor/addattributesprocessor/testdata/config.yaml
@@ -1,0 +1,18 @@
+receivers:
+  examplereceiver:
+
+processors:
+  attributes:
+  attributes/2:
+    values:
+      attribute1: 123
+      "string attribute": "string value"
+
+exporters:
+  exampleexporter:
+
+pipelines:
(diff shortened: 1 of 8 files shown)


tigrannajaryan:
>Tag means a different thing in OpenCensus concept: https://opencensus.io/tag/. Maybe rename this config to attribute?

@songy23 You are right, I renamed to attribute.

processor/addattributesprocessor/testdata/config.yaml
```diff
@@ -0,0 +1,18 @@
+receivers:
+  examplereceiver:
+
+processors:
+  attributes:
+  attributes/2:
+    values:
+      attribute1: 123
+      "string attribute": "string value"
```
pjanotti: add some key with '.', e.g.: "http.method", we need to ensure that this doesn't break.

tigrannajaryan: Added. Please check again.

PR #569, opened 2019-06-12
Build exporters based on new configuration

tigrannajaryan:
- Introduced a new unisvc binary and make target. Unisvc is
  planned to be the implementation of the unified agent and collector
  that uses the new configuration format.

- Refactored existing Application.execute() code to make it more
  readable and also reusable.
  The functionality is not changed. Reviews: please check this carefully!

- Introduced new Application.executeUnified() function that will call
  all new agent/collector execution logic that is different from
  old logic.
  Application.executeUnified() is currently partial implementation,
  which only builds the exporters. I plan to add building of processor
  pipelines and receivers in future PRs.

- Introduced ExportersBuilder which builds runtime exporters based
  on provided configuration.

- Added ability for App telemetry shutdown. This is required to be able to
  run multiple tests that involve App start and shutdown. Previously
  there was only one test - TestApplication_Start - and it was not
  correctly cleaning up telemetry resources at the end, making impossible
  to add more tests against App.

### DIFF
--- a/cmd/occollector/app/builder/exporters_builder.go
+++ b/cmd/occollector/app/builder/exporters_builder.go
@@ -1,0 +1,252 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package builder
+
+import (
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+
+	"github.com/census-instrumentation/opencensus-service/consumer"
+	"github.com/census-instrumentation/opencensus-service/data"
+	"github.com/census-instrumentation/opencensus-service/exporter"
+	"github.com/census-instrumentation/opencensus-service/internal"
+	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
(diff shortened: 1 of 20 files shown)


cmd/occollector/app/builder/exporters_builder.go
```diff
@@ -0,0 +1,252 @@
...
+			// Remember that this data type is required for the exporter and also which
+			// pipeline the requirement is coming from.
+			result[exporter][pipeline.InputType] = dataTypeRequirement{pipeline}
+		}
+	}
+	return result
+}
+
+// Convert data type enum to string.
+func getDataTypeStr(dataType configmodels.DataType) string {
```
pjanotti: this private should be a public for configmodels.DataType.

cmd/occollector/app/builder/exporters_builder.go
```diff
@@ -0,0 +1,252 @@
...
+
+	factory := factories.GetExporterFactory(config.Type())
+
+	exporter := &exporterImpl{}
+
+	inputDataTypes := exportersInputDataTypes[config]
+	if inputDataTypes == nil {
+		// No data types where requested for this exporter. This can only happen
+		// if there are no pipelines associated with the exporter.
+		eb.logger.Info("Exporter " + config.Name() +
```
pjanotti: nit: Warn instead of Info

cmd/occollector/app/builder/exporters_builder.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func typeMismatchErr(
+	config configmodels.Exporter,
+	requiredByPipeline *configmodels.Pipeline,
+	dataType configmodels.DataType,
+) error {
+	return fmt.Errorf("pipeline %q produces %q to exporter "+
+		"%s which does not support %q telemetry data. "+
+		"exporter will be detached from pipeline", requiredByPipeline.Name,
```
pjanotti: nit: `requiredByPipeline.Name` should be on the next line together with the other parameters 
(or each parameter on its own line, except format string)

tigrannajaryan:
Everything fixed, please review again.

PR #571, opened 2019-06-12
Reimplement prometheus receiver

fivesheep:
## Problems with the current implementation of prometheus receiver
The current prometheus receiver is based on the https://github.com/orijtech/promreceiver  library. in the past couple of months, we have found a couple of issues with this implementation and has reported either in gitter or the library's git repo, which including: 

- dose not support all the data types prometheus: https://github.com/orijtech/promreceiver/issues/5
- throwing error when encounter unknown metrics type which caused the rest of the metric points from the prometheus scrape query to be lost: https://github.com/orijtech/promreceiver/issues/3
- metric labels not handle properly:  https://github.com/orijtech/promreceiver/issues/1
- did not handle NaN values properly: https://github.com/orijtech/promreceiver/issues/2

## Why reimplementing instead of submitting fixes
So far, some of the above issues were fixed, some not. However, this is not the whole story. After diving deeper into the code, we have found that the the library actually has a number of serious bugs, we conclude that without a complete rewrite, it's not going to work properly:

1. A global appender is used, which can easily run into race conditions when multiple scrape target is configured
2. Failed to group metrics in the original form, such as a single group of metric from prometheus can be divided into multiple. This is a big issue if one wants to config a prometheus exporter on the other end of the agent to a prometheus server, as metric group with same name is not allowed.
3. The library is trying to cache all the data scraped from remote metrics endpoints, then feed the delta to downstream, this doesn't seem like what is expected. even worst, it has used a global cache to store the data, which can be easily corrupted when scraping multiple targets
4. The library is also trying to use the cache to generate STDDEV for histogram which is a mission impossible as the original data doesn't provide any original sampled points. 

More details for this PR trying to address can be found in the following document, which is also part of the PR: https://github.com/fivesheep/opencensus-service/blob/reimplement-prometheus-receiver/receiver/prometheusreceiver/README.md

## an example of issues (NaN) values:

with the following ocagent config
```yaml
receivers:
    prometheus:
        config:
            global:
                scrape_interval: 120s
                scrape_timeout: 8s
            scrape_configs:
            - job_name: 'cadvisor'
              scrape_interval: 15s
              static_configs:
              - targets: ['localhost:8888']

exporters:
    prometheus:
            namespace: "promdemo"
            address: "localhost:8889"
```

compare the outputs from original prometheus endpoint and the ocagent prometheus exporter

original (cadvisor)
```
...
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 5.6446976e+07
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1.56037276274e+09
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 1.501503488e+09
```

from ocagent promethues exporter
```
...
# HELP promdemo_process_resident_memory Resident memory size in bytes.
# TYPE promdemo_process_resident_memory counter
promdemo_process_resident_memory NaN
# HELP promdemo_process_start_time Start time of the process since unix epoch in seconds.
# TYPE promdemo_process_start_time counter
promdemo_process_start_time NaN
# HELP promdemo_process_virtual_memory Virtual memory size in bytes.
# TYPE promdemo_process_virtual_memory counter
promdemo_process_virtual_memory NaN
```


songy23:
Please fix the CLA. (You need to use the same Github account and email for the commits.)

fivesheep:
close this PR, I have messed up with the  author names. will create a different one instead

PR #576, opened 2019-06-14
Build pipeline processors based on new configuration

tigrannajaryan:
- Build pipeline processors and plug them into exporters.
- Added tests to verify that single exporter and multiple exporter pipelines
  (fan out) work correctly.
- Minor cleanup in exporters_builder.go
- Fixed opencensus receiver TestCreateReceiver to find available port for
  testing (instead of fixed port which would fail if already listened).

### DIFF
--- a/cmd/occollector/app/builder/pipelines_builder.go
+++ b/cmd/occollector/app/builder/pipelines_builder.go
@@ -1,0 +1,164 @@
+// Copyright 2019, OpenCensus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package builder
+
+import (
+	"fmt"
+
+	"go.uber.org/zap"
+
+	"github.com/census-instrumentation/opencensus-service/consumer"
+	"github.com/census-instrumentation/opencensus-service/internal/configmodels"
+	"github.com/census-instrumentation/opencensus-service/internal/factories"
+	"github.com/census-instrumentation/opencensus-service/processor/multiconsumer"
+)
+
+// builtProcessor is a processor that is built based on a config.
+// It can have a trace and/or a metrics consumer.
+type builtProcessor struct {
+	tc consumer.TraceConsumer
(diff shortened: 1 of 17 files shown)


cmd/occollector/app/builder/pipelines_builder.go
```diff
@@ -0,0 +1,164 @@
...
+
+		factory := factories.GetProcessorFactory(procCfg.Type())
+
+		// This processor must point to the next consumer and then
+		// it becomes the next for the previous one (previous in the pipeline,
+		// which we will build in the next loop iteration).
+		var err error
+		switch pipelineCfg.InputType {
+		case configmodels.TracesDataType:
+			tc, err = factory.CreateTraceProcessor(tc, procCfg)
```
pjanotti: Just occurred to me that we may want to pass the logger to the factories so the instance have access to it. It is unfortunate that the logger is an instance, it will require us to pass it down some way. That said let's not bring it on this PR.

cmd/occollector/app/builder/pipelines_builder.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (eb *PipelinesBuilder) buildFanoutExportersTraceConsumer(exporterNames []string) consumer.TraceConsumer {
+	builtExporters := eb.getBuiltExportersByNames(exporterNames)
+
+	var exporterConsumers []consumer.TraceConsumer
+	for _, impl := range builtExporters {
+		exporterConsumers = append(exporterConsumers, impl.tc)
+	}
+
+	// Optimize for the case when there is only one exporter, no need to create junction point.
```
pjanotti: nit: in practice this is not going to make a real difference since it is done once on initialization but you could instead check `len(builtExporters)` and return `builtExporters[0].tc` without having to allocate the slice. You can also ue this length to create exporterConsumers with the exact size that it needs.

cmd/occollector/app/builder/pipelines_builder.go
```diff
@@ -0,0 +1,164 @@
...
+		result = append(result, exporter)
+	}
+
+	return result
+}
+
+func (eb *PipelinesBuilder) buildFanoutExportersTraceConsumer(exporterNames []string) consumer.TraceConsumer {
+	builtExporters := eb.getBuiltExportersByNames(exporterNames)
+
+	var exporterConsumers []consumer.TraceConsumer
```
pjanotti: nit: `exporters` like on the fanout function for metrics

cmd/occollector/app/builder/pipelines_builder.go
pjanotti: same for the metrics function

cmd/occollector/app/builder/pipelines_builder.go
tigrannajaryan: Sure. We can add it later.

tigrannajaryan:
All fixed.

PR #583, opened 2019-06-18
 Add trace head-sampling

pjanotti:
This is to complete the sampling feature since OC Service already offers tail-sampling. This is implemented using an extra hashing as an attempt to avoid bias from trace ID generation and also to correctly spread traces for backends that also do hashing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,9 +326,9 @@ sampling:
         # sampling-percentage is the percentage of sampling to be applied to all spans, unless their service is specified
         # on sampling-percentage.
         sampling-percentage: 5
-        # hash-seed allows choosing the seed for the hash function used in the trace sampling. This is important where
-        # multiple layers of collectors are being used with different sampling rates, otherwise sampling after first
-        # layer may not work as intended since the hashes will be the ones already selected on first layer.
+        # hash-seed allows choosing the seed for the hash function used in the trace sampling. This is important when
+        # multiple layers of collectors are being used with different sampling rates, in such scenarios make sure to
+        # choose different seeds in each layer.
         hash-seed: 1
 ```
 
--- a/README.md
+++ b/README.md
@@ -309,6 +309,29 @@ global:
         keep: true # keep the attribute with the original key
 ```
 
+### <a name="probabilistic-trace-sampling"></a>Probabilistic Head-based Trace Sampling
+
+In some scenarios it may be desirable to perform probabilistic head-based trace sampling on the collector.
+This can be done using by specifying `probabilistic` policy secion under the `sampling` section of the collector configuration file.
+
+```yaml
+sampling:
+  # mode indicates if the sampling is head or tail based. For probabilistic the mode is head-based.
+  mode: head
+  policies:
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -309,6 +309,29 @@ global:
...
+    # section below defines a probabilistic trace sampler based on hashing the trace ID associated to
+    # each span and sampling the span according to the given spans.
+    probabilistic:
+      configuration:
+        # sampling-percentage is the percentage of sampling to be applied to all spans, unless their service is specified
+        # on sampling-percentage.
+        sampling-percentage: 5
+        # hash-seed allows choosing the seed for the hash function used in the trace sampling. This is important where
+        # multiple layers of collectors are being used with different sampling rates, otherwise sampling after first
+        # layer may not work as intended since the hashes will be the ones already selected on first layer.
```
tigrannajaryan: I think this requires a bit clearer instruction. Perhaps say "make sure to choose different seeds for collectors if they are chained".

processor/tracesamplerprocessor/tracesamplerprocessor.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	}
+
+	sampledTraceData.Spans = sampledSpans
+
+	return tsp.nextConsumer.ConsumeTraceData(ctx, sampledTraceData)
+}
+
+// hash is a murmur3 hash function, see http://en.wikipedia.org/wiki/MurmurHash.
+func hash(key []byte, seed uint32) (hash uint32) {
```
tigrannajaryan: Is there a reason not to use a library that implements this? https://github.com/spaolacci/murmur3

pjanotti: Avoid a new dependency for a standard like murmur3 that has a very small implementation.

README.md
```diff
@@ -309,6 +309,29 @@ global:
...
+  policies:
+    # section below defines a probabilistic trace sampler based on hashing the trace ID associated to
+    # each span and sampling the span according to the given spans.
+    probabilistic:
+      configuration:
+        # sampling-percentage is the percentage of sampling to be applied to all spans, unless their service is specified
+        # on sampling-percentage.
+        sampling-percentage: 5
+        # hash-seed allows choosing the seed for the hash function used in the trace sampling. This is important when
+        # multiple layers of collectors are being used with different sampling rates, in such scenarios make sure to
```
tigrannajaryan: Does it matter that sampling rate is different? If I have a chain of collectors configured with the same sampling rate and hash seed wouldn't it result in downstream collectors simply parroting upstream ones and sampling 100%? I may be me misunderstanding how the feature works, but it probably means we need better doc since our users will likely misunderstand too.

processor/tracesamplerprocessor/tracesamplerprocessor.go
tigrannajaryan: ok

README.md
pjanotti: You're right, it doesn't matter. Let me attempt a rewrite of this.

Issue #585, opened 2019-06-19
Segmentation fault when the collector 


asutoshpalai:
If we start Collector at the same port as Agent, we get segmentation fault due to the way the error is handled at at [cmd/occollector/app/collector/receivers.go#L54](https://github.com/census-instrumentation/opencensus-service/blob/dee9655f45a480266d5ab80ef206b5faf97c2f04/cmd/occollector/app/collector/receivers.go#L54).

### Reproducing the error

Agent config
```yaml
receivers:
  opencensus:
    address: ":55678"

exporters:
  opencensus:
    endpoint: "localhost:55680"

zpages:
  port: 8884
```

Collector config
```yaml
collector-config.yaml 
log-level: DEBUG
receivers:
  opencensus:
    port: 55678

queued-exporters:
  jaeger-all-in-one:
    num-workers: 4
    queue-size: 100
    retry-on-failure: true
    sender-type: jaeger-thrift-http
    jaeger-thrift-http:
      collector-endpoint: http://localhost:14268/api/traces
      timeout: 5s

zpages-http-port: 9889
zpages:
  port: 8889
```

First start the agent and then the collector.

### Cause
```go
...
			rec, err := receiver.runFn(logger, v, traceConsumers, asyncErrorChan)
			if err != nil {
				...
				logger.Fatal("Cannot run receiver for "+rec.TraceSource(), zap.Error(err))
			}
...
```

If there is an error, `rec` might be `nil`, so calling `rec.TraceSource()` will give a segmentation fault.

pjanotti:
Good find @asutoshpalai. Do you have cycles to submit a PR fixing it?

asutoshpalai:
I am currently slammed at work, but I can definitely get around to it during the weekend!

pjanotti:
Alright then! I'll assign this to you.

asutoshpalai:
Fixed in #592

PR #592, opened 2019-06-26
Fixes occollector crash on exporter start failure #585

asutoshpalai:


pjanotti:
Nice find @asutoshpalai! Could you please sign the CLA so we can merge your fix?

asutoshpalai:
I sighed the cla, hopefully this should be good to go now!

PR #594, opened 2019-06-27
jaeger: Don't error on nil parent links.

### LINKED ISSUE #593: collector: jaeger exporter dropping root spans
I'm testing opencensus service (naster ) with: 
```jaeger hotrod demo -> ocagent (jaeger-compact-thrift) -> occollector -> jaeger-collector```

In this configuration, all the intial root spans from the first web request are being dropped by occollector. They appear to contain a nil Link parent link. The entire batch of span the root span is in gets thrown away. I've dumped the links in spans from the agent, and it certainly looks like the jaeger client in hotrod is actually sending this nil link, ocagent is faithfully converting them.

tcolgate:
jaeger-client-go appears to include a nil parent link in root spans.
These cause export of batches to a jaeger-collector to fail.

fixes #593

### DIFF
--- a/translator/trace/jaeger/jaegerthrift_to_protospan.go
+++ b/translator/trace/jaeger/jaegerthrift_to_protospan.go
@@ -161,6 +161,12 @@ func jReferencesToOCProtoLinks(jrefs []*jaeger.SpanRef) *tracepb.Span_Links {
 	links := make([]*tracepb.Span_Link, 0, len(jrefs))
 
 	for _, jref := range jrefs {
+		if jref.TraceIdHigh == 0 &&
+			jref.TraceIdLow == 0 &&
+			jref.SpanId == 0 {
+			continue
+		}
+
 		var linkType tracepb.Span_Link_Type
 		if jref.RefType == jaeger.SpanRefType_CHILD_OF {
 			// Wording on OC for Span_Link_PARENT_LINKED_SPAN: The linked span is a parent of the current span.


pjanotti:
Thanks for reporting and having a PR @tcolgate!

pjanotti:
So at this moment you are not 100% sure yet where this nil TraceID is being generated, correct?

tcolgate:
I dumped the links from the incoming thrift, and they appeared to be in there, which makes me think that they are actually in the incoming spans, I am not 100% certain though.
It doesn't seem quite right that one dodgy span ID in one link blows out a whole batch though (that's a slightly different discussion I guess)

tcolgate:
@songy23 @pjanotti can we settle on where the right place to fix this is? It feels like, if lots of stuff assumes there will be no nil-links with the native OC spans, then it makes more sense to drop the duff links at ingest in the agent right?

tcolgate:
I've moved the fix to the agent, going on the "..., be liberal in what you accept" principle. That way the internal OC spans are always consistent.

pjanotti:
@tcolgate I agree with not dropping the whole batch due to incorrect single span, let's address that on OpenTelemetry since OpenCensus Service is going to maintenance mode. As your investigation pointed the issue seems specific to the HotRod example, in this context it seems to me that it doesn't meet the bar for a change now in OC service at this moment.

pjanotti:
Closing per previous comment.

Issue #589, opened 2019-06-26
Prometheus Receiver: Summaries with empty quantiles should map to empty percentiles


dinooliva:
For detailed discussion, see https://github.com/census-instrumentation/opencensus-service/pull/572#discussion_r296345632

Issue #588, opened 2019-06-26
Prometheus Receiver: Initial time series values and resets not handle appropriately


dinooliva:
For detailed discussion, see: https://github.com/census-instrumentation/opencensus-service/pull/572#discussion_r293964652

dinooliva:
Fixed by #598

PR #597, opened 2019-07-01
Fix starttime and summary has no quantiles for prometheus receiver

fivesheep:
To address start time for counter/histogram/summary, and take delta as values for these types.  
For issue: https://github.com/census-instrumentation/opencensus-service/issues/588

It also fixes the issue when summary has no quantiles, allow it to produce a summary with nil snapshots
For issue: https://github.com/census-instrumentation/opencensus-service/issues/589

cc @dinooliva @rghetia


receiver/prometheusreceiver/internal/metadata.go
```diff
@@ -30,31 +33,90 @@ type MetadataService interface {
...
 type mCache struct {
-	t *scrape.Target
+	t           *scrape.Target
+	startValues map[string]*ObservedValue
+	accessTimes int64
+}
+
+// EvictUnused is used at the end of each scrape run for removing cached startValues which are not accessed in the past
+// maxUnusedRuns, as metrics can come and go from a scrape target endpoint
+func (m *mCache) EvictUnused() {
```
rghetia: How does it evict if the target is down and never comes up?

fivesheep: @rghetia even if the target is down, or not in good health, prometheus still try to initialize an appender and feed data to it (with Value equals to NaN), as I commented on in the rewritten EndToEnd test.  logic can be found from: https://github.com/prometheus/prometheus/blob/851131b0740be7291b98f295567a97f32fffc655/scrape/scrape.go#L933-L935

however, I did make an assumption that a configured target will never be evicted even if it's down, during the lifecycle of opencensus agent.  the eviction is for metrics come and go, such as container metrics from a cadvisor endpoint

rghetia: SG.

fivesheep:
@rghetia @dinooliva  updated as per #598  
please do review the rewritten TestEndToEnd which integrated all the new things together, it's the best place to describe the receiver's behavior

fivesheep:
the test failure is not related to the PR:
```
=== RUN   TestReception
--- FAIL: TestReception (0.00s)
    trace_receiver_test.go:47: Starting
    trace_receiver_test.go:50: StartTraceReception failed: listen udp :6832: bind: address already in use
FAIL
coverage: 80.5% of statements
FAIL	github.com/census-instrumentation/opencensus-service/receiver/jaegerreceiver	1.173s
```

receiver/prometheusreceiver/metrics_receiver_test.go
```diff
@@ -79,44 +105,205 @@ func TestNew(t *testing.T) {
 	}
 }
 
+// Test data for EndToEnd test
+
+var target1Page1 = `
```
dinooliva: @fivesheep - yes, the endtoend test does help clarify things. A couple of comments:
1. Adding a some comments on the target pages about what's being tested would be helpful.
2. It looks like resets aren't tested - adding that to a test page would also be worthwhile
3. It also looks like empty histogragms/summaries aren't tested here either - I think they would be worthwhile to have too.

fivesheep: sure will do.  
I just checked the reset cases, it seems like when it resets, the adjuster will discard the value (use it as initial) when a reset is detected.  Is it an expected behavior? I thought it would use the timestamp from previous scrape as starttime, and value as it is (starting from 0). it's not right or wrong here, just a decision of best-effort.  if this is what we want, I can keep the test case this way.

dinooliva: Yes, a reset acts like an initial point since we don't know when the reset actually occurred. Using the timestamp from the previous scrape does give a much more constrained lower bound than the initial point and may be worth considering - I need to think through the implications.

@rghetia - thoughts?

fivesheep:
@songy23  merged
@dinooliva I have add a test for the resetting scenario,  in case we might change the behavior of using the previous scrape timestamp as starttime, I have also restructure the end to end tests to make it easy to modify and adjust

receiver/prometheusreceiver/internal/metadata_test.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package internal
+
+import (
+	"testing"
+)
+
+func Test_mCache_EvictUnused(t *testing.T) {
+
```
songy23: Empty test?

songy23:
Will merge this PR by EOD tomorrow if there're no further concerns/comments.

receiver/prometheusreceiver/metrics_receiver_test.go
rghetia: It is better to treat reset as initial point. Consider this
0. At time T0 a job reported data
1. At time T1 a job stops reporting for whatever reason.
2. At time T2 the job resets
3. At time T3 the job resumes reporting.

If time lag between T0 and T3 is within scraping interval then it is okay to use T0 as a starttime. Otherwise it will provide incorrect data. I would always prefer no-data vs incorrect data.

PR #598, opened 2019-07-01
Fix issues with initial start time and restarts in the Prometheus receiver.

dinooliva:
Code is complete - ptal. Currently working on updating/adding unit tests.


receiver/prometheusreceiver/internal/metrics_adjuster.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (ma *MetricsAdjuster) adjustTimeseries(metricType metricspb.MetricDescriptor_Type, current, initial *metricspb.TimeSeries) bool {
+	if !ma.adjustPoints(metricType, current.GetPoints(), initial.GetPoints()) {
+		return false
+	}
+	current.StartTimestamp = initial.StartTimestamp
+	return true
+}
+
+func (ma *MetricsAdjuster) adjustPoints(metricType metricspb.MetricDescriptor_Type, current, initial []*metricspb.Point) bool {
```
rghetia: you can probably get rid of adjustPoints.  In adjustTimeseries
```go
func (ma *MetricsAdjuster) adjustTimeseries(metricType metricspb.MetricDescriptor_Type, current, initial *metricspb.TimeSeries) bool {
        cPoints := current.GetPoints()
        iPoints := initial.GetPoints()
        if len(cPoints) !=1 || len(iPoints) != 1 {
            // log if needed.
            return false
        }

	if !ma.adjustPoint(metricType, cPoints[0], iPoints[0]) {
		return false
	}
	current.StartTimestamp = initial.StartTimestamp
	return true
}
```

receiver/prometheusreceiver/internal/metrics_adjuster.go
```diff
@@ -0,0 +1,169 @@
...
+// - MetricDescriptor_CUMULATIVE_DISTRIBUTION
+// - MetricDescriptor_GAUGE_DISTRIBUTION
+// - MetricDescriptor_SUMMARY
+func (ma *MetricsAdjuster) adjustPoint(metricType metricspb.MetricDescriptor_Type, current, initial *metricspb.Point) bool {
+	switch metricType {
+	case metricspb.MetricDescriptor_CUMULATIVE_DISTRIBUTION:
+		// note: sum of squared deviation not currently supported
+		initialDist := initial.GetDistributionValue()
+		currentDist := current.GetDistributionValue()
+		if currentDist.Count < initialDist.Count {
```
rghetia: it should be if either count or sum is less than previous value

receiver/prometheusreceiver/internal/metrics_adjuster.go
```diff
@@ -0,0 +1,169 @@
...
+		// this shouldn't happen
+		ma.logger.Info("adjust unexpect point type ", metricType.String(), ", skipping ...")
+	}
+	return true
+}
+
+func (ma *MetricsAdjuster) adjustBuckets(current, initial []*metricspb.DistributionValue_Bucket) bool {
+	if len(current) != len(initial) {
+		// this shouldn't happen
+		ma.logger.Info("len(current buckets) != len(initial buckets)", len(current), len(initial))
```
rghetia: should you change this  to Error?

receiver/prometheusreceiver/internal/transaction.go
```diff
@@ -47,19 +49,24 @@ type transaction struct {
 	ctx           context.Context
 	isNew         bool
 	sink          consumer.MetricsConsumer
+	jobsMap       *map[string]*map[string]*metricspb.TimeSeries
```
rghetia: may be another type for map[string]*metricspb.Timeseries would make it more readable.
```go

type timeseriesMap map[string]*metricspb.Timeseries
type jobsCache map[string]*timeseriesMap

type transaction struct {
    jobsMap *jobsCache
}
```

fivesheep:
Oops, looks like this PR is addressing the same issue with https://github.com/census-instrumentation/opencensus-service/pull/597 which I submitted earlier.   I had actually tried a similar approach at the very beginning by caching the first generated TimeSeries, however, it's not going to work properly in some cases, including:

1) as @rghetia also mentioned, to detect if a remote server has restarted, one needs to compare the current value with the last observed value
2) the program also needs to clean up the cache, as timeseries can come and go. For example, in our use case we mainly use ocagent to scrape cadvisor metrics, which includes a lot of container spec metrics the lifecycles of which are tied to the containers but not cadvisor. 
3) you might observe new labels from subsequent runs. take a look at the following example:
scrape run 1
```
# HELP container_memory_failures_total Cumulative count of memory allocation failures.
# TYPE container_memory_failures_total counter
container_memory_failures_total{failure_type="pgfault",id="/",image="",name="",scope="container"} 760009
```
from an Appender's perspective, only the label `failure_type` and `container` can be observed, as any empty labels with empty labels are removed before hitting the Add/AddFast method of an appender

then in the 2nd run, you get a new one with some empty labels are filled
```
# HELP container_memory_failures_total Cumulative count of memory allocation failures.
# TYPE container_memory_failures_total counter
container_memory_failures_total{failure_type="pgfault",id="/",image="",name="",scope="container"} 760010
container_memory_failures_total{failure_type="pgfault",id="/docker/6a26e59b145922c813444ca59b46cc3186e6e64064744f973b59f81dbd913fa8",image="somedocker-image",name="some-name",scope="container"} 1.877625e+06
```
now the timeseries will have all 4 labels in the 2nd run.  and it will be challenging to link them back.  not only the label values, the label orders are also important.

4) for any failed scrape (say remote endpoint returning some temp errors), prometheues scraper library will still feed data with the metric labels from cache and value NaN to the appender.  the appender also needs to filtered out such data

You might want to take a look at the PR I have submitted and take the works from there. I have also added quite a lot tests trying to capture all these corner cases as well as a redesigned EndToEnd test which is able to produce more stable results and simulate edge cases like remote server is not responding properly

rghetia:
#597 covers more cases than #598. However, in #598 MetricAdjuster is independent of scrapping which is useful another [envoy prometheus receiver](https://github.com/rghetia/opencensus-service/tree/envoy_prometheus_receiver/receiver/envoyreceiver) which converts from protobuf (instead of text) to opencensus-proto.

dinooliva:
@fivesheep - thank you for the detailed response and my apologies in they delay in responding. As @rghetia mentioned, I took this approach because we'd like to reuse the logic across both the text and proto version of the prometheus receiver. In response to your particular comments:

1. You're right that looking at the previous value is probably a better approach to detecting a reset than looking at the initial value - that should straightforward to add to this approach.

2. There's not really a perfect solution to this problem but adding a ttl field (or similarly) should be straightforward to add.

3. That's a very interesting point that I need to understand it better so I can't say whether or not it's feasible to add to this approach - I'll take a look over your pr shortly.

4. That's a great point too - that should also be straightforward to add.

fivesheep:
@dinooliva  it does make sense to have a generic solution which is able to adjust income metrics values for different Prometheus protocols .  I am not sure if it's possible to make it one more layer up to be a more generic solution for any metrics receivers which has the need to adjust values, with a simple interface like:
```
type MetricsAdjuster interface {
    AdjustMetric(*data.MetricsData) *data.MetricsData
}
```
and make it as an option for the metric receiver factory. 

other than that in https://github.com/census-instrumentation/opencensus-service/pull/597 I have also refactored the metricbuilder and add a layer called metric family which makes the code cleaner and easier to understand, you might also want to take a look at it.

dinooliva:
@fivesheep - I spent some time looking over your solution and your notes. I think it may be best to go with a combined solution. My module is strictly for adjusting start times and values based on discovered initial timeseries and resets.

Dealing with newly discovered labels and empty histograms/summaries are not really issues that the metrics adjuster should deal with (I just have to make sure that the signature calculation works appropriately for empty label values) but those issues do need to be dealt with in the metricsbuilder, as you have done.

I plan to add a couple of commits that deal with the issues 1 & 2. Issue 3 I think should be handled by the metricsbuilder and issue 4 I think can also be dealt with using your solution.

dinooliva:
PTAL

This pr now addresses the issue with restarts and adds more unit tests. Once this pr has been merged, I will submit another pr to deal with cleaning up the cache (I don't want this pr to get too complicated).

receiver/prometheusreceiver/internal/metrics_adjuster.go
dinooliva: Done

receiver/prometheusreceiver/internal/metrics_adjuster.go
dinooliva: I generally avoid using Error unless something major is going on but that's based on Google prod convention - what is the convention for the oc agent when Error is appropriate?

receiver/prometheusreceiver/internal/metrics_adjuster.go
dinooliva: I generally avoid using Error unless something major is going on but that's based on Google prod convention - what is the convention for the oc agent when Error is appropriate?

receiver/prometheusreceiver/internal/metrics_adjuster.go
dinooliva: I've simplified it - ptal.

receiver/prometheusreceiver/internal/transaction.go
dinooliva: Done

receiver/prometheusreceiver/internal/metrics_adjuster_test.go
```diff
@@ -0,0 +1,382 @@
...
+		[]*metricspb.Metric{},
+	}, {
+		"Summary: round 4 - instance adjusted based on round 3",
+		[]*metricspb.Metric{summary(k1k2, timeseries(4, v1v2, summ(4, 14, 96, percent0, []float64{9, 47, 8})))},
+		[]*metricspb.Metric{summary(k1k2, timeseries(3, v1v2, summ(4, 2, 30, percent0, []float64{9, 47, 8})))},
+	}}
+	runScript(t, script)
+}
+
+func Test_multiMetrics(t *testing.T) {
```
rghetia: MultiMetrics test covers everything that individual metric test covers. may be just keep the mulitmetric test.

receiver/prometheusreceiver/internal/metrics_adjuster.go
```diff
@@ -0,0 +1,220 @@
...
+// mean reallocating all of the metrics.
+func (ma *MetricsAdjuster) adjustPoint(metricType metricspb.MetricDescriptor_Type, current, initial, latest *metricspb.Point) bool {
+	switch metricType {
+	case metricspb.MetricDescriptor_CUMULATIVE_DOUBLE:
+		currentValue := current.GetDoubleValue()
+		initialValue := initial.GetDoubleValue()
+		latestValue := initialValue
+		if initial != latest {
+			latestValue += latest.GetDoubleValue()
+		}
```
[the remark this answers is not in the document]
dinooliva: Done.

receiver/prometheusreceiver/internal/metrics_adjuster_test.go
dinooliva: I considered this as well  - the individual tests are easier for debugging purposes and documenting what should happen for the different aggregation types.

Unless you object, I propose keeping them for now - if maintaining them is problematic, we can remove them subsequently.

dinooliva:
@songy23 - ptal

receiver/prometheusreceiver/internal/metrics_adjuster.go
```diff
@@ -0,0 +1,220 @@
...
+	for _, metric := range metrics {
+		if ma.adjustMetric(metric) {
+			adjusted = append(adjusted, metric)
+		}
+	}
+	return adjusted
+}
+
+// returns true if at least one of the metric's timeseries was adjusted and false if all of the timeseries are an initial occurence or a reset.
+// Types of metrics returned supported by prometheus:
```
songy23: How about gauge and cumulative int64 values?

dinooliva: These kinds of metrics aren't generated by the Prometheus -> OC Metrics translation so I didn't add support for them.

dinooliva:
@fivesheep 

We've merged this code, which resolves issue 1 of the 4 issues that you originally noted. Of the other 3 issues:

Issue 2 needs to be handled by the metrics adjuster and I'm working on a solution to it. I had been planning to cleanup at the job-level rather than at the timeseries level but that doesn't seem to work for your use-case (cadvisor) so I will look into handling individual timeseries.

Issue 3 is separate from the metrics adjuster and your code already deals with it.

Issue 4 is also separate from the metrics adjuster and, again, your code deals with it.

Would you be able to update your pr to remove the issue 1 and issue 2 related code while keeping the rest?

fivesheep:
@dinooliva sure will do.  are we still going to have a release with this two PRs, or release will only be in the new project?

dinooliva:
@fivesheep I think it would make sense to make a new release with these changes but I'll have to consult with the project owners to see what their current process is.

@songy23 - any thoughts?

songy23:
We can still make patch releases for small improvement, bug fixes etc. @flands Could you help?

pjanotti:
That are a few pending PRs that started before we made the announcement that we were in "maintenance mode". Assuming that this can wait a bit let's close those before cutting a new release.

Issue #198, opened 2018-11-15
Use of exporters drop service/process information


pjanotti:
The service (agent and collector) can receive spans from multiple processes (remote or not), while this information can be encoded as some type of tag on each span many backends receive this information as a separate block of data to minimize transport cost (e.g.: Node for oc-proto, Process for Jaeger batch). When using existing exporters this information can be lost, e.g.: Jaeger exporter sets the Process information during creation.

odeke-em:
@pjanotti thanks for filing this issue.

Exporters like Stackdriver Trace don't seem to have a field for it but perhaps let's kindly page someone from the Stackdriver team to ask what attributes we can use for it 
 https://godoc.org/google.golang.org/genproto/googleapis/devtools/cloudtrace/v2#Span_Attributes /cc @mtwo 
neither does Zipkin /cc @adriancole. Stackdriver Monitoring however takes in the processID information.

This issue would involve separate digging through the various exporters'  repositories which aren't maintained by us.

gitlw:
@pjanotti @songy23 

I ran into the same issue, when I tried to export tracing info from two types of applications A and B into the opencensus-agent, and then to the datadog-agent, and then to datadog. In the end, the service names A and B are lost in the datadog UI, and all traces have the name opencensus-app.
 
It seems the node info is lost during the translation at
https://github.com/census-instrumentation/opencensus-service/blob/master/exporter/exporterwrapper/exporterwrapper.go#L64

Before the translation, the node field in data.TraceData has the service name.
After the translation, the original service name in the node is lost.
Please let me know if you need anything else from me. Thanks!

songy23:
@gitlw Thanks for looking into this issue. Service names and other information in `Node` are dropped because the current `ExportSpan` interface doesn't accept `Node`: https://github.com/census-instrumentation/opencensus-go/blob/bf352733d4dffdd0e650c71cabc694aca654be9a/trace/export.go#L31. @sjkaris is working on adding a different interface `ExportSpanProto` that consumes both the proto format spans and `Node` information. Until that happened, for now you can try adding the service names as an attribute in the spans.

gitlw:
Thanks for your quick reply @songy23 . I'll stay tuned for the change from @sjkaris

PR #602, opened 2019-07-05
Add support for specifying service name in config

mads-hartmann:
The datadog exporter will currently drop the service name which is specified in the spans. See https://github.com/census-instrumentation/opencensus-service/issues/198 for more information.

This aligns the datadog exporter with jaeger, wavefront, and zipkin. They already have this option.

A better solution would be to add support for multiple services, as many of the other exporters have, but that is a bit over the level of contribution I can do at the moment.

I've tested this against a live datadog setup and the service name is correctly propagated and displayed in their UI.

If you are okay with this contribution I will update the docs accordingly 😊


mads-hartmann:
@pjanotti no problem 😊 thanks for reviewing.

PR #611, opened 2019-07-19
receiver/opencensus/ocmetrics: ignore io.EOF on JSON upload

### LINKED ISSUE #610: receiver/opencensus/ocmetrics: return a nil error on io.EOF for grpc-gateway
receiver/opencensus/octrace recognizes that we'll receive telemetry from the grpc-gateway and JSON inputs that won't maintain connections/streams hence currently has this special case

https://github.com/census-instrumentation/opencensus-service/blob/d4f12f7109972216fc37f6c9f42edc4136a972fb/receiver/opencensusreceiver/octrace/opencensus.go#L136-L144

Let's do the exact same for receiver/opencensus/ocmetrics

odeke-em:
When the grpc-gateway was added to this receiver/opencensus
Trace receiver in PR #270, we recognized that browsers
and other sources that send JSON in one burst won't stream
so would encounter an io.EOF. This PR makes the metrics
receiver replicates that behavior.

We'll get it this PR also ported over to the OpenTelemtry
service.

Fixes #610

### DIFF
--- a/receiver/opencensusreceiver/ocmetrics/opencensus.go
+++ b/receiver/opencensusreceiver/ocmetrics/opencensus.go
@@ -17,6 +17,7 @@ package ocmetrics
 import (
 	"context"
 	"errors"
+	"io"
 	"time"
 
 	"google.golang.org/api/support/bundler"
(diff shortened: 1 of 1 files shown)


odeke-em:

odeke-em:
The corresponding opentelemetry-service PR is at https://github.com/open-telemetry/opentelemetry-service/pull/179

PR #613, opened 2019-07-23
Add garbage collection support to the Prometheus metrics adjuster

dinooliva:


dinooliva:
@rghetia - ptal

receiver/prometheusreceiver/internal/metrics_adjuster.go
```diff
@@ -42,55 +102,102 @@ func getSignature(name string, values []*metricspb.LabelValue) string {
...
+			if !tsm.mark {
+				delete(jm.jobsMap, sig)
+			} else {
+				tsm.gc()
+			}
+			tsm.Unlock()
+		}
+		jm.lastGC = time.Now()
+	}
+	jm.Unlock()
```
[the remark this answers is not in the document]
dinooliva: Yes, I can move the tsm.lock() to tsm.gc().

It is is possible that the metricsadjuster could mark the tsm and it will be removed from the JobsMap when it didn't need to be but that would happen even with locking (in a slightly different order).

receiver/prometheusreceiver/internal/metrics_adjuster.go
```diff
@@ -42,46 +104,87 @@ func getSignature(name string, values []*metricspb.LabelValue) string {
...
 }
 
-func (jm *JobsMap) get(job, instance string) *metricsInstanceMap {
+// Remove jobs and timeseries that have aged out.
+func (jm *JobsMap) gc() {
+	current := time.Now()
+	jm.Lock()
+	defer jm.Unlock()
+	// recheck lastGC to make sure gc is necessary
+	if current.Sub(jm.lastGC) > jm.gcInterval {
```
songy23: This seems duplicated to https://github.com/census-instrumentation/opencensus-service/pull/613/files#diff-6d65923b82ce96fdb3462374e44dd2e8R140.

dinooliva: Yes, the previous check is done before locking jm - this check is to verify that no other gc() has happened in the interim before we grab the lock. I'll add a comment to clarify what's going on.

PR #618, opened 2019-08-08
Upgrade Stackdriver and OcAgent exporter versions

songy23:
Also run $ go mod tidy.


songy23:
Need one more approval, @pjanotti @tigrannajaryan PTAL :)

PR #621, opened 2019-08-28
Support environment variables when configuring exporters

c2nes:
`viper.AutomaticEnv` is enabled and environment variable overrides works
for configuration sections outside of `exporters`. However, a Viper
limitation (https://github.com/spf13/viper/issues/507) prevents this
from working as expected in "sub" viper instances.

This work around allows environment variables to be used with exporter
configurations as well.


dashpole:
This repository is now archived, and is succeeded by https://github.com/open-telemetry/opentelemetry-collector.

PR #624, opened 2019-10-18
fix: add missing tracing status from jaeger thirft

BradErz:
`status.message`, `status.code` and `error` were missing from the opencensus collector when forwarding the request to jaeger over `jaeger-http-thift` which seems to be the default config.


BradErz:
@googlebot I signed it!

PR #625, opened 2019-10-19
receiver: add UNIX Domain Socket

odeke-em:
Brings in the source code from opencensus-php/daemon
that implements the UNIX Domain Socket receiver and
performs aggregation on the daemon before forwarding
it to the ocagent via gRPC. The overall design is
present at
https://docs.google.com/document/d/1CRiRq_wpzOuG9VKM_eaImcrS12Iie2V7LePnH9AwclU/

In follow-up commits, we'll then start plumbing it to
work directly with the agent. The first step is bringing
in the source code.


odeke-em:
So initially I was going to bring it into the OpenTelemetry collector contrib repo as per https://github.com/open-telemetry/opentelemetry-collector/issues/396, but:
a) this code needs go.opencensus.io/ and a complete implementation of tracing and metrics libraries which I don't think OpenTelemetry has yet
b) this code is currently meant for use with OpenCensus-PHP metrics so having it here
reduces the burden
c) It was written for OpenCensus and as a link between the agent and libraries for which tracing and metrics implementations are very trick

PR #637, opened 2020-02-25
go.mod: update to orijtech/prometheus-export@v0.0.3

odeke-em:
This update features a dependency update to bring in a fix
which removes key length truncation to 100 which AFAIK is
an arbitrary key length. This issue was reported by @yiyijin in:
* https://github.com/orijtech/prometheus-go-metrics-exporter/issues/4
and has been fixed by:
* https://github.com/orijtech/prometheus-go-metrics-exporter/pull/5


odeke-em:
Thank you for the review @pjanotti! For the unrelated test failure, I've filed https://github.com/census-instrumentation/opencensus-service/issues/638.

odeke-em:
@yiyijin once this PR is merged, you'll be good to go!

pjanotti:
@odeke-em iirc in the past, I could re-start the build in case of failures but it seems that I don't have the right to do that anymore (at least it is no available on Travis). You may have to force-push to pass the check.

odeke-em:
>@odeke-em iirc in the past, I could re-start the build in case of failures but it seems that I don't have the right to do that anymore (at least it is no available on Travis). You may have to force-push to pass the check.

Thank you for the review @pjanotti! I was able to re-run the Travis CI build, it passed and then I merged.

PR #639, opened 2020-03-04
Allow specifying an address for all receivers

dmathieu:
Listening on all network interfaces is a security risk on our platform, as we can't easily restrict which ports are available to other components within the cluster.
In order to setup an authentication reverse proxy and have it really be enforced, we therefore need to be able to start the endpoints on the loopback network interface only.

### DIFF
--- a/cmd/occollector/app/builder/builder.go
+++ b/cmd/occollector/app/builder/builder.go
@@ -74,6 +74,13 @@ func DebugTailSamplingEnabled(v *viper.Viper) bool {
 
 // JaegerReceiverCfg holds configuration for Jaeger receivers.
 type JaegerReceiverCfg struct {
+	// Address is an IP address or a name that can be resolved to a local address.
+	//
+	// It can use a name, but this is not recommended, because it will create
+	// a listener for at most one of the host's IP addresses.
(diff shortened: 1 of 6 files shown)


flands:
@dmathieu sure, but you can always change the config. If you deploy as a collector you need to expose anyway. In addition, the receivers can support authentication if needed.

dmathieu:
I know the collector needs exposing anyway. But by limiting the network interface here, I can setup a reverse proxy on top of it, which handles authentication.
I can then make sure only the reverse proxy is available, not the collector running without authentication.

As for the receivers supporting authentication, I'm not sure what you mean. I've not seen any documentation or hints in the code about setting that up.

dmathieu:
Let me try to give additional context. This is a simplified diagram of our architecture:

heroku-occollector

Because our platform makes all ports open on the public network interface to any other container in the same cluster, a compromised process could bypass authentication in our reverse proxy and send any traces to the collector.

Specifying the network interface on which the collector needs to listen will allow us to prevent connections from outside the container itself, and remove that security concern.

dmathieu:
Closing as stale.

